PR #19, opened 2023-08-21
experimental Implementation of Minimal Encoding Method

Jacomichi:
# Added Feature:
I have implemented the minimal encoding method. Unlike the conventional VQE, we do not use the measurement result itself as the cost function. Instead, we use a function defined based on the measurement results. To include features such as cost function generation, I have designed it such that we create a `MinimalEncodedInstance` from the `MinimalEncodingBuilder`, and then utilize this instance to extract the cost function and obtain the final results.

# Concerns:
With the above-described structure, it deviates from the form of other algorithms. I would appreciate any advice or feedback on this aspect.


yuyamashiro (changes requested):
たぶん僕も同じような構成にすると思います。構成はこれで問題ないです。
しかし他の関数と出力が異なるので、Minimal encodingの概要をどこかのDocstringに書いて、なんでminimized functionが出力されるのかは書いた方がいい気がします。

Jacomichi:
コメントありがとうございます！

構成はこのままで、minimal encodingの概要を詳細に説明するという方針で進めたいと思います.

Jacomichi:
docstringの修正を行いました。

yuyamashiro (approved):

Issue #35, opened 2023-12-14
If there is no quad term, QRAO causes ValueError


Jacomichi:
# What is problem
When we try to convert Ising Hamiltonian which has only linear term to QRAO Hamiltonian, it gives ValueError
The error message is as below.
```bash
ValueError: max() arg is an empty sequence
```

# How to reproduce
```python
ising = IsingModel({}, {0:1.0,1:1.0,2: 5.0, 3: 2.0}, 6.0)
    max_color_group_size = 3
    _, color_group = greedy_graph_coloring(
        ising.quad.keys(), max_color_group_size=max_color_group_size
    )
    color_group = check_linear_term(
        color_group, ising.linear.keys(), max_color_group_size
    )

    qrac_hamiltonian, offset, encoding = jtqp.qrao.qrac31_encode_ising(
        ising, color_group
    )
```

Jacomichi:
This issue is fixed at #36.

PR #36, opened 2023-12-14
fix the issues #35

Jacomichi:
# Change
The problem is caused by try to know the current max index of the quad term like the below code
```python
max_idx = max(color_group.keys()) 
```

This causes error when the `color_group` is empty, so I change this as below
```python
max_idx = max(color_group.keys()) if color_group else -1
```
When the dictionary is empty, -1 is returned. This is because we would like to start the linear term from the max index of the quad term, so the initial linear index should be 0, so I set the return value as -1.

### DIFF
--- a/jijmodeling_transpiler_quantum/core/qrac/graph_coloring.py
+++ b/jijmodeling_transpiler_quantum/core/qrac/graph_coloring.py
@@ -92,7 +92,7 @@ def check_linear_term(
     for v in color_group.values():
         idx_in_color_group.extend(v)
 
-    max_idx = max(color_group.keys())
+    max_idx = max(color_group.keys()) if color_group else -1
     value_counter = 1
     for idx in linear_term_index:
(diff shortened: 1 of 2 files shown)


jijmodeling_transpiler_quantum/core/qrac/graph_coloring.py
```diff
@@ -92,7 +92,7 @@ def check_linear_term(
     for v in color_group.values():
         idx_in_color_group.extend(v)
 
-    max_idx = max(color_group.keys())
+    max_idx = max(color_group.keys()) if color_group else -1
```
yuyamashiro, line 95: この処理はコメントで意図を書いておいた方が良いです。

yuyamashiro (changes requested):
Please add a comment to the line for clarification.

yuyamashiro (approved):

PR #51, opened 2024-08-14
re-create qamomile

yuyamashiro:


yuyamashiro:
We have introduced an intermediate representation for quantum circuits and operators in Qamomile, which has led to a significant revision of Qamomile’s interface. The Converter class has been introduced. This Converter class provides encoders and decoders that transform instances of constrained optimization problems (compiled_instance) into objects suitable for quantum algorithms such as QAOA and QRAO. The encoder converts these instances into the intermediate representation. Additionally, Qamomile now includes the ability to modify quantum circuits and operators using this intermediate representation, allowing users to customize algorithms using only Qamomile, independent of the SDK.

Next, we introduced the Transpiler class. The Transpiler class provides functionality to convert Qamomile’s intermediate representation into SDKs like Qiskit or Quri-parts. With the introduction of the intermediate representation, we have clearly separated the processes of conversion and transpilation, and these are now part of the interface that users can directly interact with. Basic usage can be found in qamomile-docs/quickstart.ipynb.

If you need any further modifications, feel free to ask!

yuyamashiro:
Tasks
- [ ] implement rxx ryy rzz
- [ ] write quri-parts test
- [ ] write qaoa and alternative ansatz example notebook

Jacomichi (approved):
LGTM
The rest of the QRAO Hamiltonian will be added in a separate PR.

PR #57, opened 2024-08-20
add efficient su2 ansatz

Jacomichi:
# Change
A class that generates Efficient SU2 ansatz has been added.
The interface is a simplified version of qiskit's `EfficientSU2`.
However, not all argument names are exactly the same.


Jacomichi:
I finally implemented it as a function.

yuyamashiro (approved):

PR #61, opened 2024-08-23
add drawer

Jacomichi:
# Change
I created a circuitry drawer.
You need to include matplotlib in the dependencies.

# Example
```python
from qamomile.core.circuit.drawer import plot_quantum_circuit

def create_vqe_circuit(n_qubits: int, depth: int) -> QuantumCircuit:
    """Create a circuit mimicking a part of Variational Quantum Eigensolver."""
    qc = QuantumCircuit(n_qubits,n_qubits, name="VQE-like Circuit")
    for d in range(depth):
        for i in range(n_qubits):
            qc.ry(Parameter(f'θ_{d},{i}'), i)
            qc.rz(Parameter(f'φ_{d},{i}'), i)
        for i in range(n_qubits - 1):
            qc.cnot(i, i+1)
        qc.cnot(n_qubits-1, 0)
    for i in range(n_qubits):
        qc.measure(i, i)
    return qc

vqe_circuit = create_vqe_circuit(5, 3)
plot_quantum_circuit(vqe_circuit)
```
e3ac4906-540c-4f2f-9b57-50a75fb44cc6


yuyamashiro:
The draw function is not a core feature for running quantum calculations, so I am a bit hesitant to add matplotlib as a dependency. For example, Qiskit does not include matplotlib as a dependency, and it requires users to explicitly install it when using the draw function.

However, I think it’s acceptable for now to include it as a dependency and address any issues that arise later. Let’s go with that approach.

yuyamashiro:
- I added the ability to draw the Operator class.
- Introduced the decompose_level argument, allowing operators grouped within the Operator class to be decomposed and represented.
- Added variable labels to the QuantumCircuit class, which are now used when drawing.
- In the case of QAOA, I made it possible to associate with classical bits, making the display more intuitive.
- Refactored the drawer module and added comments.

yuyamashiro (approved):

PR #72, opened 2024-09-09
Feature/param map

Jacomichi:
# change
- Add `param_map` attribute for Qiskit and QuriParts transpiler.
- change the `get_parameters` in `circuit` from `set` to `dict` and `list` to maintain the order in which they were added.
- The check of whether the coefficient is 0 was converted to a distance from 1e-15.

### DIFF
--- a/qamomile/core/converters/qaoa.py
+++ b/qamomile/core/converters/qaoa.py
@@ -30,12 +30,17 @@ and decoding functionalities.
 """
 
 import typing as typ
+import math
 import jijmodeling_transpiler.core as jmt
 import qamomile.core.bitssample as qm_bs
 import qamomile.core.circuit as qm_c
 import qamomile.core.operator as qm_o
 from qamomile.core.converters.converter import QuantumConverter
 
+
+EPSILON = 1e-15
+def is_close_zero(value: float) -> bool:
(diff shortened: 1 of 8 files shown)


qamomile/core/converters/qaoa.py
```diff
@@ -30,13 +30,18 @@
...
+import math
 import jijmodeling_transpiler.core as jmt
 import qamomile.core.bitssample as qm_bs
 import qamomile.core.circuit as qm_c
 import qamomile.core.operator as qm_o
 from qamomile.core.converters.converter import QuantumConverter
 
 
+EPSILON = 1e-15
+def is_close_zero(value: float) -> bool:
```
yuyamashiro, line 42: @Jacomichi 
The dependency where the `is_close_zero function`, which is used not only in qaoa but also in other converters, is written in qaoa and called from qrao and others, may cause unintended bugs related to import order. Therefore, please prepare `converters/utils.py` or similar and write the function there to be imported.

Jacomichi: I have created `converter/utils.py` and moved `is_close_zero`

yuyamashiro (approved):

PR #81, opened 2024-10-08
adding graph_partitioning_tutorial

bluest93:
## Change
Added a notebook of the graph partitioning tutorial

## Description
1. Created a notebook on the usage Qamomile library to solve the graph partitioning problem
2. Created a folder to save the figures used in the notebook above
3. Updated the index.md and _toc.yml to include this new tutorial page


yuyamashiro (changes requested):
Please check the following three change requests and let me know when you've made the changes.

## Use `plot_quantum_circuit` provided by qamomile

Ex.
```python
from qamomile.core.circuit.drawer import plot_quantum_circuit
plot_quantum_circuit(qaoa_circuit)
```

Once you draw a quantum circuit at the Qamomile layer, you don't need to display the same quantum circuit in both Qiskit and Quri-parts.

## Change the Title

The title of the first h1 tag in the notebook will also be used as the heading for the tutorial. The current title, "What is the graph partitioning problem?", is confusing.

"The graph partitioning problem with QAOA" might be a good title. Please think of a good title.

## Differences in experimental conditions between Qiskit and Quri-Parts

If there is no reason, please use the same experimental conditions. Readers may be confused by such minor differences.

Or, please clearly explain the reason for the different experimental conditions, such as "Quri-Parts with Qulacs is faster than Qiskit, so we can increase the number of iterations, so let's increase it and try an experiment."

bluest93:
@yuyamashiro Just made three change requests, please take a look! About the title, I changed as similar as Graph Coloring Problem tutorial.

yuyamashiro (approved):

Issue #76, opened 2024-10-03
The code used to plot the distribution of solutions in the quick start is incorrect.


Jacomichi:
# Description
The code used to plot the distribution of solutions in the quick start is 
```python
energies = []
num_occurrences = []
for sample in sampleset:
    energies.append(sample.eval.objective)
    num_occurrences.append(sample.num_occurrences)
plt.bar(energies, num_occurrences, width=0.05)
plt.ylabel("Frequency")
plt.xlabel("Energy")
plt.show()
```

However, If there are solutions with the same value for the objective function, they are not summed  up together and calculated.
So, frequency can be incorrect.

Jacomichi:
This is fixed by #86

PR #86, opened 2024-10-11
fixed the distribution in quickstart

bluest93:
# Change
fixed the issue in the distribution plot in `quickstart.ipynb`

# Description 
#76


Jacomichi (changes requested):
I have rewritten your code (below).
Changes
- dictionary name
- xlabel name
- `plt.bar` part

Can you change?
```python
from collections import defaultdict
import matplotlib.pyplot as plt

sampleset = qaoa_converter.decode(qk_transpiler, job_result.data["meas"])

# Initialize a dictionary to accumulate occurrences for each energy value
frequencies = defaultdict(int)

# Define the precision to which you want to round the energy values
for sample in sampleset:
    energy = round(sample.eval.objective, ndigits = 3)  
    frequencies[energy] += sample.num_occurrences

plt.bar(frequencies.keys(), frequencies.values(), width=0.05)
plt.xlabel('Objective')
plt.ylabel('Frequency')
plt.show()
```

Jacomichi:
Can you also change #87 ?

bluest93:
Updated the code

Jacomichi (approved):
LGTM

Issue #92, opened 2024-10-17
Allow angle argument as type `float`  in all the `Parametric` gates


bluest93:
# Description
For the current situation, the angle argument has to be in type `ParameterExpression`. Perhaps it would be more user-friendly if type `float` is also allowed. 

# Objective
For all the parametric gates inside class `QuantumCircuit`, type `float` as angle argument should be allowed.
We can add a further check in all the parametric gates as, 
````
if isinstance(angle, float):
   angle = Parameter(angle)
````

# Issue
The issue was found when we append two circuits into one, where the 2nd circuit contains a parametric gate with type `float` as angle argument. 
````
circuit1 = qm.circuit.QuantumCircuit(2)
circuit1.h(0)
circuit1.h(1)

circuit2 = qm.circuit.QuantumCircuit(2)
circuit2.cx(0, 1)
circuit2.rx(0.3, 1)
circuit1.append(circuit2)

plot_quantum_circuit(circuit1)
````

When`plot_quantum_circuit` is called, it eventually arrives to getting the variable, [parameters_name](https://github.com/Jij-Inc/Qamomile/blob/1486c402549144db736f7f749f03d047d565b5c6/qamomile/core/circuit/drawer.py#L270C7-L270C77). Error pops out as 
````
AttributeError: 'float' object has no attribute 'get_parameters'
````
because only type `ParameterExpression` can be obtained.

**Comment by yuyamashiro** [comment]
@bluest93 
Thank you for the report and your suggestion. This is indeed a very good idea that would make the library more user-friendly. I agree with your proposal to accept both ParameterExpression and float types.
Could you please create a Pull Request with the following changes:

Modify all parametric gates within the QuantumCircuit class to allow float type as an angle argument.
Add the following check in each parametric gate:

```
if isinstance(angle, float):
    angle = Parameter(angle)
```


This will ensure that angles specified as floats are internally converted to ParameterExpression, maintaining compatibility with existing processes.
Please update the relevant test cases to ensure that float inputs are correctly processed.
Update the documentation to explain this new feature.

This Pull Request should resolve the issue when appending two circuits, where the second circuit contains a parametric gate with a float type angle argument. It should also prevent the error when calling plot_quantum_circuit.
Once you've created the Pull Request, we'll be happy to review it. Thank you for your contribution to improving the library!

bluest93:
@yuyamashiro Don't know if this is a good test by adding a new circuit, what do you think?
````
[...]

PR #95, opened 2024-10-18
allowed float type in all the parametric gates

bluest93:
# Change
allowed float type as angle argument in all the parametric gates

# Description 
1. added the following code in all the types inside `ParametricSingleQubitGate` and `ParametricTwoQubitGate` in `circuit.py`
 ```
   if isinstance(angle, float):
      angle = Parameter(angle)
```
2. added a corresponding test in `test_circuit.py` 

# Related issue 
#92

### DIFF
--- a/qamomile/core/circuit/circuit.py
+++ b/qamomile/core/circuit/circuit.py
@@ -299,9 +299,12 @@ class QuantumCircuit:
             \end{bmatrix}
 
         Args:
-            angle (ParameterExpression): The angle parameter for the gate.
+            angle (ParameterExpression/float): The angle parameter for the gate.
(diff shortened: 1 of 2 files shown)


yuyamashiro (changes requested):
The Parameter class is a class that represents a named parameter, such as θ. Please check the Parameter constructor (`__init__`) carefully.
Therefore, in order to accept the float in this case, you need to use the Value class that inherits ParameterExpression.
When using functions or classes that you are unfamiliar with, refer to the definition and make sure the interface and type are as intended by the implementer before using them.

yuyamashiro (approved):

PR #97, opened 2024-10-24
Support QuTiP's Hamiltonian

bluest93:
# Change 
Make a new transpiler to support QuTiP's Hamiltonian 

# Description 
1. Created a qutip directory inside qamomile
2. Created `transpiler.py`, `exception.py` and `__inti__.py` under qutip directory
3. Created a qutip directory inside tests
4. Created `test_qutip_transpiler.py` under qutip directory


tests/qutip/test_qutip_transpiler.py
```diff
@@ -0,0 +1,22 @@
...
+    qutip_hamiltonian = transpiler.transpile_hamiltonian(hamiltonian)
+
+    assert isinstance(qutip_hamiltonian, Qobj)
+    assert qutip_hamiltonian.shape == (2**hamiltonian.num_qubits, 2**hamiltonian.num_qubits)
+    
+    hamiltonian2 = qm_o.Hamiltonian()
+    hamiltonian2 += qm_o.X(0) * qm_o.Y(1) + qm_o.Z(0) * qm_o.X(1)
+    qutip_hamiltonian2 = transpiler.transpile_hamiltonian(hamiltonian2)
+    assert isinstance(qutip_hamiltonian2, Qobj)
+    assert qutip_hamiltonian.shape == (2**hamiltonian2.num_qubits, 2**hamiltonian2.num_qubits)
```
Jacomichi, line 22: It is better to create Qutip `Qobj` by your hands and compare it and `Qobj` from Qamoile`

In this case, for example,
```python
import qutip as qt

H = qt.tensor([qt.sigmax(),qt.sigmay()]) + qt.tensor([qt.sigmaz(),qt.sigmax()])

assert H == qutip_hamiltonian2
```

Jacomichi (changes requested):
I have left the comment

qamomile/qutip/transpiler.py
```diff
@@ -0,0 +1,76 @@
...
+                        op_list[pauli.index] = sigmaz()
+                    case _:
+                        raise NotImplementedError("Only Pauli X, Y, and Z are supported")
+            H += coeff* tensor(op_list)
+        return Qobj(H)
+
+    def transpile_circuit(
+        self
+    ) -> None:
+        pass
```
Jacomichi, line 72: I think it is better to raise error because we will not implement this function.

qamomile/qutip/transpiler.py
```diff
@@ -0,0 +1,76 @@
...
+    def convert_result(
+        self
+    ) -> None:
+        pass
```
Jacomichi, line 76: This function also better raises error.

Jacomichi (changes requested):
I left the comment about the error raise.

qamomile/qutip/transpiler.py
bluest93: Should I raise an error message like "transpile_circuit function is not supported by the QuTiPTranspiler class"?

qamomile/qutip/transpiler.py
```diff
@@ -0,0 +1,80 @@
...
+                        op_list[pauli.index] = sigmaz()
+                    case _:
+                        raise NotImplementedError("Only Pauli X, Y, and Z are supported")
+            H += coeff* tensor(op_list)
+        return Qobj(H)
+
+    def transpile_circuit(
+        self
+    ) -> None:
+        raise NotImplementedError(
```
Jacomichi, line 72: How about using `QamomileQuTiPTranspileError`?

@yuyamashiro In this case, which error do you think is appropriate?

Jacomichi (changes requested):
I have left the comment.

qamomile/qutip/transpiler.py
yuyamashiro: NotImplementedError

yuyamashiro (approved):

Jacomichi (approved):
LGTM

PR #99, opened 2024-11-06
add a local search function

bluest93:
# Change
Implemented the simple local search algorithm 

# Description 
under `core/post_process`
1. Added `local_search.py` 
- Implements a new class `IsingMatrix` that represents the Ising model in matrix form (np.array)
  - Converts an `IsingModel` into matrix form with quadratic and linear components
  - Calculates the energy difference when flipping a bit in the Ising state
- Implements a new class `LocalSearch` that perform local search algorithms
   - Runs the local search algorithm until convergence or until a maximum number of iterations
   - Performs first-improvement local search 
   - Performs best-improvement local search
   - Decodes the result into a `jijmodeling SampleSet`
  
2. Added `__inti__.py` 
3. Added tutorial notebook `local_search.ipynb`

under `tests/core/post_process`
1. Added a test file `test_local_search.py`


bluest93:
For the `best_improvement` and `first_improvement`, I choose to return the state instead of index because I think it is easier to understand

bluest93:
Sorry I found out there is a bug for the result when using` best_improvement`

qamomile/core/post_process/local_search.py
```diff
@@ -0,0 +1,108 @@
...
+        self.ising = converter.ising_encode()
+        
+    def decode(self, result) -> jm.experimental.SampleSet:
+        sample = BitsSample(1,np.where(result == -1, 0, 1).tolist())
+        sample_set = BitsSampleSet(bitarrays=[sample])
+        decoded_sampleset=self.converter.decode_bits_to_sampleset(sample_set)
+
+        return decoded_sampleset
+
+    def to_ising_matrix(self, ising: IsingModel) -> IsingMatrix:
```
Jacomichi, line 43: This function should be class method of `Ising Matrix`

qamomile/core/post_process/local_search.py
```diff
@@ -0,0 +1,108 @@
...
+            quad_matrix[i, j] = value
+            quad_matrix[j, i] = value
+
+        linear_vector = np.zeros(size)
+        for i, value in ising.linear.items():
+            linear_vector[i] = value
+
+        return IsingMatrix(quad=quad_matrix, linear=linear_vector)
+
+    def calc_E_diff(self, ising: IsingMatrix, state: np.ndarray, l: int) -> float:
```
Jacomichi, line 56: This should be method of `IsingMatrix`

qamomile/core/post_process/local_search.py
```diff
@@ -0,0 +1,108 @@
...
+        
+    def decode(self, result) -> jm.experimental.SampleSet:
+        sample = BitsSample(1,np.where(result == -1, 0, 1).tolist())
+        sample_set = BitsSampleSet(bitarrays=[sample])
+        decoded_sampleset=self.converter.decode_bits_to_sampleset(sample_set)
+
+        return decoded_sampleset
+
+    def to_ising_matrix(self, ising: IsingModel) -> IsingMatrix:
+        size = max(max(i, j) for i, j in ising.quad.keys()) + 1
```
Jacomichi, line 44: `IsingModel` has `num_bits` to do this
https://github.com/Jij-Inc/Qamomile/blob/fb72a1325ed8cc42e759c4785bad9b48925f0b33/qamomile/core/ising_qubo.py#L12-L17

qamomile/core/post_process/local_search.py
```diff
@@ -0,0 +1,108 @@
...
+        self.quad = quad
+        self.linear = linear
+
+class LocalSearch:
+    def __init__(self, converter: QuantumConverter):
+        self.converter = converter
+        self.ising = converter.ising_encode()
+        
+    def decode(self, result) -> jm.experimental.SampleSet:
+        sample = BitsSample(1,np.where(result == -1, 0, 1).tolist())
```
Jacomichi, line 37: I think $-1$ is mapped to $1$ because we use $Z\ket{1} = (-1)\ket{1}$

qamomile/core/post_process/local_search.py
```diff
@@ -0,0 +1,108 @@
...
+Key Features:
+- Implementation of first- and best-improvement local search algorithms
+- Construction of cost Hamiltonians for the Ising model
+- Conversion of Ising interactions from dictionary format to matrix form
+"""
+
+from qamomile.core.ising_qubo import IsingModel
+from qamomile.core.converters.converter import QuantumConverter
+import qamomile.core.operator as qm_o
+from qamomile.core.converters.utils import is_close_zero
```
Jacomichi, line 20: If you do not use them, do not import

qamomile/core/post_process/local_search.py
```diff
@@ -0,0 +1,108 @@
...
+        decoded_sampleset = self.decode(result)
+        return decoded_sampleset
+
+    def _run_local_search(self, method: Callable, initial_state: np.array, max_iter: int) -> np.ndarray:
+        current_state = initial_state.copy()
+        ising_matrix = self.to_ising_matrix(self.ising)
+        N = len(current_state)
+        counter = 0
+
+        while counter < max_iter:
```
Jacomichi, line 98: You should be able to express the case where you want to converge.
`max_iter` should be optional arguments.

qamomile/core/post_process/local_search.py
Jacomichi: You also used same code in the tutorial. Please fix it also.

Jacomichi (changes requested):
Please write the test for all the functionality

qamomile/core/post_process/local_search.py
bluest93: Do you mean it should be inside class `IsingMatrix`?

qamomile/core/post_process/local_search.py
bluest93: I am using the definition of $s_i = 2x_i -1 $. Also, I tried with `np.where(result == -1, 1, 0)`, the decoded result looks strange

qamomile/core/post_process/local_search.py
```diff
@@ -0,0 +1,119 @@
...
+
+from qamomile.core.ising_qubo import IsingModel
+from qamomile.core.converters.converter import QuantumConverter
+from qamomile.core.bitssample import BitsSample, BitsSampleSet
+import numpy as np
+import jijmodeling as jm
+from typing import Callable
+
+class IsingMatrix:
+    def __init__(self, quad: np.array = None, linear: np.array = None):
```
yuyamashiro, line 25: np.array is not type. You have to use `np.ndarray`.
image


If you are using vscode, installing the python extension will highlight these kinds of basic mistakes in red. Use the editor's functions to reduce these kinds of writing mistakes.

Python will still work even if the type hints are incorrect, so you need to use the editor's functions for these parts.

Also, in terms of type, None is the initial value, so the type is not np.ndarray in the first place. It needs to be wrapped in Optional.

```python
import typing as typ

...(quad: typ.Optional[np.ndarray], ...
```

It needs to be like this.

qamomile/core/post_process/local_search.py
```diff
@@ -0,0 +1,119 @@
...
+        delta_E = np.array([ising_matrix.calc_E_diff(current_state, i) for i in range(N)])
+        best_index = np.argmin(delta_E)
+        best_delta_E = delta_E[best_index]
+        
+        if best_delta_E < 0:
+            current_state[best_index] = -current_state[best_index]
+        
+        return current_state
+    
+    def run(self, initial_state: np.array, max_iter: int = -1, local_search_method: str = "best_improvement") -> jm.experimental.SampleSet:
```
yuyamashiro, line 89: You have to use `np.ndarray` instead of `np.array`.

qamomile/core/post_process/local_search.py
```diff
@@ -0,0 +1,119 @@
...
+        }
+        if local_search_method not in method_map:
+            raise ValueError(f"Invalid local_search_method: {local_search_method}. Choose from {list(method_map.keys())}.")
+
+        method = method_map[local_search_method]
+        result = self._run_local_search(method, initial_state, max_iter)
+        decoded_sampleset = self.decode(result)
+        return decoded_sampleset
+
+    def _run_local_search(self, method: Callable, initial_state: np.array, max_iter: int) -> np.ndarray:
```
yuyamashiro, line 102: use `np.ndarray`

yuyamashiro:
use ruff for format.

Ex.
```
ruff format qamomile/core/post_process/local_search.py
```

If you use ruff, please do not format all of the code for the Qamomile project in this Pull Request, as there will be too many changes and the PR will not be able to be reviewed.

yuyamashiro (changes requested):

bluest93:
@Jacomichi 

> Please write the test for all the functionality

I have a confusion for writing a test. For example, the function of `run` contains other functions `_run_local_searc`, which contains `first_improvement` and `best_improvement`, and `decode`.  I guess I need to check all the functions above?

yuyamashiro (changes requested):
@bluest93 
Please write in the docstring of each method what the function of each method is and how it is intended to be used.
In addition to checking the type, when testing, make sure that it works properly as a local search.

bluest93:
@yuyamashiro For the [test_local_search.py](https://github.com/Jij-Inc/Qamomile/pull/99/commits/e93a5203ee692a2691a844261c4a39f1900c24f8), I have added an assertion to check whether the energy after the local search has decreased or is equal to the original energy.

yuyamashiro (approved):
There is not enough testing and its reliability is still low, but since it is not used in core parts, we will merge it in as is for now.
Also, for local search, the algorithm itself can be improved, such as by efficiently updating it using second-order differences instead of recalculating the energy difference, but we will include these after merging as well.

tests/core/post_process/test_local_search.py
```diff
@@ -0,0 +1,131 @@
...
+    assert ising_matrix.linear.shape == (size,)
+
+
+def test_calc_E_diff(ising_matrix, qaoa_converter):
+    qaoa_encoded = qaoa_converter.ising_encode()
+    ising_matrix.to_ising_matrix(qaoa_encoded)
+    size = qaoa_encoded.num_bits()
+    state = np.random.choice([-1, 1], size=size)
+    random_index = np.random.randint(0, size)
+    delta_E = ising_matrix.calc_E_diff(state, random_index)
```
Jacomichi, line 50: In testing this function, please check that it correctly matches problems that can be calculated by hand, such as 4-bit cases, with some problems.

tests/core/post_process/test_local_search.py
```diff
@@ -0,0 +1,131 @@
...
+    mock_method.reset_mock()
+    final_state_2 = local_search_instance._run_local_search(
+        mock_method, state, max_iter=3
+    )
+    assert mock_method.call_count == 3
+    assert len(state) == len(final_state_2)
+    assert all(val in [-1, 1] for val in final_state_2)
+
+
+def test_best_improvement(local_search_instance, ising_matrix, qaoa_converter):
```
Jacomichi, line 96: For these two tests (`test_best_improvement` and `test_first_improvement`), please test by creating a small problem to see if the correct expected index is returned.

tests/core/post_process/test_local_search.py
```diff
@@ -0,0 +1,131 @@
...
+def test_run(local_search_instance, qaoa_converter):
+    size = qaoa_converter.ising_encode().num_bits()
+    state = np.random.choice([-1, 1], size=size)
+
+    invalid_method = "non_existent_method"
+    with pytest.raises(ValueError):
+        local_search_instance.run(state, local_search_method=invalid_method)
+
+
+def test_run_local_search(local_search_instance, qaoa_converter):
```
Jacomichi, line 64: When maxiter=1, please check that you have obtained the expected state.
When maxiter = -1 for small problem, please check you can obtain the solution you expected.

tests/core/post_process/test_local_search.py
```diff
@@ -0,0 +1,131 @@
...
+
+    assert len(state) == len(new_state)
+    assert all(val in [-1, 1] for val in new_state)
+    assert E2 <= E1
+
+
+def test_decode(local_search_instance, qaoa_converter):
+    size = qaoa_converter.ising_encode().num_bits()
+    state = np.random.choice([-1, 1], size=size)
+    sampleset = local_search_instance.decode(state)
```
Jacomichi, line 129: Check the original state and expected state in the sampleset is same or not and objective value is what you expect or not

docs/tutorial/local_search.ipynb
```diff
@@ -0,0 +1,497 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "4402d3c8-5632-40be-8e4e-523648aea7ae",
+   "metadata": {},
+   "source": [
+    "# Local Search Algorithm in Qamomile\n",
```
Jacomichi, line 8: Local search is intended to be used as a post-processing step, so it should be a tutorial that explains how to use it in that way.
For example, the problem in the QRAO tutorial should work with VQE up to about 24 nodes, so I think it would be good to apply local search to the state obtained by Pauli Rounding.
https://github.com/Jij-Inc/Qamomile/blob/main/docs/tutorial/qrao_tutorial.ipynb

Jacomichi (changes requested):
I left the comments.

docs/tutorial/local_search.ipynb
bluest93: Maybe this should belong to a new PR?

Jacomichi: If so, could you remove the current version of the tutorial?

bluest93: Done

Jacomichi (approved):

Issue #101, opened 2024-11-13
The instance data generation in the maxcut tutorial is incorrect.


Jacomichi:
# Description
Although `nx.to_numpy_array(G)` (https://networkx.org/documentation/stable/reference/generated/networkx.convert_matrix.to_numpy_array.html) is used to generate the weight matrix, the adjacency matrix generated is not in node number order.

If you want to sort it in node number order, you should do it like code below. However, in this case, since all the weights are 1, it seems more natural not to use the weight matrix.
```python
nx.to_numpy_array(graph, nodelist=list(range(N)))
```

Jacomichi:
This is fixed by #102

PR #102, opened 2024-11-13
Remove the weight in the graph

ahao27:
## Change

- Removed weight matrices and weight parameters required when creating instances, as they are unnecessary for this example.
- Updated instance data generation in the MaxCut tutorial to address the incorrect usage reported in Issue #101.

## Desciption
This PR addresses issue #101：
- The code now uses `nx.to_numpy_array(G, nodelist=list(range(num_nodes)))` instead of `nx.to_numpy_array(G)`.
- Since all weights in this example are 1, weight parameters are removed from the example to simplify the code.


Jacomichi (changes requested):
I think you can erase this $w_{ij}$ in https://github.com/Jij-Inc/Qamomile/blob/1d568548827e207a6f9a2cd1bb95db1efd4bd2ab/docs/tutorial/maxcut.ipynb#L135
Moreover, you can change the problem as maximization for explanation.

docs/tutorial/maxcut.ipynb
```diff
@@ -132,7 +132,7 @@
     "The Max-Cut problem can be formulated with the following equation:\n",
```
Jacomichi, line 132: It might be better to first show the model with the maximization problem, and then explain that it is turned into a minimization problem by multiplying by -1.

Jacomichi (changes requested):
Can you change explanation a little bit?

Jacomichi (approved):

PR #113, opened 2025-02-19
Introduce to Ising Hamiltonian normalization

Jacomichi:
# Change
I introduce the Ising Hamiltonian normalization to `IsingModel`.
I have implemented two types of the normalization called, `rms` and `abs_max`.
The detail explanation is in the docstring of `normalize_by_abs_max` and `normalize_by_rms` in `ising_qubo,py`.
However, I have chosen that we did not normalize Ising Hamiltonian as default because changing the default value breaks the behavior of the old version.

# Usage
You can access this option through `QuantumConverter`.
This is the example code
```python
normalize_ising = "rms"
# normalize_ising = "abs_max"
qaoa_converter = qm.qaoa.QAOAConverter(compiled_instance, normalize_ising=normalize_ising)
```

# Benchmark
I have done two benchmarks.
So far, I feel that `abs_max` works better than the other method.

### coefficient from normal distribution
I define the problem as follows:
```python
N = 16
J = np.random.randn(N,N)
J = (J + J.T) / 2
```
normal_dist

### Biased some coefficient
```python
N = 16
J = np.random.uniform(0, 10, (N, N))
J = (J + J.T) / 2
J[0,1] = 100
J[1,0] = 100
```
biased_dist

### DIFF
--- a/qamomile/core/converters/converter.py
+++ b/qamomile/core/converters/converter.py
@@ -75,6 +75,7 @@ class QuantumConverter(abc.ABC):
         compiled_instance,
         relax_method: jmt.pubo.RelaxationMethod = jmt.pubo.RelaxationMethod.AugmentedLagrangian,
         normalize_model: bool = False,
+        normalize_ising: typ.Optional[str] = None,
     ):
         """
         Initialize the QuantumConverter.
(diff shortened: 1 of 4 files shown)


qamomile/core/converters/converter.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         compiled_instance,
         relax_method: jmt.pubo.RelaxationMethod = jmt.pubo.RelaxationMethod.AugmentedLagrangian,
         normalize_model: bool = False,
+        normalize_ising: typ.Optional[str] = None,
```
yuyamashiro, line 78: change to `typ.Optional[typ.Literal["abs_max", "rms"]]`

yuyamashiro (changes requested):

yuyamashiro (approved):

PR #149, opened 2025-06-19
Fix jupyter-book-publish

ksk-jij:
# Description
Fix jupyter-book-publish actions. `upload-pages-artifact@3` is used everywhere in the `yml` file but it should be used as last, right before a deployment. Change them except for the last one into `upload-artifact@v4`.


Jacomichi (approved):
Thanks

Issue #199, opened 2025-07-11
Error after ommx support


ksk-jij:
# Description
The following code causes the error shown below. I've confirmed this error also happens after the ommx support was added in version, [625fd24](https://github.com/Jij-Inc/Qamomile/commit/625fd2476dca09c6f4c710ffa7b6ac13b2cd9bf4), which is currently on `main`. Seems like `QuantumConverter.ising_encode` causes this issue.

https://github.com/Jij-Inc/Qamomile/blob/625fd2476dca09c6f4c710ffa7b6ac13b2cd9bf4/qamomile/core/converters/converter.py#L232-L242

Moreover, this error may be rooted in the thing mentioned TODO comment.

```python
import jijmodeling as jm

# Define Knapsack problem
v = jm.Placeholder('v', ndim=1)
N = v.len_at(0, latex="N")
w = jm.Placeholder('w', ndim=1)
W = jm.Placeholder('W')
x = jm.BinaryVar('x', shape=(N,))
i = jm.Element('i', belong_to=(0, N))
problem = jm.Problem('Knapsack', sense=jm.
ProblemSense.MAXIMIZE)
problem += jm.sum(i, v[i]*x[i])
problem += jm.Constraint('weight', jm.sum(i, w[i]*x[i])<=W)
# Prepare instance data
instance_data = {
"v": [12, 7, 19, 5, 11, 3],
"w": [4, 2, 7, 3, 6, 1],
"W": 15}
# create the problem model
compiled_instance = jm.Interpreter(instance_data).eval_problem(problem)
from qamomile.core.converters.qaoa import QAOAConverter
from qamomile.qiskit import QiskitTranspiler

# Build a QAOA ansatz (p = 1)
circuit_ir = QAOAConverter(compiled_instance).get_qaoa_ansatz(p=1)
# Output platform-specific circuits
qc_qiskit = QiskitTranspiler().transpile_circuit(circuit_ir)
```

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[6], line 5
      2 from qamomile.qiskit import QiskitTranspiler
      4 # Build a QAOA ansatz (p = 1)
----> 5 circuit_ir = QAOAConverter(compiled_instance).get_qaoa_ansatz(p=1)
      6 # Output platform-specific circuits
      7 qc_qiskit = QiskitTranspiler().transpile_circuit(circuit_ir)

[...]

PR #200, opened 2025-07-16
Fix error after OMMX support

ksk-jij:
# Description
Fix the error reported in #199. The cause was as mentioned in the following comment.

https://github.com/Jij-Inc/Qamomile/blob/0b19412ca25e83bd81f1e200c9ba9e4e3b9cc51d/qamomile/core/converters/converter.py#L234-L236

This PR adds a test for the error case and updated Qamomile to resolve the added error case.

# Update
Store the converted instance for getting the QUBO problem.

https://github.com/Jij-Inc/Qamomile/blob/bfbc63795792d9ffa918e42c15b2b1fe595ac829/qamomile/core/converters/converter.py#L179-L183

And use the converted instance to get a label.

https://github.com/Jij-Inc/Qamomile/blob/bfbc63795792d9ffa918e42c15b2b1fe595ac829/qamomile/core/converters/converter.py#L236-L245

# Justification
In Qamomile `QuantumConverter.int2varlabel` is only used for `QuantumConverter.update_qubits_label`, which sets its `_qubits_label`.

https://github.com/Jij-Inc/Qamomile/blob/bfbc63795792d9ffa918e42c15b2b1fe595ac829/qamomile/core/circuit/circuit.py#L246-L255

The `_qubits_label` is accessed through the property `qubits_label`, which is only used in `_draw_quantum_circuit` function in `drawer.py`.

https://github.com/Jij-Inc/Qamomile/blob/bfbc63795792d9ffa918e42c15b2b1fe595ac829/qamomile/core/circuit/drawer.py#L178-L187

Thus, I believe that this update, which uses converted instance to make the label, is justified.


Jacomichi (approved):
Approve

Issue #218, opened 2025-10-29
`IsingModel.num_bits` can be wrong


ksk-jij:
# Description
It seems like `IsingModel.num_bits` returns the number of variables, 0-origin, by using `max` function. However, if we use `Conveter.ising_encode`, then `Conveter.instance_to_qubo` and finally `ommx.v1.Instance.to_qubo` is called. **Need to be checked**: It seems like the returned value of `to_qubo` does not always start with 0 or/and is not be sequential. If so, the keys of IsingModel.linear` and `.quad` will not start with 0 or/and be be sequential. Then, the `max` function returns a larger number than it actually is.

# Others
I haven't checked this by running code. So, the first step of this issue will be to check if there are any case that happens.

ksk-jij:
This is fixed in #225 .

Issue #214, opened 2025-10-03
Error in `Converters.decode`


ksk-jij:
# Description
The following error happens in `Converters.decode`:

<details>
<summary>
KeyError
</summary>

```bash
KeyError                                  Traceback (most recent call last)
Cell In[13], [line 1](vscode-notebook-cell:?execution_count=13&line=1)
----> [1](vscode-notebook-cell:?execution_count=13&line=1) qaoa_converter.decode(qk_transpiler, qaoa_counts)

File ~/dev/git/OMMX-OBLIB/.venv/lib/python3.10/site-packages/qamomile/core/converters/converter.py:283, in QuantumConverter.decode(self, transpiler, result)
    268 """
    269 Decode quantum computation results into a SampleSet.
    270 
   (...)
    280     ommx.v1.SampleSet: The decoded results as a SampleSet.
    281 """
    282 bitssampleset = transpiler.convert_result(result)
--> [283](https://file+.vscode-resource.vscode-cdn.net/Users/keisukesato/dev/git/OMMX-OBLIB/docs/en/tutorial/examples/~/dev/git/OMMX-OBLIB/.venv/lib/python3.10/site-packages/qamomile/core/converters/converter.py:283) return self.decode_bits_to_sampleset(bitssampleset)

File ~/dev/git/OMMX-OBLIB/.venv/lib/python3.10/site-packages/qamomile/core/converters/converter.py:308, in QuantumConverter.decode_bits_to_sampleset(self, bitssampleset)
    306 sample = {}
    307 for i, bit in enumerate(bitssample.bits):
--> [308](https://file+.vscode-resource.vscode-cdn.net/Users/keisukesato/dev/git/OMMX-OBLIB/docs/en/tutorial/examples/~/dev/git/OMMX-OBLIB/.venv/lib/python3.10/site-packages/qamomile/core/converters/converter.py:308)     index = ising.ising2qubo_index(i)
    309     sample[index] = bit
    310 state = ommx.v1.State(entries=sample)

File ~/dev/git/OMMX-OBLIB/.venv/lib/python3.10/site-packages/qamomile/core/ising_qubo.py:45, in IsingModel.ising2qubo_index(self, index)
     44 def ising2qubo_index(self, index: int) -> int:
[...]

PR #225, opened 2025-11-06
Introduce reindexing feature to HigherOrderIsing/IsingModel

### LINKED ISSUE #68: Maintain Index Consistency When Using `simplify=True` in `qubo_to_ising`
## Description
Currently, using `simplify=True` in the `qubo_to_ising` function causes an index mismatch between QUBO and Ising models. We need to address this issue to ensure that `simplify=True` can be used safely without compromising index consistency.

## Objectives
1. Maintain index consistency between QUBO and Ising models when using `simplify=True` in some cases. 
The test below is the case.
https://github.com/Jij-Inc/Qamomile/blob/27b4515191c40b5b339b9c9f19b23adc41315203/tests/quri_parts/test_quri_transpiler.py#L179-L194

2. Provide a way for users to easily access the correct indices after simplification.

## Current Behavior
- When `simplify=True` is used in `qubo_to_ising`, the resulting Ising model has different indices from the original QUBO model.
- This mismatch can lead to wrong decode when working with `var_map` or performing other operations that rely on consistent indexing.

## Desired Behavior
- Using `simplify=True` should not break the index correspondence.
- Users should be able to work with simplified models without worrying about index mismatches.

## Related PR
#67

ksk-jij:
# Description
This PR introduces the feature of re-indexing the given coefficients' indices into zero-origin and continuous integers as a post-initialisation, also, as a result, enabling `simplify=True`.

# Background
When coefficients had non-zero or discontinuous indices (e.g., 0, 2, 3, 4, ...), the decoding process with QAOA failed. That was because `num_bits` property assumed the indices started with zero and continuous integers and returned `max(indices)`, which could produce a circuit whose number of qubits are more than actually needed. During the decoding process, Ising indices are accessed as shown in the following code.

https://github.com/Jij-Inc/Qamomile/blob/01402581faf790965440f9cedc87f1c6f63606b3/qamomile/core/converters/converter.py#L307-L309

Obviously, if the indices started from non-zero or were uncontinuous, then `KeyError` is raised.

# Solution in this PR
This PR implements the feature of re-indexing coefficients to use zero-origin continuous indices to `HigherOrderIsing` and updates `num_bits` to return the number of variables, which is the number of bits, instead of `max(indices)`. Those changes also propagate to `IsingModel` as it is a child of `HigherOrderIsing`
Additionally, it adds `original_to_zero_origin_map` variable and its inverse as property. Those ensure the sampled bits are correctly mapped back to the correct places of `ommx.v1.Instance`.

Furthermore, #68 is fixed by this PR by removing the construction of `index_map` during the simplifying process and delegating it to the post-initialisation added by this PR.

# Others
Closes #68, #214 and #218.

### DIFF
--- a/qamomile/core/higher_ising_model.py
+++ b/qamomile/core/higher_ising_model.py
@@ -14,25 +14,25 @@ class HigherIsingModel:
     The zero-origin indices are re-indexed indices starting from 0 and continuous.
     The coefficients are re-indexed to use zero-origin indices in the post-initialisation process.
     Thus, the coefficients attribute uses zero-origin indices.
-    Furthermore, this model has three mappings; index_map, original_to_zero_origin_map and zero_origin_to_original_map
+    Furthermore, this model has three mappings; `index_map`, `original_to_zero_origin_map` and `zero_origin_to_original_map`
     because this model re-indexes the given coefficients to zero-origin indices in the post-initialisation process.
 
-    index_map is a mapping from the original indices to the original indices specified by users or identity.
-    Thus, let's say the given coefficients are {(2, 3): 1.0, (3,): -1.0} and the index_map is {2: 5, 3:7},
-    then the original indices are 2 and 3, and those mapped indices are 5 and 7.
-    This mapping is used in ising2original_index method and QuantumConverter class to label qubits.
-    ising2original_index method is also used in QuantumConverter class to decode the sampled bits into classical solutions.
+    `index_map` is a mapping from the original indices to the original indices specified by users or identity.
+    Let's say the given coefficients are `{(2, 3): 1.0, (3,): -1.0}` and the `index_map` is `{2: 5, 3:7}`,
+    and then the original indices are 2 and 3, and those mapped indices are 5 and 7.
+    This mapping is used in `ising2original_index` method and QuantumConverter class to label qubits.
+    `ising2original_index` method is also used in QuantumConverter class to decode the sampled bits into classical solutions.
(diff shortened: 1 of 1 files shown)


Jacomichi (commented):
I think it is better to emphasize when we will use or new indices responsible in the document.

qamomile/core/higher_ising_model.py
```diff
@@ -7,40 +7,101 @@
...
+    Furthermore, this model has three mappings; index_map, original_to_zero_origin_map and zero_origin_to_original_map
+    because this model re-indexes the given coefficients to zero-origin indices in the post-initialisation process.
+
+    index_map is a mapping from the original indices to the original indices specified by users or identity.
+    Thus, let's say the given coefficients are {(2, 3): 1.0, (3,): -1.0} and the index_map is {2: 5, 3:7},
+    then the original indices are 2 and 3, and those mapped indices are 5 and 7.
+    This mapping is used in ising2original_index method and QuantumConverter class to label qubits.
+    ising2original_index method is also used in QuantumConverter class to decode the sampled bits into classical solutions.
+    Which means this mapping is used only in quantum algorithms.
+    If usr did not have any specific mapping in mind, the identity mapping, in this example case {2: 2, 3: 3}, is used by default.
```
Jacomichi: usr -> user

ksk-jij: Thank you for the confirmation and check! I just fixed it :)

Issue #137, opened 2025-06-05
`quantum_annealing.ipynb` doesn't work


ksk-jij:
# Problem
This also happens to `main` branch.

It is not able to correctly run `quantum_annealing.ipynb`, which is [Using the QuTiPTranspiler in Qamomile](https://jij-inc.github.io/Qamomile/tutorial/usage/quantum_annealing.html). The cell

```python
qt_transpiler = qm_qt.QuTiPTranspiler()
qt_H0= qt_transpiler.transpile_hamiltonian(H0)
```

returns the following error message.

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[5], line 2
      1 qt_transpiler = qm_qt.QuTiPTranspiler()
----> 2 qt_H0= qt_transpiler.transpile_hamiltonian(H0)

File ~/dev/git/Qamomile/qamomile/qutip/transpiler.py:66, in QuTiPTranspiler.transpile_hamiltonian(self, operator)
     64     H += coeff * tensor(op_list)
     65 if operator.constant != 0:
---> 66     H += operator.constant * qeye(2 ** n)
     67 return Qobj(H)

File ~/Library/Caches/pypoetry/virtualenvs/qamomile-SJ_jAwfp-py3.11/lib/python3.11/site-packages/qutip/core/qobj.py:57, in _require_equal_type.<locals>.out(self, other)
     52     if self._dims != other._dims:
     53         msg = (
     54             "incompatible dimensions "
     55             + repr(self.dims) + " and " + repr(other.dims)
     56         )
---> 57         raise ValueError(msg)
     58     return method(self, other)
     59 if other == 0:

ValueError: incompatible dimensions [[2, 2, 2, 2, 2], [2, 2, 2, 2, 2]] and [[32], [32]]
```

ksk-jij:
By #232, this issue was resolved!

PR #126, opened 2025-04-07
start ommx support

yuyamashiro:

PR #232, opened 2025-11-20
fixing annealing tutorial

Rafe-code:
Resolving issue #137 

A fix to the QuTiP transpiler fixed the issue referenced here, but further changes wer enecessary to make this tutorial run. You can no longer iterate through the 'x' values of a sample set as done previously with line 
"for transformed in qaoa_converter.decode_bits_to_sampleset(sample_set).data"

Instead, the samples are iterated over and the bit strings are manually transformed into mappings to the MaxCut nodes using:

transformed_state=[]
for sample in samples: 
    # Start with all zeros
    cut_dict = {node: 0 for node in G.nodes()}
    
    for i, bit_num in enumerate(sample.bits):
        if bit_num == 1:        
            cut_dict[i] = 1
    
    transformed_state.append(cut_dict)

The tutorial now completes and demonstrates the expected outcomes of 4 degnerate states. You can observe the solved MaxCut images and plot of the energy evolution to observe this. Slightly different solutions are found to the original sheet, but this is due to the symmetry in the problem and is okay.

### DIFF
--- a/docs/ja/tutorial/usage/quantum_annealing.ipynb
+++ b/docs/ja/tutorial/usage/quantum_annealing.ipynb
@@ -27,7 +27,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 13,
+   "execution_count": 1,
    "id": "f94226bd-5710-4655-8f73-cc105f3e8f2d",
    "metadata": {},
    "outputs": [],
@@ -55,7 +55,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 14,
+   "execution_count": 2,
    "id": "be3ddbdf-1794-4c7e-96eb-5eb0a3994bf0",
    "metadata": {},
    "outputs": [
@@ -65,10 +65,10 @@
        "$$\\begin{array}{cccc}\\text{Problem:} & \\text{Maxcut} & & \\\\& & \\min \\quad \\displaystyle (-0.5) \\cdot \\sum_{e \\in E} \\left(- s_{e[0]} \\cdot s_{e[1]} + 1\\right) & \\\\\\text{{where}} & & & \\\\& x & 1\\text{-dim binary variable}\\\\\\end{array}$$"
       ],
       "text/plain": [
-       "<jijmodeling.Problem at 0x1e1e20ded30>"
+       "<jijmodeling.Problem at 0x25c1c9c2730>"
       ]
      },
-     "execution_count": 14,
+     "execution_count": 2,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -97,7 +97,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 15,
+   "execution_count": 3,
    "id": "b3b5562e-686c-4412-b4ac-5d482795fba9",
    "metadata": {},
    "outputs": [
@@ -128,7 +128,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 16,
+   "execution_count": 4,
    "id": "9b27d5c1-4bc9-4365-b1ac-dbe48bda2eaa",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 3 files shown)


ksk-jij (changes requested):
Hi @Rafe-code, thank you for the PR again.

I had a look at your comments and the notebook. It seems this PR is not self-contained; the fix for the QuTiP transpiler's error is not included. Could you include that as well? I believe changing `qeye(2 ** n)` to `tensor([qeye(2)] * n)` should resolve the error, but feel free to use a different approach since I didn't look into it deeply and might have overlooked something.

Also, optionally, we have [the same notebook in Japanese](https://github.com/Jij-Inc/Qamomile/blob/main/docs/ja/tutorial/usage/quantum_annealing.ipynb). If you're comfortable updating it (keeping comments in Japanese), that would be great, but feel free to skip it if not.

Looking forward to the update!

docs/ja/tutorial/usage/quantum_annealing.ipynb
```diff
@@ -279,35 +326,61 @@
...
+    "transformed_state=[]\n",
+    "for sample in samples: \n",
+    "    # Start with all zeros\n",
+    "    cut_dict = {node: 0 for node in G.nodes()}\n",
+    "    \n",
+    "    for i, bit_num in enumerate(sample.bits):\n",
+    "        if bit_num == 1:        \n",
+    "            cut_dict[i] = 1\n",
+    "    \n",
+    "    transformed_state.append(cut_dict)\n",
```
ksk-jij: [suggestion] (This suggestion is as same as one in English notebook.)
I'm sorry that I had overlooked this at the last review, but I just realised that we can make the modification simpler utilising [OMMX](https://jij-inc.github.io/ommx/en/introduction.html) functionality (which we started to support from PRs #126 and #198):

```python
sample_set = BitsSampleSet(bitarrays=samples)

bits_sample_set = BitsSampleSet(samples)
ommx_sample_set = qaoa_converter.decode_bits_to_sampleset(bits_sample_set)
```

And then, iterating over the solutions using `solution_id` and using its property should work at the next cell.

```python
for i, sample_id in enumerate(ommx_sample_set.sample_ids):
    solution = ommx_sample_set.get_sample_by_id((sample_id))
    
    edge_colors, node_colors = get_edge_colors(G, solution.state.entries)
    ...
```

Then, we don't need to create `cut_dict` by hands. What do you think of this modification? I believe this should work and also simplify the code a bit. If you have any concerns about that, please let me know.

OMMX library has much richer functionalities. Please have a look at the documents for OMMX functionalities for more details if you are interested.

- [ommx.v1.SampleSet](https://jij-inc.github.io/ommx/python/ommx/autoapi/ommx/v1/index.html#ommx.v1.SampleSet)
- [ommx.v1.Solution](https://jij-inc.github.io/ommx/python/ommx/autoapi/ommx/v1/index.html#ommx.v1.Solution)
- [ommx.v1.State](https://jij-inc.github.io/ommx/python/ommx/autoapi/ommx/v1/solution_pb2/index.html#ommx.v1.solution_pb2.State)

ksk-jij (changes requested):
Hi @Rafe-code, thank you for the update! The QuTiP solution looks good to me as well as the Japanese notebook.

I apologise for overlooking this in the last review, but I found one thing that I prefer to change in both notebooks regarding your bugfix. So, please have a look at the comments below. The last one is just a question, not a suggestion.

Rafe-code:
Ah - this onnx functionality is much better! Thank you - I'll make sure to read the documentation here before future PRs. I've updated the Japanese & English workbooks to use this.

I unpacked the inputs to `tensor` as this function's docustring expects Qobj, rather than a list of Qobj, as flagged by Pylance. This is purely docustring though and has no impact on functionality. I have changed it back to be consistent with other parts of the codebase. Perhaps a future PR could include the docustring input: `def tensor(*args: Qobj | list[Qobj]) -> Qobj:` ? But I have not checked further on this.

ksk-jij:
Hi @Rafe-code,

Thank you for the update and for getting back to me!

> This is purely docustring though and has no impact on functionality. I have changed it back to be consistent with other parts of the codebase. Perhaps a future PR could include the docustring input: def tensor(*args: Qobj | list[Qobj]) -> Qobj: ? But I have not checked further on this.

This is quite true. Thank you for sharing this with me.

PR #244, opened 2025-12-11
Fix the portfolio optimization tutorial

ahao27:
## Description:

This PR updates the FQAOA tutorials in the documentation with the following change:

## Summary of Change

1. Replaced the line to `import ommx_pyscipopt_adapter as OMMXPySCIPOptAdapte`

### DIFF
--- a/docs/en/tutorial/opt_advance/portfolio_optimization.ipynb
+++ b/docs/en/tutorial/opt_advance/portfolio_optimization.ipynb
@@ -1,466 +1,545 @@
 {
- "cells": [
-  {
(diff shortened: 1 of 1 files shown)


ksk-jij (changes requested):
I've left comments. Please have a look. It feels like the result I have figured out to execute the notebook swimmingly just removes the change done in this PR according to  Summary of Change you wrote. If so, I think this PR is not really needed. If not so, I'd appreciate it if you tell me what is the key change of this PR.

ksk-jij:
This is for v0.9.0. Close this issue.

Issue #249, opened 2026-01-22
CM-1: Migrate core → optimization


yuyamashiro:
## Description
Migrate all code from `qamomile/core/` to `qamomile/optimization/` as the core module is scheduled for deletion.

## Dependencies
None

## Tasks
- [ ] Migrate core/converters/ to optimization/converters/
- [ ] Migrate core/ansatz/ to optimization/ansatz/
- [ ] Migrate core/modeler/ to optimization/modeler/
- [ ] Migrate core/post_process/ to optimization/post_process/
- [ ] Update dependencies (fix import statements)
- [ ] Migrate and update tests
- [ ] Delete core directory

## Related
See [Development Roadmap](https://github.com/Jij-Inc/Qamomile/discussions/248) for full context.

ksk-jij:
The current main 0db9931944f7375a30c612dbba20846eb1ca2821 does have old core in optimization.

PR #260, opened 2026-01-28
Remove transpiler.py

ahao27:
## Overview
A new transpiler base has replaced the old transpiler implementation. This PR cleans up the codebase by removing the transpiler.py.

## Changes
- Remove deprecated transpiler implementation and related tests:
  - `qamomile/core/transpiler.py`
- Issue link: https://github.com/Jij-Inc/Qamomile/issues/249


ksk-jij (commented):
[my opinion]
Is there any reason that you kept test files for [tests/qiskit](https://github.com/Jij-Inc/Qamomile/tree/new_qamomile/tests/qiskit)? If I get correctly, those are for the old qiskit transpiler. test files for new transpilers are in [tests/transpiler](https://github.com/Jij-Inc/Qamomile/tree/new_qamomile/tests/transpiler). So, if there are no concerns, we can remove the old qiskit test files as well.

This might be applied to [tests/quri-parts](https://github.com/Jij-Inc/Qamomile/tree/new_qamomile/tests/quri_parts). If they are completely the old transpiler, then we can remove them.

Also, [quri_parts/parameter_converter.py](https://github.com/Jij-Inc/Qamomile/blob/new_qamomile/qamomile/quri_parts/parameter_converter.py) might completely depends on the old design? In that case, even though this is in quri-parts, which we have discussed we would leave, we can remove this. (Sorry for this late check! I didn't really look into each file when we discussed about this.)

Could you look into those points?

PR #262, opened 2026-01-28
Ignore temporary test files for future-available-backends and fix docs

ksk-jij:
# Description
This PR contains two topics.

1. Ignoring unnecessary test files
PR #258 caused the test file for quri-parts (which is not currently supported but planned for future support) to start failing, as this test relied on classes that were removed in PR #258. Additionally, I found test files for other backends that are not yet supported. This PR configures GitHub CI to ignore these test files. I decided not to remove them as they may be useful when we add support for these backends in the new version of Qamomile.

2. Fixing doc bugs
PR #258 changed the steps of transpiling. The docs did not adapt the change. This PR fixes the docs.

### DIFF
--- a/docs/en/transpile/transpile_flow.ipynb
+++ b/docs/en/transpile/transpile_flow.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "c7df5c22",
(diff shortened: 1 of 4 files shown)


Jacomichi (approved):
okay

PR #267, opened 2026-02-05
Remove redundant _DictValueType class

ksk-jij:
## Summary

- Remove the `_DictValueType` placeholder class from `value.py`
- Update `DictValue.type` property to use `DictType` from `primitives.py`
- Make `DictType.key_type` and `value_type` optional with default `None`

## Why is `_DictValueType` unnecessary?

The `_DictValueType` class was a private placeholder that existed solely to provide a type interface for `DictValue.type`. However, `DictType` already exists in `primitives.py` with full `ValueType` inheritance and proper classical type semantics:

| Class | Location | Inheritance |
|-------|----------|-------------|
| `_DictValueType` | `value.py` | None (ad-hoc interface) |
| `DictType` | `primitives.py` | `ClassicalTypeMixin`, `ValueType` ✓ |

The `_DictValueType`:
- Did not inherit from `ValueType`, making it inconsistent with the type system
- Duplicated the `is_classical()`, `is_quantum()`, and `label()` methods
- Was likely added before `DictType` existed, or they were added at different times without consolidation

Using `DictType` directly ensures type system consistency and eliminates code duplication.

## Changes

1. **`value.py`**: 
   - Remove `_DictValueType` class
   - Import `DictType` from `.types`
   - Update `DictValue.type` to return `DictType()`

2. **`primitives.py`**:
   - Make `key_type` and `value_type` optional (default `None`) to support generic Dict without specific type parameters
   - Update `label()` to return `"Dict"` when types are not specified

## Test plan

- [x] All circuit tests pass (`pytest tests/circuit/`)
- [x] Ruff check passes

### DIFF
--- a/qamomile/circuit/ir/types/primitives.py
+++ b/qamomile/circuit/ir/types/primitives.py
@@ -107,12 +107,15 @@ class DictType(ClassicalTypeMixin, ValueType):
 
     Unlike simple types, DictType stores the key and value types,
     so equality and hashing depend on those types.
+    When key_type and value_type are None, represents a generic Dict type.
     """
 
-    key_type: "ValueType"
-    value_type: "ValueType"
+    key_type: "ValueType | None" = None
+    value_type: "ValueType | None" = None
 
     def label(self) -> str:
+        if self.key_type is None or self.value_type is None:
+            return "Dict"
         return f"Dict[{self.key_type.label()}, {self.value_type.label()}]"
 
     def __eq__(self, other: object) -> bool:
--- a/qamomile/circuit/ir/value.py
+++ b/qamomile/circuit/ir/value.py
@@ -13,7 +13,7 @@ import dataclasses
 import typing
 import uuid
 
-from .types import ValueType
+from .types import DictType, ValueType
 
 T = typing.TypeVar("T", bound=ValueType)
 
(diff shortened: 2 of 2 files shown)


ksk-jij:
## Response to Copilot's Review Comments

Thank you for the review. After investigating the codebase, I've decided not to make the suggested changes for the following reasons:

### Current Usage Patterns

`DictType` is only used in two patterns in the codebase:
1. `DictType(key_type=X, value_type=Y)` - both specified (in `func_to_block.py:110`)
2. `DictType()` - both `None` (in `DictValue.type`)

The case where only one parameter is `None` **does not occur** in the current codebase.

### Regarding `__eq__` (Comment 1)

The concern about `None == ValueTypeInstance` raising an error is not accurate. In Python:
- When `ValueType.__eq__` returns `NotImplemented` for `None`, Python falls back to identity comparison
- The result is simply `False`, not an error

### Regarding `label()` with `or` (Comment 2)

Since only-one-`None` cases don't occur, using `or` vs `and` makes no practical difference. Adding defensive code for `Dict[Any, X]` or `Dict[X, Any]` labels would be dead code.

### YAGNI Principle

Following the YAGNI (You Aren't Gonna Need It) principle, I prefer not to add complexity for cases that don't exist. If we need to support partial type specifications in the future, we can address it then.

PR #269, opened 2026-02-06
Unify UInt arithmetic to accept Float operands

ksk-jij:
## Summary
- All `UInt` arithmetic methods (`__add__`, `__sub__`, `__mul__`, `__floordiv__`, `__pow__` and their reflected variants) now accept `float`/`Float` operands via `_coerce`, returning `Float` when the other operand is `Float`
- Removes inconsistency where only `__mul__` and `__truediv__` handled float operands
- Refactors `__mul__` to use `_coerce` instead of manual `isinstance` + `Value` construction
- Extracts `_result_for()` helper to eliminate repeated inline ternary for result type selection
- Sorts imports alphabetically

## Test plan
- [x] `uv run ruff check` passes
- [x] All 165 existing circuit tests pass
- [x] 30 new tests verify `UInt op float` returns `Float` and `UInt op int` returns `UInt` for all operators

### DIFF
--- a/qamomile/circuit/frontend/handle/primitives.py
+++ b/qamomile/circuit/frontend/handle/primitives.py
@@ -2,15 +2,15 @@ from __future__ import annotations
 
 import dataclasses
 
+from qamomile.circuit.ir.operation.arithmetic_operations import BinOpKind, CompOpKind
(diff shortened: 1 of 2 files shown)


ksk-jij:
For the time being

PR #331, opened 2026-04-13
Fix documentation search URL doubling on ReadTheDocs

ahao27:
## Summary
- Fix `BASE_URL` in `docs/build.sh` to use path-only value (`/latest/`) instead of full URL, which caused MyST theme to double the origin when constructing search URLs
- Search was broken with error: "Search is not enabled for this site" because `myst.search.json` was fetched from an invalid doubled URL like `https://host/https:/host/latest/en/myst.search.json` and  `https://host/https:/host/latest/jp/myst.search.json`

## Test plan
- [ ] Verify search works on RTD after deployment (both EN and JP)
- [ ] Confirm `myst.search.json` is fetched from `https://jij-inc-qamomile.readthedocs-hosted.com/latest/{en,ja}/myst.search.json`
- [ ] Locally tested by simulating RTD env vars: `READTHEDOCS=True READTHEDOCS_CANONICAL_URL="https://..." ./docs/build.sh build-en` — confirmed `BASE_URL=/latest/en` in generated HTML


docs/build.sh
```diff
@@ -113,7 +113,14 @@ build_lang() {
...
+        # Strip origin, keep only path (MyST expects path-only BASE_URL)
+        local base_url stripped
+        stripped="${READTHEDOCS_CANONICAL_URL#https://}"
+        stripped="${stripped#http://}"
+        base_url="/${stripped#*/}"
+        # Handle domain-only case (no path after host)
+        [[ "$base_url" == "/${stripped}" ]] && base_url=""
+        base_url="${base_url%/}/${lang}"
         info "Read the Docs detected. Using BASE_URL=${base_url}"
         BASE_URL="$base_url" uv run jupyter-book build --html
```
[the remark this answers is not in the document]
ahao27: Now `is_rtd && [[ -n "${READTHEDOCS_CANONICAL_URL:-}" ]]` is checked together, and a warning is emitted when RTD is detected but `READTHEDOCS_CANONICAL_URL` is unset or empty.

docs/build.sh
```diff
@@ -112,11 +112,21 @@ build_lang() {
...
```
ksk-jij: [question]
Let me know if I read this wrongly. I'll write what I understand and expect based on these.

OBJECTIVE: According to https://mystmd.org/guide/deployment, the objective here is to set `BASE_URL` as `/latest/{lang}`.

THE CURRENT CODE:
Once the first two lines are executed, we will have `READTHEDOCS_CANONICAL_URL` without `https://` or `http://`. Then, the third line removes the any letters before the first `/` and sets it to `base_url`.

So, based on what https://docs.readthedocs.com/platform/stable/canonical-urls.html says,

0. `READTHEDOCS_CANONICAL_URL` is `https://jij....readthedocs.io/en/latest/`
1. `stripped` becomes `jij....readthedocs.io/en/latest/` by removing `https://` (line 118)
2. `stripped` does not change since there is no `http://`. (line 119)
3. `base_url` becomes `/en/latest/`  (line 120)
4. `base_url` becomes `/en/latest/en` (line 123)

will be BASE_URL? If my understanding is correct, I think we need to update this.

IDEA:
According to https://docs.readthedocs.com/platform/stable/reference/environment-variables.html#envvar-READTHEDOCS_VERSION, seems like we could get the version from READTHEDOCS_VERSION variable. If this is correct, we can make `base_url` as `"${READTHEDOCS_VERSION}/${lang}"`?

docs/build.sh
```diff
@@ -112,11 +112,21 @@ build_lang() {
...
-    if is_rtd; then
-        local base_url="${READTHEDOCS_CANONICAL_URL%/}/${lang}"
+    if is_rtd && [[ -n "${READTHEDOCS_CANONICAL_URL:-}" ]]; then
+        # Strip origin, keep only path (MyST expects path-only BASE_URL)
+        local base_url stripped
+        stripped="${READTHEDOCS_CANONICAL_URL#https://}"
+        stripped="${stripped#http://}"
+        base_url="/${stripped#*/}"
+        # Handle domain-only case (no path after host)
+        [[ "$base_url" == "/${stripped}" ]] && base_url=""
```
ksk-jij: [question]
Is this kind of safe pattern needed? If we do not expect this case in any scenarios, it might be nice to remove those since this sort of guard patterns could make codebase complex.

ksk-jij (changes requested):
@ahao27 Thank you for the update! I've left comments! Please have a look. I wrote several things, but this is based on small investigation. I'd appreciate it if you double check what I wrote even if what I wrote sounds alright for you :)

docs/build.sh
ahao27: Switched to `READTHEDOCS_VERSION` as you suggested.

docs/build.sh
ahao27: Removed along with the URL parsing since `READTHEDOCS_VERSION` is already a clean version identifier.

ksk-jij (approved):
Looks good to me! Thank you! Hopefully this works well!!!

PR #332, opened 2026-04-13
Migrate LocalSearch to accept BinaryModel

### LINKED ISSUE #272: Migrate LocalSearch from old Qamomile
## Description
We will migrate `LocalSearch` in old Qamomile to the new version. When we migrate it, we will remove `IsingMatrix` and also make `LocalSearch` accept `BinaryModel`, both spin and binary types.

ahao27:
## Summary
- Migrated `LocalSearch` from the removed `qamomile.core.post_process` to `qamomile.optimization.post_process` (closes #272)
- Removed dependency on `IsingModel`, `IsingMatrix`, `QuantumConverter`, and `BitsSample` — `LocalSearch` now accepts `BinaryModel` directly
- Supports both **SPIN** (±1) and **BINARY** (0/1) models: internally converts to SPIN for the search, then converts results back to the original vartype
- Returns `BinarySampleSet` instead of `ommx.v1.SampleSet`, aligning with the new optimization API

## Test plan
- [x] 35 tests covering SPIN and BINARY convergence, edge cases (single-variable, linear-only, quad-only models), `max_iter` boundary, random initial states with fixed seeds, and energy-diff correctness
- [ ] Verify no regressions in existing optimization tests (`uv run pytest tests/optimization/ -v`)
- [ ] Lint passes (`uv run ruff check qamomile/optimization/post_process/`)


qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,155 @@
...
+            "first_improvement": self._first_improvement,
+        }
+        if method not in methods:
+            raise ValueError(
+                f"Invalid method: {method}. Choose from {list(methods.keys())}."
+            )
+
+        spin_state = self._to_spin(np.asarray(initial_state))
+        spin_state = self._search(methods[method], spin_state, max_iter)
+        return self._to_sampleset(spin_state)
```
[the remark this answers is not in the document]
ahao27: ** Fixed. `run()` now validates `initial_state` length against `num_bits` and checks value domain (±1 for SPIN, 0/1 for BINARY) before proceeding. Added corresponding tests in `TestInputValidation`.

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,155 @@
...
+        max_iter: int = -1,
+        method: str = "best_improvement",
+    ) -> BinarySampleSet:
+        """Run local search starting from *initial_state*.
+
+        Args:
+            initial_state: Variable values in the model's vartype domain
+                (±1 for SPIN, 0/1 for BINARY).
+            max_iter: Maximum iterations (-1 for unlimited).
+            method: ``"best_improvement"`` or ``"first_improvement"``.
```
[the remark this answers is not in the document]
ahao27: Fixed. Added `ValueError` for `max_iter < -1` at the top of `run()`, with a test covering `[-2, -3, -10]`.

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,155 @@
...
+            if model.vartype != VarType.SPIN
+            else model
+        )
+
+        n = self._spin_model.num_bits
+        self._quad = np.zeros((n, n))
+        for (i, j), v in self._spin_model.quad.items():
+            self._quad[i, j] = v
+            self._quad[j, i] = v
+
```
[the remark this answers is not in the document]
ahao27: Fixed. Replaced the dense `n×n` numpy matrix with a sparse adjacency list (`dict[int, list[tuple[int, float]]]`). `_calc_e_diff` now iterates only over neighbors, so cost is O(degree) per flip and memory is O(edges).

tests/optimization/post_process/test_local_search.py
```diff
@@ -0,0 +1,175 @@
...
+        result = ls.run([1, -1, -1], method=method)
+        _, energy, _ = result.lowest()
+        assert np.isclose(energy, -8.0)
+
+    def test_max_iter_limits_search(self, spin_model):
+        """Search stops after max_iter iterations even if not converged."""
+        ls = LocalSearch(spin_model)
+        result = ls.run([1, 1, -1], max_iter=1, method="best_improvement")
+        _, energy, _ = result.lowest()
+        assert np.isclose(energy, spin_model.calc_energy([-1, 1, -1]))
```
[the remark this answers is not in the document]
ahao27: Fixed. Updated the docstring to precisely describe the test: "After 1 iteration only the best single flip is applied."

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,197 @@
...
+        n: int,
+    ) -> np.ndarray:
+        """Flip the single bit that gives the largest energy decrease."""
+        deltas = [
+            LocalSearch._calc_e_diff(state, neighbors, linear, i) for i in range(n)
+        ]
+        best = int(np.argmin(deltas))
+        if deltas[best] < 0:
+            state[best] = -state[best]
+        return state
```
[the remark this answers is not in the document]
ahao27: Fixed. Added an early-return guard in `_best_improvement` when `n == 0` to avoid `np.argmin([])`. Added tests in `TestConstantOnlyModel` covering constant-only models with both `best_improvement` and `first_improvement` methods.

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,197 @@
...
+
+        n = self._spin_model.num_bits
+
+        # Sparse adjacency list: neighbors[i] = [(j, coeff), ...]
+        self._neighbors: dict[int, list[tuple[int, float]]] = {i: [] for i in range(n)}
+        for (i, j), v in self._spin_model.quad.items():
+            self._neighbors[i].append((j, v))
+            self._neighbors[j].append((i, v))
+
+        self._linear_dict: dict[int, float] = dict(self._spin_model.linear)
```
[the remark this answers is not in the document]
ahao27: Fixed. `LocalSearch.__init__` now raises `ValueError` when `self._spin_model.order > 2` or `self._spin_model.higher` is non-empty, preventing silent incorrect results on HUBO models. Documented in the class docstring's `Raises` section and covered by T`estUnsupportedModels::test_hubo_model_raises`.

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,238 @@
...
+            if model.vartype != VarType.SPIN
+            else model
+        )
+
+        if self._spin_model.order > 2 or self._spin_model.higher:
+            raise ValueError(
+                "LocalSearch only supports quadratic (order <= 2) models, "
+                f"got order={self._spin_model.order} with "
+                f"{len(self._spin_model.higher)} higher-order term(s)."
+            )
```
ksk-jij: [question]
Would you mind if I ask you to explain why we need this assumption, only QUBO? I suppose the current implementation of `_calc_e_diff` is indeed for only QUBO, but we can generalise it, or more simply we could utilise `BinaryModel.calc_energy`. Also, `_neighbours` as well?

Anyways, if this assumption is only due to the limitation of the current implementation (not the limitation of qamomile itself or mathematically impractical), I reckon we should support HUBO if possible.

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,238 @@
...
+    then converts results back to the original vartype.
+
+    Args:
+        model: The binary model to optimize.
+
+    Raises:
+        ValueError: If *model* contains higher-order (order > 2) terms.
+    """
+
+    def __init__(self, model: BinaryModel) -> None:
```
ksk-jij: [question]
Could we take ommx as well similar to converters?

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,238 @@
...
+        deltas = [
+            LocalSearch._calc_e_diff(state, neighbors, linear, i) for i in range(n)
+        ]
+        best = int(np.argmin(deltas))
+        if deltas[best] < 0:
+            state[best] = -state[best]
+            return True
+        return False
+
+    def _to_sampleset(self, spin_state: np.ndarray) -> BinarySampleSet:
```
ksk-jij, line 435: [my opinion]
If we could take ommx format, then it would be nice to output ommx file as well instead of BinarySampleSet.

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,238 @@
...
+        energy delta in place, and reports ``True``. If no flip improves
+        energy, *state* is left unchanged and ``False`` is returned.
+        """
+        for i in range(n):
+            if LocalSearch._calc_e_diff(state, neighbors, linear, i) < 0:
+                state[i] = -state[i]
+                return True
+        return False
+
+    @staticmethod
```
ksk-jij: [question]
Do we need to make this function static?

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,238 @@
...
+    ) -> float:
+        """Calculate the energy difference when flipping bit *idx*.
+
+        Uses the sparse adjacency list so cost is O(degree) per flip,
+        not O(n).
+        """
+        interaction = sum(v * state[j] for j, v in neighbors[idx])
+        return float(-2 * state[idx] * (interaction + linear.get(idx, 0.0)))
+
+    @staticmethod
```
ksk-jij: [question]
Do we need to make this function static?

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,238 @@
...
+            self._neighbors[i].append((j, v))
+            self._neighbors[j].append((i, v))
+
+        self._linear_dict: dict[int, float] = dict(self._spin_model.linear)
+
+    def run(
+        self,
+        initial_state: list[int],
+        max_iter: int = -1,
+        method: str = "best_improvement",
```
ksk-jij: [my opinion]
It might be nice to create `Enum` class to define a set of valid method. Also, `improvement` might be redundant at this point since both of them assume improvement. For instance, 

```python
class LocalSearchMethod(enum.StrEnum):
    """Strategy for picking a bit to flip at each step."""
    BEST = "best"
    FIRST = "first"

    def run(
        self,
        initial_state: list[int],
        max_iter: int = -1,
        method: LocalSearchMethod | str = LocalSearchMethod.BEST,
    ) -> BinarySampleSet:
        try:
            method = LocalSearchMethod(method)
        except ValueError as e:
            valid = [m.value for m in LocalSearchMethod]
            raise ValueError(
                f"Invalid method: {method!r}. Choose from {valid}."
            ) from e

        methods: dict[LocalSearchMethod, Callable[[np.ndarray], bool]] = {
            LocalSearchMethod.BEST: self._best,
            LocalSearchMethod.FIRST: self._first,
        }
```
(You don't need to stick with the above example, but as one example.)

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,238 @@
...
+        else:
+            if not all(v in (0, 1) for v in initial_state):
+                raise ValueError(
+                    "All elements of initial_state must be 0 or 1 for BINARY models."
+                )
+
+        methods: dict[str, _StepFn] = {
+            "best_improvement": self._best_improvement,
+            "first_improvement": self._first_improvement,
+        }
```
ksk-jij: [my opinion]
Creating `methods` every time feels a bit redundant. How about the following-ish style? (The example code assumes the `Enum` suggested above is employed.)

```python
    match method:
        case LocalSearchMethod.BEST:
            step = self._best
        case LocalSearchMethod.FIRST:
            step = self._first
        case _:
            raise error...
```

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,238 @@
...
+    def __call__(
+        self,
+        state: np.ndarray,
+        neighbors: dict[int, list[tuple[int, float]]],
+        linear: dict[int, float],
+        n: int,
+    ) -> bool: ...
+
+
+class LocalSearch:
```
ksk-jij, line 48: [my opinion]
It would be nice to clarify if this search thinks lower is better or not in the docstrings (class itself and `run` method probably?). (I assume so).

tests/optimization/post_process/test_local_search.py
```diff
@@ -0,0 +1,224 @@
...
+    @pytest.mark.parametrize("seed", range(10))
+    def test_random_initial_never_increases_energy(self, spin_model, seed):
+        """Local search from a random initial state never increases energy."""
+        rng = np.random.default_rng(seed)
+        n = spin_model.num_bits
+        initial = rng.choice([-1, 1], size=n).tolist()
+        ls = LocalSearch(spin_model)
+        result = ls.run(initial, method="best_improvement")
+        _, energy, _ = result.lowest()
+        assert energy <= spin_model.calc_energy(initial) + 1e-10
```
ksk-jij: [my opinion]
If we don't use the structure of the spin mode, which means we don't assume anything about the model itself, could we randomly create the model itself as well? For instance, adding `@pytest.mark.parametrize("model_seed", range(5))` and create 5 different model randomly and start from random initial state generated based on `seed`? (It obviously depends on how heavy the test case would be tho)

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,238 @@
...
+    # ------------------------------------------------------------------
+    # Internal helpers
+    # ------------------------------------------------------------------
+
+    def _to_spin(self, state: np.ndarray) -> np.ndarray:
+        """Convert state from the model's vartype to SPIN (±1)."""
+        if self._model.vartype == VarType.BINARY:
+            return np.where(state == 0, 1, -1).astype(float)
+        else:
+            return state.astype(float)
```
ksk-jij: [my opinion]
I reckon this cast `astype(float)` does not work well? I mean not really beneficial since the state must have 0 or 1, or 1 or -1. Could we remove `astype(float)` if I understand this correctly?

Maybe for else branch, we put `copy()` not to break state itself.

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,238 @@
...
+    def _from_spin(self, spin_state: np.ndarray) -> list[int]:
+        """Convert SPIN (±1) state back to the model's vartype."""
+        if self._model.vartype == VarType.BINARY:
+            return [0 if s > 0 else 1 for s in spin_state]
+        else:
+            return [int(s) for s in spin_state]
+
+    def _search(
+        self,
+        step: _StepFn,
```
ksk-jij: [my opinion]
Even if we leave this, personally the name `step` is a difficult to understand. `search_fn` or something is better? `step_fn`?

tests/optimization/post_process/test_local_search.py
ksk-jij, line 1: [my opinion]
For the deterministic case, which is I believe most of them, it would be nice why the test is valid. For example,

```python
    @pytest.mark.parametrize("method", ["best_improvement", "first_improvement"])
    def test_converges_to_optimum(self, spin_model, method):
        """Both methods reach the global optimum from the same starting state."""
        ls = LocalSearch(spin_model)
        result = ls.run([1, -1, -1], method=method)
        _, energy, _ = result.lowest()
        assert np.isclose(energy, -8.0)
```

In this case, for both methods,
1. Since z0 is better and first to be flipped, [-1, -1, -1] is obtained for the first round,
2. For both cases again, z1 is better to be flipped (I believe), [-1, 1, -1] is obtained for the second round,
3. It is optimal and further flip is no longer happens, thus, the result is [-1, 1, -1] and energy is -8.

ksk-jij, line 1: [my opinion]
For the deterministic case, everything that is determined should be asserted. including the solution itself, not only the energy.

ksk-jij (commented):
@ahao27 Hi! Thank you for the PR! I've left some comments! Please have a look!

qamomile/optimization/post_process/local_search.py
ahao27: Refactored. 

`_calc_e_diff` now uses a unified sparse per-index adjacency built from l`inear / quad / higher,`. 
Added
- a`TestLocalSearchHubo` covering a cubic SPIN fixture 
- a `E-vs-brute-force` consistency test 
- a randomized `test_random_hubo_never_increases_energy` 

over 5 seeds × 5 seeds × 2 methods.

qamomile/optimization/post_process/local_search.py
ahao27: Done. `LocalSearch` now accepts either `BinaryModel` or `ommx.v1.Instance`.

qamomile/optimization/post_process/local_search.py
ahao27: Done. `run()` now returns `ommx.v1.Solution` when the input was `ommx.v1.Instance`, and `BinarySampleSet` otherwise.

qamomile/optimization/post_process/local_search.py
ahao27: Converted to an instance method.

qamomile/optimization/post_process/local_search.py
ahao27: Converted to an instance method.

qamomile/optimization/post_process/local_search.py
ahao27: Done. `Added LocalSearchMethod(enum.StrEnum)` with `BEST = "best" / FIRST = "first"` (dropped the redundant `_improvement` suffix as suggested).

qamomile/optimization/post_process/local_search.py
ahao27: Fixed to the following style.
```python
match method:
    case LocalSearchMethod.BEST:
        step_fn: _StepFn = self._best_improvement
    case LocalSearchMethod.FIRST:
        step_fn = self._first_improvement
    case _:
        assert False, "unreachable"
```

qamomile/optimization/post_process/local_search.py
ahao27: Done. Class docstring now shows with "Local search energy minimizer" and explicitly states "lower is better". `run()` reiterates that only strictly energy-decreasing flips are accepted, so the returned state is a local minimum.

qamomile/optimization/post_process/local_search.py
ahao27: Removed.
Due to values arrive as ±1 / 0-1 ints after `run()`'s domain validation, so the float cast was unnecessary.
Added `.copy()` in the SPIN branch as you suggested, so later in-place flips never reach back through to the caller's ndarray.

qamomile/optimization/post_process/local_search.py
ahao27: Renamed to `step_fn`

tests/optimization/post_process/test_local_search.py
ahao27: Done. 

- Added `_random_spin_model` / `_random_binary_model` / `_random_hubo_model` helpers at module scope, and three `test_random_`*`_never_increases_energy` tests parametrized over `model_seed × init_seed × method` (5 × 5 × 2 = 50 cases each).

- The only asserted invariant is energy ≤ initial_energy + 1e-10, which is structure-independent.

tests/optimization/post_process/test_local_search.py
ahao27: Done.

ahao27: Done. Every deterministic test now carries a ΔE derivation in its docstring (starting state → per-index ΔE → chosen flip → repeat → termination condition). Every fixture docstring additionally lists either the full brute-force energy landscape (for 3-variable+ fixtures) or the unique / degenerate minima (for 1- and 2-variable fixtures), so the asserted optima are verifiable without re-running the math

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,363 @@
...
+            # instance, so `evaluate` reports feasibility against the
+            # user's constraints. But the sample dict includes slack
+            # decision variables introduced by `to_hubo` on the working
+            # copy — those IDs don't exist on the original instance, and
+            # passing them to `evaluate` raises. Filter to the original
+            # decision-variable IDs before handing off.
+            original_ids = {v.id for v in self._ommx_instance.decision_variables}
+            sample = {
+                k: v for k, v in sample_set.samples[0].items() if k in original_ids
+            }
```
ksk-jij: [question]
I wonder if we could use `used_decision_variables` instead of `decision_variables`. Is there any specific reasons that you decided to use `decision_variables`?

If I understand these properties correctly, `used_decision_variables` only returns variables that are not fixed and used in the objective function, whereas `decision_variables` returns everything. This means if some decision variables in the ommx instance have been fixed, this way could make different mappings or just break the process since some decision variables would not be contained in `sample`.
(Especially, `to_hubo` function does return only free variables, which are not fixed.)

ksk-jij: [question]
I wonder if we could use `used_decision_variables` instead of `decision_variables`. Is there any specific reasons that you decided to use `decision_variables`?

If I understand these properties correctly, `used_decision_variables` only returns variables that are not fixed and used in the objective function, whereas `decision_variables` returns everything, which include fixed variables as well. I don't think the current implementation breaks anything, but having a standard way to use `used_decision_variables` feels nice to avoid confusion. If we use `used_decision_variable_ids()` here, then I reckon we could simply write
```python
sample = { k: v for k, v in sample_set.samples[0].items() if k in self._ommx_instance.used_decision_variables_ids() }
```

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,363 @@
...
+        For an :class:`ommx.v1.Instance` with constraints, ``to_hubo``
+        introduces slack bits on top of the user's decision variables, so
+        this value can exceed the number of declared variables. Use this
+        to size ``initial_state``.
+        """
+        return self._model.num_bits
+
+    def run(
+        self,
+        initial_state: Sequence[int] | np.ndarray,
```
ksk-jij: [my opinion]
I'm terribly sorry, but at this point (with ommx support), it would be nice to accept ommx.v1.Solution as well? What do you think?

ksk-jij (commented):
@ahao27  I appreciate you addressing my comments! I've left two comments. I look froward to hearing your thoughts on them.

qamomile/optimization/post_process/local_search.py
ahao27: done. run() now accepts Sequence[int] | np.ndarray | ommx.v1.Solution for initial_state.

qamomile/optimization/post_process/local_search.py
ahao27: Changed. The filter is now
```python
used_ids = self._ommx_instance.used_decision_variable_ids()
sample = {
    k: v for k, v in sample_set.samples[0].items() if k in used_ids
}
return self._ommx_instance.evaluate(sample)
```

qamomile/optimization/post_process/local_search.py
```diff
@@ -0,0 +1,408 @@
...
+
+        Args:
+            solution: The ommx Solution to warm-start from.
+
+        Returns:
+            A list of length :attr:`num_bits` in the internal BinaryModel's
+            vartype domain (±1 for SPIN, 0/1 for BINARY). The domain is
+            validated by :meth:`run` after this conversion.
+        """
+        entries = dict(solution.state.entries)
```
ksk-jij, line 306: [my opinion]
Please let me know if I missed some cases.

I feel it would be nice to verify the given `solution` corresponds to the `self._ommx_instance`. How about checking

1. if `self._ommx_instance` is `None`, then we should raise an error, and
2. if `self._ommx_instance.used_decision_variable_ids()` is completely in `solution.state.entries`?

This is not perfect still, however, it still can prevent problems. In this case, also we can remove `get` and use just `entries[self._model.index_new_to_origin[i]]`, which may make it easy to find internal bugs, which could be  silently grasped by `get`.

tests/optimization/post_process/test_local_search.py
ksk-jij, line 1: [my opinion]
It may be nice to have test cases having `substitution_value` is not None. Those can be done by `ommx.v1.Instance.partial_evaluiation`. The function returns `ommx.v1.Instance` with fixed variables. Then, the fixed values will not be appeared in `used_decision_variables`. We can make several cases such as the first variable is fixed, the middle variable is fixed, the last variable is fixed and so on. It would be nice to have test cases that show that the local search function works with those.

ksk-jij (commented):
@ahao27 Thank you for the PR! I've left comments with regard of ommx support. I'd appreciate it if you could have a look and let me know your thoughts.

ksk-jij (approved):
@ahao27 Thank you for addressing my comments! Looks good to me!

PR #343, opened 2026-04-21
test: stabilize flaky test_hadamard_all_sampling

ksk-jij:
## Summary
- The `seeded_executor` fixture in [tests/circuit/conftest.py](tests/circuit/conftest.py) was advertised as seeded but did not actually pin `BasicSimulator`'s RNG, so sampling was non-deterministic across CI runs. `test_hadamard_all_sampling[4]` flakes at ~7% per run because 28 per-outcome 3σ bounds are checked in aggregate (n=2,3,4).
- Pin the simulator RNG via `backend.set_options(seed_simulator=_SIMULATOR_SEED)` in the fixture so any failure is reproducible.
- Widen the per-outcome bound from 3σ to 5σ so the test passes **for any seed**, not just the one we pinned. Rationale: Bonferroni correction for aggregate `P_fail <= 1e-5` across the 28 checks needs ~5.09σ; 5σ yields aggregate flake probability ~1.6e-5. A 10 000-run Monte Carlo saw max deviation 4.52σ, and a 200-seed `BasicSimulator` sweep saw max 3.55σ — zero failures at 5σ. The 5σ bound is still tight enough to catch a broken H gate.

## Test plan
- [x] `uv run pytest tests/circuit/test_composite_gate.py tests/circuit/test_qft.py` — 171/171 pass (all seven `seeded_executor` consumers).
- [x] `uv run pytest tests/` — 6011 passed, 14 skipped.
- [x] Monte Carlo validation: 200 distinct seeds × n∈{2,3,4} → 0 failures at 5σ, max observed 3.55σ.
- [x] Analytic validation: 10 000-run multinomial simulation → max 4.52σ (≪ 5σ bound).

### DIFF
--- a/tests/circuit/conftest.py
+++ b/tests/circuit/conftest.py
@@ -11,6 +11,9 @@ if TYPE_CHECKING:
     from qiskit.circuit import QuantumCircuit
 
 
+_SIMULATOR_SEED = 901
+
+
 @pytest.fixture
 def qiskit_transpiler():
     """Get Qiskit transpiler."""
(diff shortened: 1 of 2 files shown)


ksk-jij (commented):
Looks good to me.

PR #353, opened 2026-04-27
Support OMMX round-trip output via polymorphic decode()

ahao27:
## Summary

- `MathematicalProblemConverter.decode()` is now polymorphic on construction input: OMMX-backed converters return an `ommx.v1.SampleSet`, `BinaryModel`-backed converters keep returning `BinarySampleSet`. 
- Adds public `decode_to_binary_sampleset()` for callers that need the QUBO-domain `BinarySampleSet` (e.g. driving COBYLA with penalty-included `energy`). Two-method shape mirrors the `Support-local-search` precedent.
- Fixes a silent caller-side mutation: `Instance.to_qubo` rewrites the instance it is called on (drops constraints, penalizes objective, adds slack DVs). `MathematicalProblemConverter.__init__` and `FQAOAConverter.__init__` now run `to_qubo` on a deep copy via `Instance.from_bytes(model.to_bytes())`. The deep copy retains original-constraint metadata so OMMX-side feasibility is reported against the user's *original* constraints, not the absorbed-penalty form.

## Scope

- `qamomile/optimization/converter.py` — polymorphic `decode()`, new `decode_to_binary_sampleset()`, mutation fix.
- `qamomile/optimization/binary_model/sampleset.py` — `BinarySampleSet.to_ommx_samples()` helper (groups `num_occurrences` into `ommx.v1.Samples` sample IDs).
- `qamomile/optimization/fqaoa.py` — same mutation fix; wires `self.instance` to the post-qubo working copy so FQAOA also gets the OMMX round-trip.
- Tutorials (`docs/{en,ja}/optimization/qaoa_graph_partition.{py,ipynb}`) — minimal migration: `cost_fn` and analysis-section call sites use the new public `decode_to_binary_sampleset()` so COBYLA keeps seeing penalty-aware energy. Full narrative migration to OMMX primitives (eliminating the hand-rolled `is_feasible`/`count_cut_edges` helpers) is a deliberate follow-up — see below.

## Test plan

- [x] `uv run pytest tests/optimization/ -q` — 885 passed (was 883 on main; +12 new tests minus 1 negative test that lost meaning under polymorphic decode()).
- [x] `uv run pytest -m docs -k "qaoa_graph_partition"` — 2 passed (en + ja). `.ipynb` re-executed via `jupytext --to notebook --execute`.
- [x] `uv run ruff check qamomile/optimization/ tests/optimization/` — clean.
- [x] `uv run ruff format --check ...` — clean.
- [x] `uv run zuban check qamomile/` — 26 errors (matches `main` baseline; no new diagnostics).
- [x] Caller-instance immutability regressions for both `QAOAConverter` and `FQAOAConverter` (bytes-snapshot equality after construction).
- [x] End-to-end OMMX round-trip on Qiskit — parametrized over seeds `[0, 1, 42]` for both unconstrained max-cut and equality-constrained graph partition.
- [x] Integer-slack reconstruction — verifies `evaluate_samples` rebuilds integer DVs from `to_qubo`'s log-encoded slack bits, parametrized over seeds.

## Follow-ups (out of scope for this PR)

- Migrate the graph-partition tutorial's analysis section away from `decode_to_binary_sampleset` + hand-rolled `is_feasible` / `count_cut_edges` to OMMX primitives (`sample_set.feasible`, `summary_with_constraints`, `best_feasible`). Tracked as the explicit reason `decode_to_binary_sampleset` was promoted to public — it's the right layer for that callsite, but the surrounding markdown narrative needs reframing too.
- Audit other tutorials for similar manual feasibility/objective workarounds.


qamomile/optimization/fqaoa.py
```diff
@@ -98,15 +98,28 @@ def __init__(
...
+        # one, not the post-qubo working copy.
         self._original_instance = instance
 
         # Pass the BinaryModel to the base class (which converts to SPIN internally)
         super().__init__(binary_model)
+        # Base class set self.instance = None because we passed a BinaryModel.
+        # Wire in the post-qubo working copy so decode() can round-trip back
+        # to ommx.v1.SampleSet (feasibility evaluated against the user's
+        # original constraints, which OMMX retains internally).
+        self.instance = working
```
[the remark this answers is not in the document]
ahao27: Added the FQAOA decode round-trip test

qamomile/optimization/converter.py
ksk-jij, line 1: [my opinion]
With the current implementation, the converters of QRAO cannot get results in ommx format since I believe `decode` itself is pretty much for algorithms that the sampled bitstrings directly correspond to the decision variables. It would be nice to support ommx format with QRAO as well.

How about adding `decode` to `QRACConverterBase` class in qrao/base_converter.py such as 

```python
def decode(self, rounded_spins_list: list[list[int]]) -> BinarySampleSet | ommx.v1.SampleSet:
    # 0. Check if the rounded_spins_list is valid. (if all of them are spins, if the length of every binary is the same as the number of the given problem's decision variables, and more if you think we can verify more)
    # 1. create SampleResult[list[int]] from rounded_bits.
    # 2. call super().decode() with the crested SampleResult.
```

I'd appreciate it if you could let me your thoughts.

tests/optimization/test_qaoa_decode.py
ksk-jij, line 1: [must]
Could you test if the type of returned value is as expected: `ommx.v1.SampleSet`?

qamomile/optimization/binary_model/sampleset.py
```diff
@@ -23,3 +24,53 @@ def energy_mean(self) -> float:
         _e = np.array(self.energy)
         _n = np.array(self.num_occurrences)
         return float(_e @ _n / np.sum(_n))
+
+    def to_ommx_samples(self) -> ommx.v1.Samples:
```
ksk-jij: [my opinion]
At the moment, the conversion from ommx to BinaryModel is done in converter not BinaryModel itself. So, I felt like it would be nice to put all the conversion stuff in converter not in `SampleSet` itself. I'd appreciate it if you could share your thoughts with me.

ksk-jij (changes requested):
@ahao27 Thank you for the PR! I've left small comments, please have a look at them.

tests/optimization/test_qaoa_decode.py
ahao27: Done in `a5a33a4`. Added `assert isinstance(result, ommx.v1.SampleSet)` to verify the return type when decoding from an OMMX instance input.

qamomile/optimization/binary_model/sampleset.py
ahao27: The converter now owns both directions of the OMMX conversion.

qamomile/optimization/converter.py
ahao27: Added `decode(rounded_spins_list)` to `QRACConverterBase` , so QRAO converters can also return `ommx.v1.SampleSet`

ksk-jij (approved):
@ahao27 Looks good to me. Thank you.

PR #360, opened 2026-04-28
feat(frontend): broadcast single-qubit gates over qubit arrays

yuyamashiro:
## Summary
- Single-qubit gates (`h`, `x`, `y`, `z`, `s`, `t`, `sdg`, `tdg`, `rx`, `ry`, `rz`, `p`) now accept `Vector[Qubit]` in addition to `Qubit`. `qs = qmc.h(qs)` is now the idiomatic way to apply a gate across a whole register.
- Broadcast lowers to the same `ForOperation` IR a hand-written `for i in qmc.range(n): qs[i] = gate(qs[i])` produces, so transpiler / resource-estimation / visualization passes need no broadcast-specific handling. Follows the overload pattern already used by `measure`.
- Cross-backend tests (Qiskit, QuriParts, CUDA-Q) cover sampling + expval paths with randomized angles and register sizes (`n ∈ {1, 2, 3, 5}`, seeds `{0, 1, 2, 42}`), validated against the analytic `<sum_i Z_i> = n * cos(theta)` reference.
- Tutorial 02 (`docs/{en,ja}/tutorial/02_parameterized_kernels.py`) documents the broadcast form alongside the explicit-loop form.

## Test plan
- [x] `uv run pytest tests/circuit/test_gate_broadcast.py` — 143 passed across Qiskit, QuriParts, CUDA-Q
- [x] `uv run pytest tests/ --ignore=tests/docs --ignore=tests/cudaq` — 6570 passed
- [x] `uv run pytest tests/ -m cudaq` — 359 passed
- [x] `uv run pytest tests/docs -m docs` — 30 passed (tutorial 02 EN/JA both execute cleanly)
- [x] `uv run ruff check qamomile/ tests/circuit/test_gate_broadcast.py` — clean
- [x] `uv run zuban check qamomile/circuit/frontend/operation/qubit_gates.py` — clean


docs/en/tutorial/02_parameterized_kernels.py
ksk-jij, line 1: [must]
Could you re-create and re-execute the corresponding notebook?

docs/ja/tutorial/02_parameterized_kernels.py
ksk-jij, line 1: [must]
Could you re-create and re-execute the corresponding notebook?

qamomile/circuit/frontend/operation/qubit_gates.py
ksk-jij, line 1: [my opinion]
Using `for_loop` to broadcast single-qubit gate is fine, but as Qamomile philosophy, it would be nicer to add `GateVectorOperation` (such as `MeasurementVectorOperation`) and try to emit the operation with a native way of the target quantum SDK (such as qiskit). If there is no native way to apply one qubit gate to qubits object (in the way of the target quantum SDK), then we can naively lower `GateVectorOperation` to `ForOperation` + `GateOperation`.

ksk-jij (changes requested):
I'd appreciate the PR. I've left some comments, please have a look.

qamomile/circuit/frontend/operation/qubit_gates.py
ksk-jij: Sorry for confusion this comment may cause, but ignore this comment.

qamomile/circuit/frontend/operation/qubit_gates.py
```diff
@@ -120,44 +192,206 @@ def _apply_three_qubit_gate(
...
+        gate_type: The gate kind to apply.
+
+    Returns:
+        A `Qubit` for scalar input, a `Vector[Qubit]` for array input.
+
+    Raises:
+        TypeError: If `target` is neither a `Qubit` nor a
+            `Vector[Qubit]`.
+    """
+    if isinstance(target, VectorClass) and target.element_type is Qubit:
```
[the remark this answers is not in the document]
ksk-jij: Switched `is` to `==` in [3f64605f](https://github.com/Jij-Inc/Qamomile/commit/3f64605f) for consistency with the `measure` dispatch pattern in `frontend/operation/measurement.py`. Applied across `p()` and `_dispatch_rotation_gate()` as well.

qamomile/circuit/frontend/operation/qubit_gates.py
```diff
@@ -120,44 +192,206 @@ def _apply_three_qubit_gate(
...
+def h(target: Qubit) -> Qubit: ...
+@overload
+def h(target: Vector[Qubit]) -> Vector[Qubit]: ...
+def h(target: Union[Qubit, Vector[Qubit]]) -> Union[Qubit, Vector[Qubit]]:
+    """Hadamard gate.
 
-def x(qubit: Qubit) -> Qubit:
-    """Pauli-X gate (NOT gate)."""
-    return _apply_single_qubit_gate(qubit, GateOperationType.X)
+    Applied to a single `Qubit` it returns the rotated qubit. Applied to a
```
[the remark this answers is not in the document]
ksk-jij: Reworded the `h()` docstring in [3f64605f](https://github.com/Jij-Inc/Qamomile/commit/3f64605f) to use "transformed qubit" — H is not a rotation, so the gate-agnostic phrasing is more accurate.

tests/circuit/test_gate_broadcast.py
```diff
@@ -0,0 +1,419 @@
...
+
+# ---------------------------------------------------------------------------
+# Backend execution: Qiskit and QuriParts (sampling + expval)
+# ---------------------------------------------------------------------------
+
+# QuriParts is the more permissive of the two for symbolic shapes;
+# exercise both on every test where the SDKs are available.
+pytest.importorskip("qiskit")
+pytest.importorskip("quri_parts")
+pytest.importorskip("quri_parts.qulacs")
```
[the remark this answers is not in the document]
ksk-jij: Fixed in [3f64605f](https://github.com/Jij-Inc/Qamomile/commit/3f64605f). The module-level `pytest.importorskip` calls were removed; each backend now has its own try/except + per-entry `skipif` on the `BACKENDS` parametrize list, mirroring the cudaq pattern. The IR-shape tests no longer get skipped in environments without qiskit / quri-parts.

tests/circuit/test_gate_broadcast.py
```diff
@@ -0,0 +1,419 @@
...
+    def _circuit(theta: qmc.Float) -> qmc.Vector[qmc.Bit]:
+        qs = qmc.qubit_array(n, "qs")
+        m = qs.shape[0]
+        for i in qmc.range(m):
+            qs[i] = qmc.rx(qs[i], theta)
+        return qmc.measure(qs)
+
+    return _circuit
+
+
```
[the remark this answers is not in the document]
ksk-jij: Removed `_kernel_rx_loop` in [3f64605f](https://github.com/Jij-Inc/Qamomile/commit/3f64605f). The broadcast-vs-loop equivalence test now defines the explicit-loop kernel inline rather than via a shared helper.

docs/en/tutorial/02_parameterized_kernels.py
ksk-jij: 対応しました。[8ebc50c2](https://github.com/Jij-Inc/Qamomile/commit/8ebc50c2) で `jupytext --to ipynb` + `jupyter nbconvert --execute` により再生成・再実行しています。新しい "Broadcasting Single-Qubit Gates Over an Array" セクション（`rotation_layer_broadcast.draw` の画像出力 + bind/sweep の print 出力）も反映済みです。

docs/ja/tutorial/02_parameterized_kernels.py
ksk-jij: 日本語版も同様に [8ebc50c2](https://github.com/Jij-Inc/Qamomile/commit/8ebc50c2) で再生成・再実行済みです。新しい「単一量子ビットゲートを配列にブロードキャストする」セクションの draw 画像と bind/sweep の出力が含まれています。

docs/en/tutorial/02_parameterized_kernels.py
ksk-jij, line 1: [my opinion]
When the kernel is drawn with this feature with `fold_loops=False`, gates are drawn like the attached one, but since the gates are just a single-qubit gate, it is not good. Each gate must be drawn as same as using `for ... qmc.range(...)`, for instance,
```python
for i in qmc.range(n):
    q[i] = qmc.h(q[i])
```

Image

So, the drawing feature needs to be updated as well.

ksk-jij (changes requested):
Thank you for addressing my comments, please have a look at the additional comments!

docs/en/tutorial/02_parameterized_kernels.py
ksk-jij: Thanks for the feedback.

First, with `fold_loops=False` alone, on my end both broadcast and explicit-loop forms unroll to four individual single-qubit gates — I couldn't reproduce the wide multi-qubit-tile rendering shown in the attached screenshot. The notebook re-execution in [88404a6f](https://github.com/Jij-Inc/Qamomile/commit/88404a6f) also no longer shows that state.

That said, the underlying issue surfaces with `fold_loops=True` (the default): broadcast was leaking the internal display name `_bcast_<gate>_idx` into the rendering, so it showed

```
for _bcast_h_idx in qm.range(4)
q[_bcast_h_idx] = h(q[_bcast_h_idx])
```

instead of the explicit-loop form's

```
for i in qm.range(4)
q[i] = h(q[i])
```

I read this as the actual concern behind "Each gate must be drawn as same as using `for ... qmc.range(...)`", and addressed it in [57926fb4](https://github.com/Jij-Inc/Qamomile/commit/57926fb4).

**Changes:**
- `_broadcast_single_qubit_gate` / `_broadcast_rotation_gate` / `_broadcast_phase_gate` now pass `var_name="i"` to `for_loop(...)`. `var_name` is display-only — IR identity is carried by UUIDs — so this is purely a rendering change.
- With `fold_loops=True`, broadcast now renders with the exact same header/body text as a hand-written `for i in qmc.range(n): q[i] = h(q[i])`.
- Added a regression test `test_broadcast_loop_var_renders_as_i` pinning the loop var name.

One caveat: when the user writes two consecutive broadcasts (`q = qmc.h(q); q = qmc.ry(q, theta)`), broadcast emits two `ForOperation`s — so the folded rendering shows two separate `for i in qm.range(n)` blocks rather than one fused block (which is what the explicit form gets when both gates are inside a single `for`). This mirrors the user's source structure (one broadcast = one for-loop) and isn't fused. Happy to discuss if you'd like fusion behavior; that would be a larger IR-level change.

ksk-jij: You're right — I confirmed the published readthedocs preview at https://jij-inc-qamomile--360.com.readthedocs.build/360/en/tutorial/parameterized-kernels/ still showed the wide-block rendering. Investigating, the broadcast cell's image embedded in `de002538`'s notebook was the wide-block one despite my having "re-executed" it (probably stale runtime state during my nbconvert run).

A fresh `jupyter nbconvert --to notebook --execute --inplace --ExecutePreprocessor.kernel_name=python3` produces the correct per-qubit single-gate rendering matching the explicit-loop form. Pushed in [aabeffd1](https://github.com/Jij-Inc/Qamomile/commit/aabeffd1). Once the readthedocs preview rebuilds for this branch, the page should show four individual H + Ry gates instead of the two wide blocks.

Sorry for the back-and-forth.

docs/ja/tutorial/02_parameterized_kernels.py
```diff
@@ -98,6 +98,36 @@ def bad_iteration(n: qmc.UInt) -> qmc.Vector[qmc.Qubit]:
 # %% [markdown]
 # 常にインデックスベースのアクセスを使用してください：`for i in qmc.range(n): q[i] = qmc.h(q[i])`。
 
+# %% [markdown]
+# ## 単一量子ビットゲートを配列にブロードキャストする
```
ksk-jij, line 102: [must]
In Japanese docs, artificial breakdown makes weird spaces, please do not insert it.

tests/circuit/test_gate_broadcast.py
ksk-jij, line 1: [my opinion]
Could you test all the single-qubit, rotation gates?

ksk-jij (changes requested):
Thank you for addressing all the comments and sorry for this second round comment. Please have a look.

docs/ja/tutorial/02_parameterized_kernels.py
ksk-jij: Fixed in [33407abe](https://github.com/Jij-Inc/Qamomile/commit/33407abe). Collapsed the soft line breaks in the broadcast section's Japanese markdown into single lines per paragraph, per the translation skill's rule 8 (人工改行の扱い) — soft line breaks render as half-width spaces in the notebook and look unnatural in Japanese.

tests/circuit/test_gate_broadcast.py
ksk-jij: Added in [33407abe](https://github.com/Jij-Inc/Qamomile/commit/33407abe).

Previously only X (`test_x_broadcast_flips_all_qubits`), H (`test_h_broadcast_uniform_distribution`), and RX (`test_rx_broadcast_matches_loop_expval`, `test_rx_broadcast_expval_matches_n_cos`) had **execution-level** coverage. The remaining nine gates (Y, Z, S, SDG, T, TDG, RY, RZ, P) only had IR-shape assertions in `TestAllPrimitivesBroadcast`.

New `TestAllPrimitivesBroadcastSampling` class:
- `test_non_parametric_broadcast_matches_loop_sampling`: sweeps {h, x, y, z, s, sdg, t, tdg} × all backends. Asserts the broadcast and explicit-loop forms produce the same set of measurement outcomes (the kernel applies an H broadcast first so the trailing gate has observable effect, then asserts equality between the two sample dicts).
- `test_parametric_broadcast_matches_loop_sampling`: sweeps {rx, ry, rz, p} × all backends × seeds {0, 1, 2, 42}. Mixes boundary angles (0, π, 2π) with a random angle so each gate's degenerate identity / -identity cases are covered alongside the generic case.

Local run on the qiskit backend: 24 new tests pass (8 gates + 4 gates × 4 seeds). The full `tests/circuit/test_gate_broadcast.py` suite now has 84 passed, 132 skipped (skips are quri_parts / cudaq parametrize entries since those SDKs aren't installed locally; CI exercises them).

ksk-jij (approved):
Looks good to me.

PR #375, opened 2026-05-07
Feature/alternating ansatz

baptistechev:
## Summary

Adds Qamomile support for the Alternating Operator Ansatz (AOA).

This includes:

- XY rotation mixers with different architectures (ring, fully-connected…)
- Efficient preparation of arbitrary Dicke states as initial states

## Motivation

AOA extends QAOA by enabling constrained optimization via XY mixers and structured initial states (Dicke states).

## Files

- `qamomile/circuit/algorithm/aoa.py` — backend-agnostic qkernel for AOA using XY mixers
- `qamomile/circuit/algorithms/dicke.py` — backend-agnostic qkernel for Dicke state preparation (kept separate since Dicke states are useful beyond AOA)
- `qamomile/optimization/aoa.py` — wires the above kernels to the cost Hamiltonian; entry point for most users
- `qamomile/optimization/schedules/dicke.py` — computes indices and angles consumed by `dicke.py`; pure helper, no quantum circuit logic

Tests files (support for cross-backend and randomized paramters)
- `tests/circuit/algorithm/test_aoa.py` — tests for AOA qkernel 
- `tests/circuit/algorithm/test_dicke/py` — tests for Dicke qkernel (partial)
- `tests/optimization/test_aoa.py` — tests for AOA converter (partial)
- `tests/optimization/schedules/test_dickes` —tests for dickes preparation helper (partial, maybe no need for thorough testing)

Docs
- `docs/en/algorithm/aoa_coloring_problem.ipynb` — tutorial notebook in EN
- `docs/en/algorithm/aoa_coloring_problem.py` — notebook pytext in EN
 - `docs/ja/algorithm/aoa_coloring_problem.ipynb` — tutorial notebook in JA
- `docs/ja/algorithm/aoa_coloring_problem.py`—notebook pytext in JA

## Questions for reviewers

1. Semantically would it make more sens for QAOAConverter to be a subclass of AOAConverter?  but I’m not sure what is the best for choice API-wise
2. This is about user interface API. Custom mixer is hard to use for people unfamiliar with the paper. Do we keep it or just ring or/and fully-connected? (Fully-connected is faster to mix but uses more gates/depth)
3. The qkernel for dicke states preparation uses a condition `if` condition flagged by copilot review (non-optimal complexity in extreme cases). Removing the condition forces to put at least one additional argument in the qkernel signature and will degrade the API readibility. Should I implement modify or keep as it is? 

## Implementation notes

**About AOAConverter**

- `AOAConverter` is implemented as a subclasses `QAOAConverter`
- `get_cost_hamiltonian()` is reused from the parent class
- `transpile()` is overridden to support XY mixers

**About XY mixers**

- Implemented two XY mixers `"ring"` and `"fully-connected"` . Same approach as discussed previously( Can be re-explained if needed).
- User can choose which mixer to use using transpile argument `mixer="fully-connected"`
- Custom mixer can be used by giving predefined indices `pair_indices_mixer: np.ndarray`
- As in the technical test, the implementation uses the *partition approximation* rather than full trotterization.

**About Dicke Preparation**

- I made a separate module `circuit/algorithm/dicke.py` since Dicke state usage is not limited to AOA. Same thing for the helper functions in `optimization/schedule/dicke.py`
- The `AOAConverter` can prepare AOA ansatzs using `initial_state="dicke"` . If so, we prepare a product of Dicke states with one Dicke state per block.
- `block_size : int`  is used to defined the size of the block (group of qubits) on which we prepare a Dicke state and on which the XY mixer acts. If not provided, this is set on `self.spin_model.num_bits` (no block)

**Notebook tutorial : AOA for graph coloring**
The notebook present a solution to the graph coloring problem using AOA and show the benefit of using AOA over QAOA. 
The tutorial is based on the QAOA graph partitioning tutorial and follow the same steps, adding high level explanation to guide the user through the choice of AOA parameters.
The mathematical details are skipped and we invite the reader to refer to the original arXiv paper if needed.

## Status

- [x]  Implement XY mixers
- [x]  Implement Dicke state preparation
- [x]  Tests
- [x]  Documentation / tutorial

## How to test

Example:

```python
from qamomile.optimization.aoa import AOAConverter

converter = AOAConverter(instance)
converter.transpile(
    transpiler,
    p=5,
    initial_state="dicke",
    hamming_weight=1,
    mixer="fully-connected",
    block_size=num_colors,
)
```

Run tests:

```bash
uv run pytest tests/circuit/algorithm/test_aoa.py
uv run pytest tests/optimization/test_aoa.py
uv run pytest tests/circuit/algorithm/test_dicke.py
uv run pytest tests/optimization/schedules/test_dicke.py
```

## Notes

- Opening as draft for early feedback on API and design
- Happy to adjust implementation if needed


qamomile/optimization/aoa.py
```diff
@@ -0,0 +1,785 @@
...
+        """
+        mixer = MixerName(mixer)
+
+        if pair_indices is not None:
+            if mixer != MixerName.RING:
+                warnings.warn(
+                    f"pair_indices was provided; the mixer={mixer!r} argument is ignored.",
+                    stacklevel=3,
+                )
+            return self._normalize_pair_indices(pair_indices)
```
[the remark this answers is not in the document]
baptistechev: @copilot apply changes based on [this feedback](https://github.com/Jij-Inc/Qamomile/pull/375#discussion_r3199146529)

qamomile/optimization/aoa.py
```diff
@@ -0,0 +1,818 @@
...
+
+        effective_block_size = (
+            self.spin_model.num_bits if block_size is None else block_size
+        )
+
+        resolved_pair_indices_mixer = self._resolve_pair_indices(
+            mixer=mixer,
+            pair_indices=pair_indices_mixer,
+            block_size=effective_block_size,
+        )
```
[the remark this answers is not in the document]
baptistechev: @copilot apply changes based on [this feedback](https://github.com/Jij-Inc/Qamomile/pull/375#discussion_r3200249792)

qamomile/optimization/schedules/dicke.py
```diff
@@ -0,0 +1,161 @@
...
+    pair_indices = []
+    triplets_indices = []
+    pair_angles = []
+    triplets_angles = []
+
+    for n in range(n_dicke, 1, -1):
+        # At column n, at most n-1 excitations can flow to the qubits below.
+        k_sub = min(k_dicke, n - 1)
+        p, tr, pa, ta = _scs_schedule(n, k_sub)
+        pair_indices += p.tolist()
```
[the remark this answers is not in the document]
baptistechev: @copilot apply changes based on [this feedback](https://github.com/Jij-Inc/Qamomile/pull/375#discussion_r3200249860)

qamomile/circuit/algorithm/state_preparation/dicke.py
```diff
@@ -0,0 +1,158 @@
...
+        stage_c2 = pair_indices[stage_idx, 1]
+        for triplet_row in qmc.range(triplets_indices.shape[0]):
+            if triplets_indices[triplet_row, 2] == stage_c2:
+                q = scs_gate_3q(
+                    q,
+                    triplets_indices[triplet_row, 0],  # t (target)
+                    triplets_indices[triplet_row, 1],  # c1 (first control)
+                    triplets_indices[triplet_row, 2],  # c2 (second control)
+                    triplets_angles[triplet_row],
+                )
```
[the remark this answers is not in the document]
baptistechev: Not sure if I want to fix that

qamomile/optimization/aoa.py
```diff
@@ -0,0 +1,830 @@
+"""Converter for the Alternating Operator Ansatz (AOA).
+
+AOA reuses the same Ising cost Hamiltonian as QAOA but replaces the standard
+X mixer with a constraint-aware XY mixer. The XY mixer schedule is encoded as a
+matrix of qubit pairs, which can either be supplied explicitly or generated
+from a named built-in schedule.
+"""
```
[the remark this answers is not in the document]
baptistechev: @copilot apply changes based on [this feedback](https://github.com/Jij-Inc/Qamomile/pull/375#discussion_r3232747417)

qamomile/circuit/algorithm/aoa.py
```diff
@@ -0,0 +1,405 @@
+"""AOA (Alternating Operator Ansatz) circuit building blocks.
+
+This module provides the quantum circuit components for the Alternating Operator Ansatz (AOA),
+including Givens circuit for efficient initial Dicke state preparation and xy mixer.
+
+All functions are decorated with ``@qmc.qkernel`` and use Handle-typed
+parameters so they can be composed inside other ``@qkernel`` functions.
+"""
```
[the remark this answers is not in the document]
baptistechev: @copilot apply changes based on [this feedback](https://github.com/Jij-Inc/Qamomile/pull/375#discussion_r3232747455)

qamomile/circuit/algorithm/state_preparation/dicke.py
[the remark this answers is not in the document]
baptistechev: Not sure I want to fix this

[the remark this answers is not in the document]
baptistechev: Not sure I want to fix that

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,517 @@
...
+#
+# ### Feasibility Check
+#
+# We can now check the advantage of AOA compare to QAOA.
+#  
+# The problem-tailored XY mixer that we implemented should
+# allow one to remain within the feasible space. Thus, 
+# unlike the QAOA case, **candidate solutions** proposed by AOA
+# should all live into the feasible space.
+# That means, they respect the constraint hamming weight $=1$
```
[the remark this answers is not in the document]
baptistechev: @copilot apply changes based on [this feedback](https://github.com/Jij-Inc/Qamomile/pull/375#discussion_r3232948263)

qamomile/optimization/aoa.py
```diff
@@ -0,0 +1,827 @@
...
+                    p=p,
+                    pair_indices_mixer=resolved_pair_indices_mixer,
+                    initial_ones=initial_ones,
+                )
+            case _:
+                assert False, (
+                    f"unreachable: quadratic={quadratic}, initial_state={initial_state!r}"
+                )
+
+    def _transpile_quadratic(
```
ahao27: Also for line 642 `def _transpile_hubo()`.

These methods coincidentally share the name of the parent's private helpers but are not used polymorphically. The override relationship is therefore accidental rather than intentional, and the diverging signature is what makes it a hard error for the type checker.

For comparison, `FQAOAConverter` extends `MathematicalProblemConverter` directly rather than inheriting from `QAOAConverter`, so it sidesteps this entirely.

**Options to fix (any one is enough):**

1. **Rename the AOA-specific helpers** so they no longer collide with the parent's private methods, e.g. `_transpile_aoa_quadratic`, `_transpile_aoa_hubo`. This is the smallest diff and keeps the current inheritance shape.
2. **Switch `AOAConverter` to compose `QAOAConverter`** (or extend `MathematicalProblemConverter` and call into a `QAOAConverter` instance just to reuse `get_cost_hamiltonian()`). 

Option 1 is the lowest-risk unblock; option 2 is the more durable answer. Either way, `uv run zuban check qamomile/` should be clean before merge.

qamomile/optimization/aoa.py
```diff
@@ -0,0 +1,827 @@
...
+                "quad": self.spin_model.quad,
+                "n": self.spin_model.num_bits,
+                "p": p,
+                "pair_indices_mixer": pair_indices_mixer,
+                "initial_ones": initial_ones,
+            },
+            parameters=["gammas", "betas"],
+        )
+
+    def _transpile_hubo(
```
ahao27: Check R464

qamomile/circuit/algorithm/__init__.py
ahao27: Could you check with  `uv run ruff check --fix qamomile/circuit/algorithm/state_preparation/__init__.py`

qamomile/optimization/aoa.py
```diff
@@ -0,0 +1,827 @@
...
+        num_blocks = self.spin_model.num_bits // block_size
+        match mixer_name:
+            case MixerName.RING:
+                return self._build_ring_pair_indices(num_blocks, block_size)
+            case MixerName.FULLY_CONNECTED:
+                return self._build_fully_connected_pair_indices(num_blocks, block_size)
+            case _:
+                assert False, f"unreachable: unhandled MixerName {mixer_name!r}"
+
+    def compute_dicke_composition_schedule(
```
ahao27, line 282: R248-270
Missing the docstring for "Raises" message.

tests/circuit/algorithm/test_aoa.py
ahao27, line 1: If could please add docstring for helper function also for test_dicke.py

ahao27: Also add the expval tests for `hubo_aoa_state_dicke` and `hubo_aoa_state_basis_state` .

qamomile/optimization/aoa.py
```diff
@@ -0,0 +1,827 @@
...
+        >>> converter = AOAConverter(model)
+        >>> executable = converter.transpile(
+        ...     QiskitTranspiler(),
+        ...     p=2,
+        ...     mixer="ring",
+        ...     block_size=3,
+        ... )
+    """
+
+    def _normalize_pair_indices(self, pair_indices: np.ndarray) -> np.ndarray:
```
ahao27, line 86: A row like (3, 3) passes this check and ends up at `xy_pair_rotation(q, 3, 3, beta)`, where the line
```
q[i], q[j] = qmc.cx(q[i], q[j])
```
trips the no-cloning / no-aliasing rule deep inside the IR layer. 

Suggest like
``` python
def _normalize_pair_indices(self, pair_indices: np.ndarray) -> np.ndarray:
    normalized = np.asarray(pair_indices, dtype=np.uint64)
    if normalized.ndim != 2 or normalized.shape[1] != 2:
        raise ValueError("pair_indices must have shape (num_pairs, 2).")

    self_pair_rows = np.flatnonzero(normalized[:, 0] == normalized[:, 1])
    if self_pair_rows.size:
        offending = normalized[self_pair_rows].tolist()
        raise ValueError(
            f"pair_indices contains self-pairs at rows {self_pair_rows.tolist()}: "
            f"{offending}. XY mixer pairs must connect two distinct qubits."
        )

    n = self.spin_model.num_bits
    if (normalized >= n).any():
        raise ValueError(
            f"pair_indices contains qubit indices >= num_bits ({n}); "
            f"max value present is {int(normalized.max())}."
        )

    return normalized
```

ahao27 (changes requested):
I have leave some comments

qamomile/circuit/algorithm/__init__.py
baptistechev: All checks passed!

tests/circuit/algorithm/test_aoa.py
```diff
@@ -0,0 +1,1005 @@
...
+    pass
+# cudaq imports ``torch`` at import time, whose OpenMP runtime segfaults
+# alongside qiskit-aer, so the collection-isolation guard
+# (tests/_cudaq_isolation.py) forbids a module-level ``import cudaq``.
+# ``importlib.util.find_spec`` probes availability without loading the runtime
+# (importing ``CudaqTranspiler`` alone does not pull in cudaq).
+if importlib.util.find_spec("cudaq") is not None:
+    from qamomile.cudaq.transpiler import CudaqTranspiler
+
+    BACKENDS.append(("cudaq", CudaqTranspiler))
```
jij-yuichiro-nakano, line 51: The QuriParts and CUDA-Q entries are appended as ordinary, unmarked parameters. The default pytest session excludes `quri_parts` and `cudaq`, while the dedicated jobs select only `-m quri_parts` or `-m cudaq`. As a result, these new AOA/Dicke tests are not executed by either optional-backend job.

I reproduced this on the latest head: the QuriParts command selected no tests (`206 deselected`), although the new HUBO Dicke and basis-state expectation-value cases pass when run explicitly with QuriParts.

Please attach the corresponding marks to the backend parameter values, for example with `pytest.param(..., marks=pytest.mark.quri_parts)` and `pytest.param(..., marks=pytest.mark.cudaq)`, or use marked backend fixtures. Please apply the same fix consistently to `test_aoa.py`, the Dicke tests, and `tests/optimization/test_aoa.py`, so both sampling and expectation-value paths actually run in each backend job.

tests/circuit/algorithm/test_aoa.py
```diff
@@ -0,0 +1,1005 @@
...
+
+
+def _cudaq_gate_counts(exe) -> dict[str, int]:
+    """Counts gates in the transpiled Cudaq circuit by name."""
+    source = exe.compiled_quantum[0].circuit.source
+    return {name: len(pat.findall(source)) for name, pat in _CUDAQ_PATTERNS.items()}
+
+
+@qmc.qkernel
+def _wrap_xy_pair_rotation(
```
jij-yuichiro-nakano, line 122: The previous review requested docstrings for the test helpers, but this commit documents only the gate-count and model-construction helpers. The module-level wrapper qkernels still have no docstrings—14 wrappers in this file and four in `tests/circuit/stdlib/state_preparation/test_dicke.py`.

Please add Google-style docstrings describing which qkernel each wrapper exercises and whether it exposes the sampling or expectation-value path. Wrapper qkernels with docstrings were already supported before this change, so this does not require a compiler workaround.

docs/ja/algorithm/index.md
```diff
@@ -8,6 +8,12 @@ Qamomileで実装した具体的な量子アルゴリズム例です。
 
 ::::{grid} 1 1 1 1
 
+:::{card}
+:header: **グラフ彩色のためのAlternating Operator Ansatz**
+:link: aoa_coloring_problem
+Dicke状態による初期状態とXYミキサーを使い、ワンホットの可行部分空間に留まったままK彩色問題をAlternating Operator Ansatzで解きます。
```
jij-yuichiro-nakano, line 14: Please revise the Japanese card title and description so that they use the Japanese name of the algorithm and state the tutorial’s purpose more clearly.

- Title: `グラフ彩色のための量子交互演算子アンザッツ`
- Description: `Dicke状態による初期状態とXYミキサーを使い、グラフ彩色問題の解を実行可能な解の空間上で探索します。`

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+# \sum_{i=0}^{K-1} x_{u, i}=1, \forall u \in\{0, \ldots, N-1\}
+# $$
+#
+# where $x_{u, i} \in\{0,1\}$ indicates whether color $i$ is used for vertex $u$ or not.
+
+# %% [markdown]
+# ## Define the Problem with JijModeling
+
+# %%
+import jijmodeling as jm
```
jij-yuichiro-nakano, line 78: Please revise this tutorial to follow the current algorithm-page conventions in `docs/skeletons/algorithm.py`, and mirror the resulting structure in the Japanese page.

- Organize the page around `Background`, `Problem Settings`, `Algorithm`, `Implementation`, `Result`, optional `Qamomile Built-ins`, and `Summary`.
- Place the installation cell first, followed immediately by a single import cell containing every library used in the tutorial.
- Remove italics that do not add meaningful emphasis, such as those around “block,” “inside each block,” “entire feasible space,” “parity,” “partition,” and “original.”
- Add assertions after each result-producing execution cell so that notebook execution verifies the behavior being demonstrated instead of relying only on displayed output.
- Reduce the final summary to at most three simple, non-technical take-home messages.

Please re-execute both the English and Japanese notebooks after applying these changes.

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+# ---
+# tags: [algorithm, optimization, variational]
+# ---
+#
+# # Alternating Operator Ansatz for Graph Coloring
+#
+# This tutorial demonstrates how to solve the **K - graph coloring problem**
+# with Qamomile using the Alternating Operator Ansatz (AOA).
+#
+# >Hadfield, S.; Wang, Z.; O’Gorman, B.; Rieffel, E.G.; Venturelli, D.; Biswas, R. From the Quantum Approximate Optimization Algorithm to a Quantum Alternating Operator Ansatz. Algorithms 2019, 12, 34. https://doi.org/10.3390/a12020034
```
jij-yuichiro-nakano, line 28: Please replace the manually formatted block quote and raw DOI URL with MyST’s DOI-backed citation syntax. For example:

```md
The AOA formulation used here follows Hadfield *et al.* {cite:p}`10.3390/a12020034`.
```

Please apply the same citation form to the Japanese page.

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+# ---
+
+# %% [markdown]
+# ---
+# tags: [algorithm, optimization, variational]
+# ---
+#
+# # Alternating Operator Ansatz for Graph Coloring
+#
+# This tutorial demonstrates how to solve the **K - graph coloring problem**
```
jij-yuichiro-nakano, line 25: Please refer to this simply as the “graph coloring problem,” rather than the “K-graph coloring problem.” Here, $K$ represents the number of available colors and is not part of the problem name.

For example:

“This tutorial demonstrates how to solve the graph coloring problem with Qamomile using the Alternating Operator Ansatz (AOA).”

docs/ja/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,515 @@
...
+# ---
+
+# %% [markdown]
+# ---
+# tags: [algorithm, optimization, variational]
+# ---
+#
+# # Alternating Operator Ansatzによるグラフ彩色問題
+#
+# このチュートリアルでは、Qamomile を用いて **K-グラフ彩色問題** を Alternating Operator Ansatz, 略してAOA で解く方法を紹介します。
```
jij-yuichiro-nakano, line 25: Please replace the opening sentence with the following wording:

`このチュートリアルでは、Qamomileを用いてグラフ彩色問題を量子交互演算子アンザッツ(AOA)で解く方法を紹介します。`

docs/ja/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,515 @@
...
+#
+# >Hadfield, S.; Wang, Z.; O'Gorman, B.; Rieffel, E.G.; Venturelli, D.; Biswas, R. From the Quantum Approximate Optimization Algorithm to a Quantum Alternating Operator Ansatz. Algorithms 2019, 12, 34. https://doi.org/10.3390/a12020034
+#
+# AOA アルゴリズムは、より一般的なミキサー (mixer) と初期状態を用いることで QAOA を拡張したものです。
```
jij-yuichiro-nakano, line 29: Please rewrite this explanation as follows so that the role of the mixer Hamiltonian is explicit:

`AOAは、より一般的なミキサーハミルトニアンと初期状態を用いることで、QAOA回路を拡張したものです。`

docs/ja/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,515 @@
...
+#
+# ### なぜこれがグラフ彩色問題で重要なのか？
+# ワンホット符号化では $N \times K$ 個の二値変数を用いますが、$2^{NK}$ 個のビット列のうち
```
jij-yuichiro-nakano, line 32: Please standardize this term as `one-hot` instead of `ワンホット` throughout the Japanese page.

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+# ### Why does this matter for graph coloring?
+# The one-hot encoding uses $N \times K$ binary variables but only $K^N$
+# of the $2^{NK}$ bitstrings are feasible (one color per node).
+# On the 5-node, 3-color instance below, that is 243 feasible states inside
+# a $2^{15} = 32768$-dimensional Hilbert space, i.e. less than $1\%$.
+# Standard QAOA starts in a uniform superposition over the full space and uses a transverse-field mixer
+# ($\sum X_i$) that freely rotates qubits in and out of the feasible subspace,
+# so most of the sampled bitstrings violate the one-hot constraint and get discarded.
+# AOA fixes both ends: it starts inside the feasible subspace and uses an XY mixer that
+# preserves Hamming weight, so every sample is feasible by construction.
```
jij-yuichiro-nakano, line 41: The XY mixer is central to why the ansatz remains in the feasible subspace, but its definition and role are not explained here. Please add a short mathematical explanation. For example, the pairwise interaction can be defined as

$$
H_{ij}^{XY}=\frac{1}{2}(X_iX_j+Y_iY_j),
\qquad
U_{ij}^{XY}(\beta)=e^{-i\beta H_{ij}^{XY}}.
$$

Please also explain that this interaction mixes the amplitudes of \(\lvert01\rangle\) and \(\lvert10\rangle\) without changing the number of ones. Applying it only within each node’s color block therefore preserves the one-hot constraint.

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+# The constraint $\sum_{i=0}^{K-1} x_{u,i} = 1$ says: among the $K$ binary variables attached to node $u$, exactly one is $1$. When we lay these out as a bitstring, each node occupies a *block* of $K$ consecutive qubits, and feasibility means **each block has Hamming weight exactly 1**.
+#
+# For our 5-node, 3-color instance the qubit layout is:
+#
+# ```
+#   [q0 q1 q2] [q3 q4 q5] [q6 q7 q8] [q9 q10 q11] [q12 q13 q14]
+#    node 0     node 1     node 2     node 3        node 4
+# ```
+#
+# A feasible state has one `1` in each bracketed block. This block structure is exactly what `block_size` will refer to later: we tell the converter to prepare a Dicke state *inside each block* and to mix *only within each block*, which keeps every block at Hamming weight 1 throughout the circuit.
```
jij-yuichiro-nakano, line 117: `block_size` has not been introduced at this point, so this forward reference interrupts the problem explanation. Please remove the sentence beginning “This block structure is exactly what `block_size` will refer to later” and explain the parameter where it is first used with `AOAConverter`.

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+
+# %% [markdown]
+# ## Graph Instance
+#
+# We use a fixed 5-node graph with 6 edges for reproducibility.
```
jij-yuichiro-nakano, line 122: “For reproducibility” feels unnecessary here because this is simply the graph selected for the tutorial. Please simplify the sentence to:

“In this tutorial, we use a fixed 5-node graph with 6 edges.”

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+
+# %%
+num_colors = 3
+instance_data = {"N": num_nodes, "E": edge_list, "K": num_colors}
+instance = graph_coloring_decorated.eval(instance_data)
+
+# %% [markdown]
+# ## Set Up the AOAConverter
+#
+# `AOAConverter` takes an OMMX instance and internally convert it into a QUBO form and build the Hamiltonian (see QAOA tutorial for more details).
```
jij-yuichiro-nakano, line 163: Please link the referenced QAOA tutorial so that readers can navigate directly to the explanation of the converter workflow. The most relevant destination here is:

`[QAOA for Graph Partitioning](qaoa_graph_partition)`

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+# %% [markdown]
+# ## Transpile to an Executable Circuit
+#
+# `converter.transpile()` works exactly as in the QAOA case.
+#
+# In addition, we can select what kind of XY mixer and initial state we want to use.
+#
+# For initial state, we can choose between:
+#
+# - `single_basis_state` : a single computational-basis state with the right Hamming weight in each block (e.g. $|100\rangle|100\rangle\ldots$, meaning every node starts colored with color 0). Feasible, deterministic, cheap to prepare. The mixer is responsible for spreading amplitude to other feasible states. This is the simplest choice and a good first thing to try.
```
jij-yuichiro-nakano, line 186: “Feasible, deterministic, cheap to prepare” is terse and may be unclear to tutorial readers. Please replace it with a plain explanation such as:

“This starts from one valid coloring and requires relatively few gates to prepare.”

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+#
+# - `dicke` : a Dicke state in each block: an equal superposition over all bitstrings with the chosen Hamming weight. For Hamming weight 1 and block size $K$, this is $\frac{1}{\sqrt{K}}(|10\ldots 0\rangle + |01\ldots 0\rangle + \cdots + |0\ldots 01\rangle)$ in each block, so the full initial state is already a uniform superposition over the *entire feasible space*. Costs extra gates to prepare ($O(K)$ XY rotations per block) but gives the optimizer an unbiased starting point.
```
jij-yuichiro-nakano, line 188: This Dicke-state expression is too long to read comfortably as inline mathematics. Please move it into a separate display-math block.

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+#
+# - `uniform` : Hadamard on every qubit, the standard QAOA initial state. **This puts amplitude on infeasible states**, so the feasibility guarantee of AOA is lost; samples will need to be filtered like in plain QAOA. Useful mainly as a baseline or for ablations comparing initial states.
```
jij-yuichiro-nakano, line 190: “Ablations” is not explained and is unnecessarily specialized for this tutorial. Please replace it with a plain phrase such as “comparisons of how different initial states affect the result.”

docs/ja/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,515 @@
...
+#
+# `hamming_weight` パラメータは、各ブロックでの目標ハミング重みを設定します。ワンホット符号化を用いたグラフ彩色問題では常に `1` になります。他の問題（例えば、ちょうど $k$ 個のアイテムを選択する必要がある基数制約付き最適化問題）では、これは異なる値になります。
+#
+# ミキサーとしては、以下から選択できます：
+#
+# - `ring` : 各量子ビットをブロック内のサイクル状の 2 つの隣接量子ビットとのみ接続します。*パリティ* 分解（奇数ペア、次に偶数ペア、最後にラップアラウンド）として実装されているため、各層のコストはブロックあたり $O(K)$ の 2 量子ビットゲートとなります。回路は浅くなりますが、振幅がブロック全体に広がるにはより多くの層が必要です。
+#
+# - `fully-connected` : ブロック内のすべての量子ビットのペアを接続します。*パーティション* 分解（$\binom{K}{2}$ ペアを重ならないラウンドに分割）として実装されているため、各層のコストはブロックあたり $O(K^2)$ の 2 量子ビットゲートとなります。回路は深くなりますが、振幅は単一の層でブロック全体に混合されます。
+#
+# 経験則として：$K$ が大きく回路深さがボトルネックになる場合は `ring` を、$K$ が小さく（例えば $\leq 4$）、より速いミキシングと多様なサンプルが欲しい場合は `fully-connected` を選ぶとよいでしょう。実際、小さなインスタンスでは `fully-connected` の方が 1 ショットあたりにより多くの異なる実行可能解を生成する傾向があります。
```
jij-yuichiro-nakano, line 199: The Japanese page uses both `ミキシング` and `混合` for the mixing process. Please standardize the prose on `混合` throughout the page. The component name `XYミキサー` can remain unchanged.

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+
+import numpy as np
+from qiskit_aer import AerSimulator
+from scipy.optimize import minimize
+
+# Seed the simulator so re-executing the notebook reproduces the same
+# COBYLA trajectory and final sampling distribution. Without a seed,
+# every shot draws fresh randomness, COBYLA sees a noisy cost surface,
+# and each notebook run converges to a different (but equivalent) local
+# optimum.
```
jij-yuichiro-nakano, line 338: This extended explanation of the simulator seed and the COBYLA trajectory is not needed for the tutorial’s main learning goal. Please remove this comment block. The seed itself can remain in the code.

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+        shots=sample_shots,
+        bindings={"gammas": gammas, "betas": betas},
+    )
+    result = job.result()
+    # decode_to_binary_sampleset returns the QUBO-domain BinarySampleSet
+    # whose `energy` is the penalized objective — what COBYLA needs to
+    # see infeasibility costs. The polymorphic decode() returns an
+    # ommx.v1.SampleSet whose `objective` is the un-penalized true
+    # objective; using it here would let the optimizer settle on
+    # infeasible all-zero / all-one bitstrings.
```
jij-yuichiro-nakano, line 366: This detailed comparison of the internal decoding paths is beyond the scope of this tutorial and interrupts the optimization workflow. Please remove this comment block and keep the example focused on how the cost is evaluated.

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+    executor,
+    shots=1000,
+    bindings={"gammas": gammas_opt, "betas": betas_opt},
+).result()
+
+# `decode()` on an OMMX-backed converter returns an `ommx.v1.SampleSet`
+# evaluated against the *original* (un-penalized) instance, so
+# feasibility, the true objective, and per-constraint diagnostics are
+# available through OMMX's own API — no hand-rolled feasibility or
+# objective helper is needed.
```
jij-yuichiro-nakano, line 412: This implementation-level explanation of the OMMX-backed `decode()` behavior is not needed for the tutorial’s main flow. Please remove this comment block.

docs/en/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,525 @@
...
+#
+# ### Feasibility Check
+#
+# We can now check the advantage of AOA compared to QAOA.
+#
+# The problem-tailored XY mixer that we implemented should
+# allow one to remain within the feasible space. Thus,
+# unlike the QAOA case, **candidate solutions** proposed by AOA
+# should all live into the feasible space.
+# That means, they respect the constraint hamming weight $=1$
```
jij-yuichiro-nakano, line 426: It is unclear what “one” refers to in “allow one to remain within the feasible space.” Please make the quantum state the explicit subject and simplify the explanation. For example:

“The problem-specific XY mixer should keep the quantum state within the feasible subspace throughout the search. Therefore, unlike standard QAOA, every candidate sampled by this AOA circuit should satisfy the Hamming-weight-one constraint.”

docs/ja/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,515 @@
...
+    if row["value"] > 0.5:
+        best_coloring[int(node)] = int(color)
+
+print("Best coloring:", best_coloring)
+print("Number of Conflicts (neighbor nodes with the same color):", int(round(best.objective)))
+
+# %% [markdown]
+# ### 目的関数値の分布
+#
+# 真の目的関数値（衝突数）の分布をプロットします。
```
jij-yuichiro-nakano, line 453: `衝突数` is used here without being introduced. Please either replace it with a plain description such as `同じ色が割り当てられた隣接頂点ペアの数`, or define a “conflict” in the problem-setting section as an edge whose endpoints have the same color.

Please then use the selected wording consistently in the paragraph, printed output, and plot labels.

docs/ja/algorithm/aoa_coloring_problem.py
```diff
@@ -0,0 +1,515 @@
...
+plt.xlabel("Conflicts (objective value)")
+plt.ylabel("Frequency")
+plt.title("Distribution of Solutions")
+plt.show()
+
+
+# %% [markdown]
+# ### 最良の 3 彩色の可視化
+#
+# 交互演算子アンザッツアルゴリズムで見つかった最良の実行可能彩色に従って、グラフのノードを色付けします。
```
jij-yuichiro-nakano, line 469: Please replace `最良の実行可能彩色` with the simpler `最良の彩色パターン`. Feasibility has already been established in the preceding section, so repeating it here makes the wording unnecessarily technical.

jij-yuichiro-nakano (changes requested):
@baptistechev
Thank you for your work on this PR. I left a few comments on the remaining test coverage, helper documentation, and tutorial updates.

tests/circuit/algorithm/test_aoa.py
baptistechev: Fixed in 07a69c05. The QuriParts and CUDA-Q backends are now `pytest.param(..., marks=pytest.mark.quri_parts)` / `pytest.param(..., marks=pytest.mark.cudaq)` in `tests/circuit/algorithm/test_aoa.py`, `tests/circuit/stdlib/state_preparation/test_dicke.py` and `tests/optimization/test_aoa.py`, so the sampling and expval cases run in each backend job. Locally: default run 116 passed, `-m quri_parts` 101 passed, `-m cudaq` 101 passed (the CUDA-Q gate-count helper fix from f6fa4c96 makes the cudaq job pass).

tests/circuit/algorithm/test_aoa.py
baptistechev: Fixed in 07a69c05. All 14 wrapper qkernels in `test_aoa.py` and the 4 in `test_dicke.py` now have Google-style docstrings. Each one says which qkernel it exercises and whether it goes through the sampling or the expectation-value path, with `Args` / `Returns`.

docs/ja/algorithm/index.md
baptistechev: Done in 5fea3ae9. The card title is now `グラフ彩色のための量子交互演算子アンザッツ` and the description is `Dicke状態による初期状態とXYミキサーを使い、グラフ彩色問題の解を実行可能な解の空間上で探索します。`

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9, for both en and ja. The page now follows `docs/skeletons/algorithm.py`: Background, Problem Settings, Algorithm, Implementation, Result, Summary. There is no Qamomile Built-ins section because the page already uses `AOAConverter`. The install cell comes first, followed by one import cell. I removed the italics you listed and added assertions after each result cell: graph and instance sizes, qubit count, Dicke/mixer schedule shapes, optimizer history, `SampleSet` type, shot count, 100% feasibility, and the best coloring's conflict count against `best.objective`. The summary is now three short take-home messages. Both notebooks were re-executed without `QAMOMILE_DOCS_TEST`.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. It now reads: The AOA formulation used here follows Hadfield *et al.* {cite:p}`10.3390/a12020034`. The Japanese page uses `Hadfieldら{cite:p}`10.3390/a12020034``.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. The opening sentence is now the wording you suggested, and "K-coloring" / "K - graph coloring" is gone from the page, the summary and the index card.

docs/ja/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. The opening sentence is now the wording you suggested.

docs/ja/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. The Background section now opens with that sentence.

docs/ja/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. The Japanese page and card now use `one-hot` throughout.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Added in 5fea3ae9, in the new Algorithm section. It defines $H_{ij}^{XY}$ and $U_{ij}^{XY}(\beta)$ as you suggested, which matches the `xy_pair_rotation` implementation. It then explains that $H_{ij}^{XY}$ swaps $|01\rangle \leftrightarrow |10\rangle$ and annihilates $|00\rangle$ and $|11\rangle$, so the mixer keeps the number of ones, and that applying it only inside each node's color block keeps the one-hot constraint. The same explanation is in the Japanese page.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. I removed the forward reference. `block_size` is now explained where it is first used, just before the `AOAConverter.transpile` call.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. The sentence now reads "In this tutorial, we use a fixed 5-node graph with 6 edges." The Japanese page matches.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. The converter section now links [QAOA for Graph Partitioning](qaoa_graph_partition). The Japanese page links `[QAOAによるグラフ分割](qaoa_graph_partition)`.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. It now reads "This starts from one valid coloring and requires relatively few gates to prepare." While editing I also fixed the example: the converter sets the last qubit of each block, so the state is $|001\rangle|001\rangle\ldots$ (every node starts with the last color), not $|100\rangle\ldots$ as the page said before.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. The Dicke state $|D^K_1\rangle$ is now a display-math block in the Algorithm section.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. It now says "comparisons of how different initial states affect the result".

docs/ja/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. The prose now uses `混合` throughout, and `XYミキサー` is unchanged.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Removed in 5fea3ae9. The seed stays in the code.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Removed in 5fea3ae9. `cost_fn` now only samples, decodes and returns the mean energy.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Removed in 5fea3ae9.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. The paragraph now uses your wording, with the quantum state as the subject. The notebook also asserts right after it that every sample is feasible (1000 / 1000 in the executed notebook).

docs/ja/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. The Problem Settings section now defines `衝突` as an edge whose endpoints share a color and says the objective is the number of conflicts. The same term is used in the paragraph, the printed output (`衝突の数（同じ色が割り当てられた隣接頂点ペアの数）`) and the plot description. The plot axis labels are in English, as on the other Japanese algorithm pages. The English page defines "conflict" in the same place.

docs/ja/algorithm/aoa_coloring_problem.py
baptistechev: Done in 5fea3ae9. The heading and the text now use `最良の彩色パターン`.

docs/en/algorithm/aoa_coloring_problem.py
baptistechev: Follow-up in a832df9f: the drawing cells (full AOA circuit, Dicke preparation, cost layer, XY mixer), the cost-history plot and the final coloring plot now have assertions too, so every result cell checks something. Both notebooks were re-executed without `QAMOMILE_DOCS_TEST`.

PR #376, opened 2026-05-07
Add sample-based subspace diagonalization

yuyamashiro:
## Summary
- Add `qamomile.linalg.subspace` with three public helpers for hybrid quantum-classical workflows that build an effective Hamiltonian from sampled bitstrings (Z basis) or mixed-Pauli-basis product states and diagonalise it classically:
  - `subspace_hamiltonian` — Z-basis vectorised builder (XOR/parity, no matrix products) for plain QSCI.
  - `generalized_subspace_matrices` — returns `(H_sub, S_sub)` for non-orthogonal product-state bases (X / Y / Z mix).
  - `solve_subspace` — end-to-end (generalised) eigenvalue solve with a low-rank fallback for ill-conditioned overlap matrices.
- Re-export the new helpers from `qamomile.linalg.__init__`.
- Add a tutorial under `docs/en/vqa/qsci` walking through the four-step QSCI recipe on a four-qubit transverse-field Ising model with the QURI Parts backend.

## Test plan
- [x] `uv run pytest tests/linalg/` — 265 passed (46 new tests for `test_subspace.py`).
- [x] `uv run pytest tests/` — 6684 passed (no regressions).
- [x] `uv run pytest -m docs -k "en/vqa/qsci" -v` — tutorial executes without error.
- [x] `uv run ruff check qamomile/linalg/ tests/linalg/` — clean.
- [x] Public API: `from qamomile.linalg import subspace_hamiltonian, generalized_subspace_matrices, solve_subspace` works.


yuyamashiro:
Addressed both CI failures and the Copilot inline review in commit b355bdf0:

**CI**
- ruff-format: applied `ruff format` to the new `subspace.py` / `test_subspace.py`.
- base-test: registered `qsci` under `OPTIONAL_SKIP_MODULES` in `tests/docs/test_tutorials.py` so jobs without the `quri_parts` extra cleanly skip the tutorial (matches the existing `vqe_for_hydrogen` → `openfermion` pattern).

**Copilot inline comments**
1. **scipy dependency** — `solve_subspace` no longer touches SciPy. The generalised eigen-problem `H_sub v = λ S_sub v` is now always reduced to a standard one by whitening `S_sub` with its eigendecomposition (the same projection that previously handled only the rank-deficient fallback). Module is NumPy-only now.
2. **`diff` XOR cached** — hoisted `diff = bits[:, None, :] ^ bits[None, :, :]` out of the per-term loop in `subspace_hamiltonian`.
3. **`overlap_per_q.copy()` per term** — `generalized_subspace_matrices` no longer materialises a fresh `(n, n, num_qubits)` tensor each term. Per-term contribution is now `prod(overlap on untouched qubits) × prod(Pauli factors on touched qubits)`.
4. **`overlap_tol` swallows everything** — `solve_subspace` now raises `ValueError` (with the largest seen overlap eigenvalue and a hint) when `keep` is all-False instead of returning empty arrays. Added regression test `test_solve_raises_when_overlap_tol_kills_all_directions`.
5. **`hash()` seed in test** — replaced with deterministic `1000 + 10 * num_qubits + basis.value`.
6. **docs / quri_parts guard** — handled at the test-discovery level via `OPTIONAL_SKIP_MODULES` rather than wrapping every import in the tutorial, keeping the tutorial readable while still skipping cleanly when `quri_parts` is absent.

Local checks after the changes:
- `uv run pytest tests/` → 6685 passed, 96 skipped
- `uv run pytest -m docs -k "en/vqa/qsci"` → 1 passed (skipped on systems without `quri_parts`)
- `uv run ruff check qamomile/ tests/` → clean
- `uv run ruff format --check qamomile/ tests/` → clean

PR #382, opened 2026-05-11
Add BinaryModel.from_higher_ising() for higher-order SPIN input

ahao27:
## Summary

- Add `BinaryModel.from_higher_ising()` classmethod that builds a SPIN model from arbitrary-order Ising coefficients (linear, quadratic, cubic, quartic, ...) packed into a single `dict[tuple[int, ...], float]`. Fills the SPIN-side gap left by the existing `from_hubo()` (BINARY-side higher-order factory).
- Apply SPIN idempotency (`s_i^2 = 1`) when reducing repeated indices within a term: pairs cancel, so `(0, 0, 2) → (2,)` and `(0, 0, 1, 1, 2) → (2,)`. Fully-paired terms (e.g. `(0, 0)`) collapse into the constant. A `UserWarning` is emitted on any reduction. This mirrors the SPIN branch of `BinaryExpr._reduce_indices`.
- Empty tuples `()` accumulate into the constant term, and `simplify=True` strips near-zero coefficients post-accumulation — matching the `from_hubo` API surface so callers can switch between the two factories with no other changes.

## Test plan

- [x] 11 unit tests + 50 randomized parametrizations under `tests/optimization/test_binary_model.py::test_from_higher_ising_*` (61 cases, all pass).
- [x] SPIN → BINARY → SPIN round-trip preserves energy (4th- and 5th-order coverage, `atol=1e-10`).
- [x] `from_higher_ising` agrees with `from_ising` on quadratic-only inputs (linear, quad, higher dicts all equal).
- [x] Duplicate-index reduction paths: one pair, multiple pairs, full cancel-to-constant, partial cancel — each verified with `pytest.warns(UserWarning, ...)`.
- [x] `simplify=True` removes near-zero coefficients before construction.
- [x] `/local-review` reports no findings; `/ultrareview` reports no findings.


qamomile/optimization/binary_model/model.py
```diff
@@ -169,6 +169,70 @@ def from_hubo(
...
+            ... )
+            >>> model.vartype
+            <VarType.SPIN: 'SPIN'>
+        """
+        expr = BinaryExpr(vartype=VarType.SPIN, constant=constant, coefficients={})
+        for indices, coeff in higher_ising.items():
+            counts: dict[int, int] = {}
+            for i in indices:
+                counts[i] = counts.get(i, 0) + 1
+            reduced = tuple(sorted(i for i, c in counts.items() if c % 2 == 1))
```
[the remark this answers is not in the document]
ahao27: Promoted `BinaryExpr._reduce_indices` to a public
`@staticmethod` `BinaryExpr.reduce_indices(vartype, inds)` and routed
both `from_hubo` and `from_higher_ising` through it, so the BINARY/SPIN
idempotency rules are now shared with `BinaryExpr.__imul__`.

tests/optimization/test_binary_model.py
```diff
@@ -234,6 +234,227 @@ def test_from_hubo_with_simplify():
...
+            higher_ising={(0, 0): 2.5, (1,): 1.0}, constant=0.75
+        )
+
+    # (0,0) collapses to 1 → adds to constant: 0.75 + 2.5 = 3.25
+    assert np.isclose(bm.constant, 3.25)
+    assert bm.num_bits == 1
+    assert np.isclose(bm.linear[0], 1.0)
+
+
+def test_from_higher_ising_duplicate_indices_partial_cancel():
```
jij-yuichiro-nakano: It seems to me that this test is essentially testing the same behavior as the previously introduced `test_from_higher_ising_duplicate_indices_multi_pair_cancel()`. If there is a difference between these two tests, I would appreciate it if you could explain the intention behind each of them.

jij-yuichiro-nakano (changes requested):
Thank you for the PR! I think this is a very concise and smart implementation! I left a few comments, so please take a look.

tests/optimization/test_binary_model.py
```diff
@@ -234,6 +234,227 @@ def test_from_hubo_with_simplify():
...
+    """from_higher_ising should treat empty-index term () as a constant contribution."""
+    bm = BinaryModel.from_higher_ising(
+        higher_ising={(): 1.25, (0, 2): 2.0}, constant=0.75
+    )
+    assert np.isclose(bm.constant, 2.0)  # 0.75 + 1.25
+    assert bm.num_bits == 2
+    assert np.isclose(bm.quad[(0, 1)], 2.0)
+
+
+def test_from_higher_ising_duplicate_indices_pair_cancel():
```
jij-yuichiro-nakano, line 311: The four `test_from_higher_ising_duplicate_indices_*` tests confirm that `UserWarning` is raised whenever a term's indices are reduced, but no test confirms that the warning is not raised on the common case where every term's indices are already unique (e.g. `test_from_higher_ising_manually`, `_4th_order`, `_5th_order`, `_duplicate_accumulation`, `_with_simplify`, the 50-seed _equivalence_random, etc.).

Combined with the fact that there is no `filterwarnings = error` in `pyproject.toml`, a future change to `reduce_indices` that spuriously fires the warning on non-duplicate input would slip through the suite silently.

ahao27: Added test_from_higher_ising_no_warning_on_unique_indices, which wraps construction in warnings.catch_warnings() + simplefilter("error") so any spurious UserWarning on already-unique input escalates to a test failure.

tests/optimization/test_binary_model.py
ahao27: Thanks! They were redundant; both exercised the same shape. I've repurposed this test into `test_from_higher_ising_duplicate_indices_odd_count_above_two`.

jij-yuichiro-nakano (approved):
Thank you for updating! It looks good to me!

PR #387, opened 2026-05-15
Enforce affine type rule for Vector broadcast gates

ksk-jij:
## Summary

- Fixes an affine-type-rule asymmetry: `qmc.h(qs)` on a `Vector[Qubit]` previously left the Vector handle reusable when the return value was dropped, while the scalar `qmc.h(q)` correctly raised `QubitConsumedError`. The Vector broadcast path now consumes its input handle up-front, mirroring `_measure_vector_qubit` and `pauli_evolve`.
- Three frontend helpers updated in `qamomile/circuit/frontend/operation/qubit_gates.py`: `_broadcast_single_qubit_gate`, `_broadcast_rotation_gate`, `_broadcast_phase_gate`. `ArrayBase.consume` runs `validate_all_returned` internally, so the previous explicit pre-check is now redundant and removed.
- Emitted IR is unchanged: `consume()` returns a new handle wrapping the same `ArrayValue`, so the `ForOperation` and child gate ops are identical. Existing broadcast sampling / expval / loop-equivalence tests continue to pass without modification.
- Docstrings now describe the consume contract and list `QubitConsumedError` in `Raises`, with types spelled out in `Args` / `Returns` per the project's Google-style mandate.

## Test plan

- [x] New `TestBroadcastConsumesVectorHandle` (14 cases) parametrizes over all 12 broadcast gates (`h`, `x`, `y`, `z`, `s`, `sdg`, `t`, `tdg`, `rx`, `ry`, `rz`, `p`), asserts `QubitConsumedError` on return-drop, and confirms the canonical `qs = qmc.h(qs)` idiom still compiles cleanly with the expected `ForOperation` count.
- [x] TDD verified: the new tests fail on the previous implementation (13 fail / 1 pass) and pass on the fix (14 pass).
- [x] Full `tests/circuit/` suite: 4601 passed / 0 failed.
- [x] `ruff check` and `zuban check` on the changed files are clean.

### DIFF
--- a/tests/circuit/test_gate_broadcast.py
+++ b/tests/circuit/test_gate_broadcast.py
@@ -654,8 +654,11 @@ class TestBroadcastConsumesVectorHandle:
         """The canonical ``qs = qmc.gate(qs)`` pattern remains valid for every gate.
 
         Sanity check that the new consume contract does not regress the
-        common reassign idiom — every gate type chains through a single
-        kernel and the resulting block is non-empty.
+        common reassign idiom: every gate type chains through a single
+        kernel, lowers to one ``ForOperation`` per call in source order,
+        and each loop body contains exactly one ``GateOperation`` of the
+        matching gate kind (no spurious or reordered ops introduced by
+        the consume-at-start refactor).
         """
 
         @qmc.qkernel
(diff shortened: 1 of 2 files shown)


tests/circuit/test_gate_broadcast.py
```diff
@@ -572,3 +573,110 @@ def test_rx_broadcast_rejects_non_qubit_target(self):
...
+            qs = qmc.ry(qs, 0.4)
+            qs = qmc.rz(qs, 0.5)
+            qs = qmc.p(qs, 0.6)
+            return qmc.measure(qs)
+
+        assert _good.block is not None
+        fors = [op for op in _good.block.operations if isinstance(op, ForOperation)]
+        # 12 broadcast gate calls → 12 ForOperations (one per gate); measure
+        # lowers to a single MeasureVectorOperation, not a ForOperation.
+        assert len(fors) == 12
```
ksk-jij: Could you check the internal operations as well? Like in the ForOperation, we need to check if those operations in it are as same as expected.

ksk-jij: Fixed in 799a6d43: extended `test_broadcast_reassign_pattern_compiles_cleanly` to verify, for each of the 12 chained broadcast calls, that the matching `ForOperation` body holds exactly one `GateOperation` of the expected gate kind in source order. This catches regressions where the consume-at-start refactor would silently reorder gates, swap gate types, or introduce spurious ops inside the broadcast loop.

PR #393, opened 2026-05-19
docs(integration): add OMMX Quantum Benchmarks + QAOA tutorial (EN/JA)

jij-yuichiro-nakano:
## Summary
Add a new `usage` tutorial that drives QAOA on a LABS instance loaded straight from the OMMX Quantum Benchmarks dataset and compares the result against the classical SCIP solver via `ommx-pyscipopt-adapter`. Both paths consume the *same* `ommx.v1.Instance`, so the only thing that differs is the algorithm itself — making this a clean apples-to-apples end-to-end benchmark.

## What was added
- New tutorial paired sources in EN and JA: `docs/en/usage/ommx_quantum_benchmarks_qaoa.{py,ipynb}` and `docs/ja/usage/ommx_quantum_benchmarks_qaoa.{py,ipynb}`. Each `.ipynb` is regenerated from its `.py` via `jupytext --update --to ipynb` and ships with execution outputs.
- Index entries in `docs/en/usage/index.md` and `docs/ja/usage/index.md`.
- `OPTIONAL_SKIP_MODULES` entry (`ommx_quantum_benchmarks_qaoa → ommx_quantum_benchmarks`) in `tests/docs/test_tutorials.py` so the docs-test harness skips the new tutorial gracefully when the optional dataset library is not installed.
- `ommx-quantum-benchmarks>=0.0.2` added to the `dev` dependency group in `pyproject.toml` (the 0.0.1 release is not compatible with `minto>=2.0`); `uv.lock` regenerated.
- `.gitignore` entry for `.minto_experiments/`, the side-effect directory that `minto.Experiment` creates when the dataset loader pulls instances from the GitHub Container Registry.

## What can now be done
Any QOBLIB instance distributed through OMMX Quantum Benchmarks can be loaded as an `ommx.v1.Instance` and run through Qamomile's QAOA pipeline (`Instance.to_qubo()` → `BinaryModel` → hand-written `@qkernel` ansatz → `QiskitTranspiler` + `AerSimulator`) and the classical SCIP path (`ommx_pyscipopt_adapter.OMMXPySCIPOptAdapter.solve`) against the same instance, yielding directly comparable `.objective` values together with the bundled reference optimum.

## Test plan
- [x] `uv run pytest -m docs -k ommx_quantum_benchmarks_qaoa -v` — 4 tests passed (tag-whitelist EN+JA, tutorial-execution EN+JA).
- [x] Both `.py` files run end-to-end with `QAMOMILE_DOCS_TEST=1` against `ommx-quantum-benchmarks==0.0.2`, producing `Reference E* = 2` and `SCIP E(s) = 2`.
- [x] `.ipynb` files contain execution outputs in 9/11 code cells (the 2 without outputs are the commented `!pip install` cell and the `@qkernel` definition block, neither of which produces output by construction).


docs/en/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,483 @@
...
+#
+# **OMMX** ([Open Mathematical prograMming eXchange](https://jij-inc.github.io/ommx/en/introduction.html))
+# is a data format for exchanging mathematical optimization problems across
+# tools — its `ommx.v1.Instance` carries the objective, constraints,
+# decision-variable metadata, and (optionally) a reference solution.
+#
+# **OMMX Quantum Benchmarks** is a curated collection of optimization
+# benchmark instances distributed in this `ommx.v1.Instance` format. The
+# first dataset shipped is **QOBLIB** (Quantum Optimization Benchmarking
+# Library) [](https://arxiv.org/abs/2504.03832), which contains nine
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Fixed in 6505dea712e8072848e6c278a764b1bf5c87f941: replaced the empty-text markdown link with a named link `[arXiv:2504.03832](https://arxiv.org/abs/2504.03832)`. The `.py` is the source of truth and there is only one arXiv reference in the file (line 51 in the original); line 96 is the same reference rewrapped after the edit, not a separate instance.

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,376 @@
...
+# %%
+# 本チュートリアルで追加で必要なパッケージのインストール
+# # !pip install qamomile ommx-quantum-benchmarks ommx-pyscipopt-adapter
+
+# %% [markdown]
+# ## OMMX Quantum Benchmarksとは?
+#
+# **OMMX** ([Open Mathematical prograMming eXchange](https://jij-inc.github.io/ommx/en/introduction.html))は、数理最適化問題をツール間で受け渡すためのデータフォーマットです。`ommx.v1.Instance`には目的関数、制約条件、決定変数のメタデータ、そして必要に応じて参照解が格納されます。
+#
+# **OMMX Quantum Benchmarks**は、この`ommx.v1.Instance`形式で配布される最適化ベンチマークインスタンス集です。最初に提供されているデータセットは**QOBLIB** (Quantum Optimization Benchmarking Library) [](https://arxiv.org/abs/2504.03832)で、近年の量子最適化研究で頻繁に取り上げられる9つの問題ファミリーを収録しています。具体的にはLABS、Market Split、Independent Set、Steiner Tree Packingなどが含まれます。
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Fixed in 6505dea712e8072848e6c278a764b1bf5c87f941: replaced the empty-text markdown link with a named link `[arXiv:2504.03832](https://arxiv.org/abs/2504.03832)`. The `.py` has only one arXiv reference (line 35); line 62 refers to the paired `.ipynb`, which was regenerated via `jupytext --update --to ipynb` so EN and JA stay in sync.

docs/en/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,483 @@
...
+
+# %%
+def evaluate_with_ommx(
+    sample_result, spin_model: BinaryModel, ommx_instance: ommx.v1.Instance
+) -> ommx.v1.SampleSet:
+    """Decode SPIN samples, flip to BINARY, and evaluate against the OMMX instance."""
+    binary_ss = spin_model.decode_from_sampleresult(sample_result)
+    ommx_samples = ommx.v1.Samples({})
+    next_id = 0
+    for sample, occ in zip(binary_ss.samples, binary_ss.num_occurrences):
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Fixed in 27af32c30f6a41912f53ad5ae0f14f939da0dd86: renamed `binary_ss` to `spin_ss` in `evaluate_with_ommx` so the variable name matches the actual SPIN-domain contents of `spin_model.decode_from_sampleresult(...)`. The loop now reads `for sample, occ in zip(spin_ss.samples, spin_ss.num_occurrences)`, and the explicit `x = (1 - s) / 2` SPIN -> BINARY conversion stays in place. Paired `.ipynb` regenerated.

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,376 @@
...
+
+# %%
+def evaluate_with_ommx(
+    sample_result, spin_model: BinaryModel, ommx_instance: ommx.v1.Instance
+) -> ommx.v1.SampleSet:
+    """SPINサンプルをデコードし、BINARYに変換してOMMXインスタンスで評価する。"""
+    binary_ss = spin_model.decode_from_sampleresult(sample_result)
+    ommx_samples = ommx.v1.Samples({})
+    next_id = 0
+    for sample, occ in zip(binary_ss.samples, binary_ss.num_occurrences):
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Fixed in 27af32c30f6a41912f53ad5ae0f14f939da0dd86: renamed `binary_ss` to `spin_ss` in `evaluate_with_ommx` so the variable name matches the actual SPIN-domain contents of `spin_model.decode_from_sampleresult(...)`. The loop now reads `for sample, occ in zip(spin_ss.samples, spin_ss.num_occurrences)`, and the explicit `x = (1 - s) / 2` SPIN -> BINARY conversion stays in place. Paired `.ipynb` regenerated.

docs/en/usage/index.md
ksk-jij, line 1: [question]
I feel like this can go under integration since OMMX Quantum Benchmarks is not a part of Qamomile at lease and usage feels like more usage of Qamomile. What do you think?

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,376 @@
...
+# %% [markdown]
+# ---
+# tags: [usage, optimization, variational]
+# ---
+#
+# # OMMXベンチマークの活用 (1): Qamomileによる量子アルゴリズムの実装とベンチマーク
+#
+# 本チュートリアルでは、Qamomileで構築した量子アルゴリズムを公開ベンチマークデータセットから取得した問題で動かし、その解の品質を古典ソルバーと同一のパイプライン上で比較する方法を示します。
+#
+# **目標。** Qamomileで自前のQAOAソルバーを構築し、[OMMX Quantum Benchmarks](https://github.com/Jij-Inc/OmmxQuantumBenchmarks)データセットからロードした**Low Autocorrelation Binary Sequences (LABS)** インスタンスに対して実行し、[`ommx-pyscipopt-adapter`](https://github.com/Jij-Inc/ommx-pyscipopt-adapter)経由で利用できる古典ソルバーSCIPと結果をベンチマークします。QAOAパスとSCIPパスはどちらも*同一*の`ommx.v1.Instance`を入力として受け取るため、両者の差は実質的にアルゴリズムそのものだけになり、フェアな比較が可能です。
```
ksk-jij: [my opinion]
"目標。" feels a bit off as Japanese. Could you update here? Like "目標："?

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,376 @@
...
+#     language: python
+#     name: python3
+# ---
+
+# %% [markdown]
+# ---
+# tags: [usage, optimization, variational]
+# ---
+#
+# # OMMXベンチマークの活用 (1): Qamomileによる量子アルゴリズムの実装とベンチマーク
```
ksk-jij: [question]
Is there any specific reason that you named this (1)? Are you planning to create the second at this moment? If you are not planning to create a new one, I think for now we just say "OMMXベンチマークの活用" since we always can update the name when needed.

ksk-jij: [my opinion]
In this documentation, we only use `ommx-quantum-benchmarks`. Thus, it might be nice to say "OmmxQuantumBenchmarksの活用" or something like clearly mentioning ommx-quantum-benchmarks. The current name could be for any ommx benchmarks.

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
ksk-jij: [question]
Is there any reason that you didn't use `QAOAConverter`? I believe this documentation is for describing how to integrate OMMX Quantum Benchmarks with Qamomile. So, in that sense, using `QAOAConverter` is more easier to get it?

As you might think, in that case, the reader might not understand how to corporate OMMX Quantum Benchmarks with their own qkernel, but we have [BinaryModelの使い方](https://jij-inc-qamomile.readthedocs-hosted.com/latest/ja/usage/binary-model/) page. So, we can put the link at the appropriate section/cell.

What do you think?

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,376 @@
...
+# \qquad k = 1, 2, \dots, n-1
+# $$
+#
+# をできるだけ0に近づけるものを求める問題です。ベンチマークの目的関数は**自己相関の二乗和**
+#
+# $$
+# E(\boldsymbol{s}) = \sum_{k=1}^{n-1} c_k(\boldsymbol{s})^2,
+# $$
+#
+# であり、これを*最小化*します(等価的に、**メリットファクター** $F = n^2 / (2 E)$を最大化することに相当します)。LABSはNP-hardであり、古典・量子の両方のヒューリスティクスのストレステストとして長く使われてきました。
```
ksk-jij: [question]
Could you let me know why "メリットファクター" is mentioned here for me?

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,376 @@
...
+
+scip_E = int(round(scip_solution.objective))
+print(f"SCIP E(s):    {scip_E}")
+print(f"SCIP feasible: {scip_solution.feasible}")
+print(f"Wall time:    {scip_solve_time:.3f} s")
+
+# %% [markdown]
+# ## 結果の比較
+#
+# SCIPは最適解を一意に返す決定論的なソルバーですが、QAOAはビット列の*分布*を返します。そこでQAOA側は**ベストショット**(全サンプル中で最も低い目的値を達成したビット列)と、参照最適値に対する**ヒット率**(その値を達成したショットの割合)を報告します。
```
ksk-jij: [my opinion]
"報告します。" feels a bit off as Japanese here. "みます" or something like that feels better personally. What do you think?

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,376 @@
...
+print(f"{'SCIP (classical)':<22} {scip_E:>8} {scip_solve_time:>12.3f}")
+print(f"{'QAOA (best shot)':<22} {qaoa_best_E:>8} {qaoa_optimize_time:>12.2f}")
+print()
+print(f"QAOA hit rate on E* = {ref_E}: {hit_rate:.1%}  ({final_shots} shots)")
+
+# %% [markdown]
+# このベンチマークから読み取れることは大きく2点あります。
+#
+# 1. **最適値の到達。** SCIPとQAOAベストショットのどちらも参照最適値$E^\star = 2$に到達しており、QAOAは$n = 5$かつ$p = 3$レイヤーという軽い設定でも最適系列を*見つけられる*ことが分かります。
+# 2. **集中度。** QAOAの価値はサンプリング確率を低エネルギーのビット列に集中させる点にあります。上記のヒット率(およびヒストグラム左端の集中)が、その性質を定量化したものです。
```
ksk-jij: [my opinion]
As same as the first "目標". "最適値の到達。" and "集中度。" feel a bit off as Japanese.

ksk-jij: [question]
With this sentence and only saying 0.8 percent, this notebook says QAOA is quite bad. Is that intentional?

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,376 @@
...
+#
+# 経過時間の列は両者の性質の*違い*を捉えるための値で、勝敗判定として読むものではありません。SCIPはQUBOに対してCPU上で直接動作するのに対し、QAOA側の計測値はステートベクタシミュレータ上での古典・量子最適化ループ全体を含みます。$n$が大きくなったときに、厳密だが指数的に伸びるbranch-and-boundと、ヒューリスティックだが多項式深さで済む回路という*定性的*なトレードオフが立ち現れる、そここそがOMMX Quantum Benchmarksのようなデータセットが両陣営の評価に役立つ理由です。
```
ksk-jij: [my opinion]
I feel we don't need "勝敗判定として読むものではありません". We just can say "経過時間の列は両者の性質の*違い*を捉えるための値です。SCIPは....". What do you think?

ksk-jij: [my opinion]
I believe "branch-and-bound" means SCIP solver, but we didn't explain about SCIP, so I feel either just saying SCIP without branch-and-bound or explaining this is SCIP is better. What do you think?

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,376 @@
...
+# %% [markdown]
+# ## まとめ
+#
+# 本チュートリアルでは次のことを行いました。
+#
+# 1. **OMMX Quantum Benchmarks**データセットからLABSインスタンスをそのまま`ommx.v1.Instance`として読み込みました。
+# 2. `Instance.to_qubo()`でQUBOを取り出して`BinaryModel`にラップし、スピン領域に切り替えた上で、自作のQAOA ansatz(`@qkernel`)を`QiskitTranspiler` + `AerSimulator`を通じて実行しました。
+# 3. QAOAの出力(ベストショット、ヒット率、サンプリング分布)を、*同じ*インスタンスに対する`ommx_pyscipopt_adapter.OMMXPySCIPOptAdapter.solve`の結果、およびベンチマークに同梱された参照最適値と比較しました。
+#
+# 同じパイプラインは他のQOBLIBデータセット(`Marketsplit`、`IndependentSet`、`Network`など)にもそのまま適用できます。対応する`BaseDataset`サブクラスでロードし、`Instance.to_qubo()`でQUBOを取り出して、同じ`BinaryModel` + QAOA ansatz + transpileのループを再利用するだけです。インスタンスサイズがローカルシミュレータの能力を超えた場合は、同じ`executable`をQamomileの他のバックエンド(`QuriPartsTranspiler`、`CudaqTranspiler`など)や実機にそのまま切り替えられます。
```
ksk-jij: [my opinion]
It is first time mentioning "BaseDataset" in this tutorial and I feel this is not necessary since the users can just download instances with `Marketsplit`, `IndependentSet` and so on. What do you think?

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,376 @@
...
+#
+# **次のステップ:**
+#
+# - QAOAそのものの数学については、[QAOA for MaxCut](../algorithm/qaoa_maxcut)と[QAOA for Graph Partitioning](../algorithm/qaoa_graph_partition)を参照してください。
```
ksk-jij: [my opinion]
Since our documentation pages will grow more and more, maybe we don't need to do this linking at the end, but at the appropriate section/cell. For instance, we can put this link at the first time mentioning QAOA in this tutorial instead of here.

ksk-jij (commented):
@jij-yuichiro-nakano Thank youp! I've left some comments. Please have a look :)

docs/en/usage/index.md
jij-yuichiro-nakano: Thank you for the suggestion. I agree — OMMX Quantum Benchmarks is not part of Qamomile itself, so this article fits better under `integration/`. I moved both the EN and JA versions from `usage/` to `integration/`, updated the `tags:` frontmatter from `[usage, ...]` to `[integration, ...]`, removed the entries from `usage/index.md`, added them to `integration/index.md`, and dropped the now-unneeded `OPTIONAL_SKIP_MODULES` entry in `tests/docs/test_tutorials.py` (since `integration/` is excluded from docs CI execution).

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Thank you for the catch. You are right that the trailing 。 reads awkwardly. I changed `**目標。**` to `**目標:**` in the JA `.py` and `.ipynb`.

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Thank you for asking. There is a follow-up article in mind — a larger-scale tutorial that uses PCE — and the `(1)` is meant to anchor that. PCE is still in PR review, so I have not started writing the second article yet. I would like to keep `(1)` for now; if the follow-up does not materialize, we can drop the suffix later.

jij-yuichiro-nakano: Thank you for the feedback. You are right that "OMMXベンチマーク" / "OMMX Benchmarks" alone is too generic. I updated both titles to use the actual dataset name: `OMMX Quantum Benchmarksの活用 (1): ...` / `Using OMMX Quantum Benchmarks (1): ...`. The link text in `integration/index.md` (EN and JA) is updated to match.

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Thank you for the question. The intent of this tutorial is to show how to plug a small hand-written qkernel together with OMMX Quantum Benchmarks. The follow-up article (the `(1)` I mentioned in the title) will use PCE on a larger benchmark instance, and that one will demonstrate the `QAOAConverter` path. For that reason I would like to keep the from-scratch qkernel build here. The BinaryModel-page link suggestion is a good idea — I will fold it into the follow-up article where it fits more naturally.

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Thank you for the question. You are right — the merit factor is mentioned but never reused as a comparison metric in this article. I removed both mentions (the problem-statement and the reference-solution paragraphs) and normalized the narrative around $E$ in EN and JA.

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Thank you for the suggestion. I changed `を報告します。` to `を見てみます。` as you suggested.

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Thank you. The same fix as your earlier `目標。` comment applies here. I changed `**最適値の到達。**` to `**最適値の到達:**` and `**集中度。**` to `**集中度:**` so the three headings are consistent.

jij-yuichiro-nakano: Thank you for the question. The 0.8% number was not great — it came from optimizing on a noisy shot estimate, which made the optimizer struggle. The intent of this section is not to argue that QAOA beats SCIP, but to show how the *same* `ommx.v1.Instance` lets us compare comparable quantities (best shot, hit rate) across quantum and classical solvers. The hit rate itself can be improved by tuning, but tuning is not the point of this article.

While fixing the optimization side, I switched the optimizer to drive the **exact** expectation value of the cost Hamiltonian (via Aer's `EstimatorV2`) instead of a shot estimate. With sampling noise removed from the gradient, BFGS converges properly: the hit rate is now 4.6% (~6× the previous), and the `集中度` paragraph now matches the number it cites.

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Thank you for the suggestion. I agree — the warning was redundant. I removed `勝敗判定として読むものではありません` and connected the sentence directly: `経過時間の列は比較の*形*を示すものです。SCIPはCPU上で…`.

jij-yuichiro-nakano: Thank you for the feedback. I added a one-line intro to SCIP at its first mention: `**SCIP** is a branch-and-bound-based MILP/QUBO solver that can certify an optimum when the solve completes.` The discussion section can then mention branch-and-bound without surprising the reader. In the JA version I also normalized "branch-and-bound" to 「分枝限定法」 so the body uses Japanese consistently.

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Thank you for the suggestion. You are right — readers use `Labs()`, `Marketsplit()`, etc. directly and do not need to know about `BaseDataset`. I changed `BaseDataset サブクラス` / `BaseDataset subclass` to `データセットクラス` / `dataset class` in both EN and JA summary sections.

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Thank you for the feedback. I went one step further and removed the entire `次のステップ:` / `Next steps:` section in both languages. The QAOA-math link (`QAOA for MaxCut: Building the Circuit from Scratch`) is already present at the algorithm intro, so the trailing list was mostly duplicating it. The benchmark-family hint was already covered in the paragraph just above it.

docs/en/integration/ommx_quantum_benchmarks_qaoa.ipynb
```diff
@@ -0,0 +1,972 @@
...
+     "output_type": "stream",
+     "text": [
+      "[2026-05-27 13:24:35]       ├─ \u001b[36mPython: 3.11.15\u001b[0m\n"
+     ]
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "[2026-05-27 13:24:35]       ├─ \u001b[36mVirtual Environment: /Users/yuichironakano/マイドライブ/40_Qamomile/docs-ommx-benchmark/.venv\u001b[0m\n"
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Thank you for catching this. You are right — `ommx-quantum-benchmarks` uses `minto` internally and `minto` prints a host-environment dump (Python version, virtualenv path, CPU model, etc.) at experiment start by default. The previously committed outputs leaked those local-machine details.

In commit 8e1c2837 I set `MINTO_TESTING=true` before instantiating the dataset, which disables `minto`'s verbose logging path, and re-executed both notebooks. The committed outputs now contain only the tutorial's own prints — no `/Users/...` paths, hostnames, or CPU model strings remain. I also took the opportunity to consolidate all imports into a single cell at the top so the `MINTO_TESTING` env-var assignment lives in the natural place next to `import os`.

docs/ja/integration/ommx_quantum_benchmarks_qaoa.ipynb
```diff
@@ -0,0 +1,861 @@
...
+     "output_type": "stream",
+     "text": [
+      "[2026-05-27 13:26:19]       ├─ \u001b[36mPython: 3.11.15\u001b[0m\n"
+     ]
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "[2026-05-27 13:26:19]       ├─ \u001b[36mVirtual Environment: /Users/yuichironakano/マイドライブ/40_Qamomile/docs-ommx-benchmark/.venv\u001b[0m\n"
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Thank you for catching this. You are right — `ommx-quantum-benchmarks` uses `minto` internally and `minto` prints a host-environment dump (Python version, virtualenv path, CPU model, etc.) at experiment start by default. The previously committed outputs leaked those local-machine details.

In commit 8e1c2837 I set `MINTO_TESTING=true` before instantiating the dataset, which disables `minto`'s verbose logging path, and re-executed both notebooks. The committed outputs now contain only the tutorial's own prints — no `/Users/...` paths, hostnames, or CPU model strings remain. I also took the opportunity to consolidate all imports into a single cell at the top so the `MINTO_TESTING` env-var assignment lives in the natural place next to `import os`.

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,427 @@
...
+# %% [markdown]
+# ---
+# tags: [integration, optimization, variational]
+# ---
+#
+# # OMMX Quantum Benchmarksの活用 (1): Qamomileによる量子アルゴリズムの実装とベンチマーク
+#
+# 本チュートリアルでは、公開ベンチマークデータセットの問題にQamomileの量子アルゴリズムを適用し、その解の品質を同じワークフロー上で古典ソルバーと比較する方法を示します。
+#
+# **目標:** QamomileでQAOAソルバーを構築し、[OMMX Quantum Benchmarks](https://github.com/Jij-Inc/OmmxQuantumBenchmarks)データセットからロードした**Low Autocorrelation Binary Sequences (LABS)**インスタンスに対して実行し、[`ommx-pyscipopt-adapter`](https://github.com/Jij-Inc/ommx-pyscipopt-adapter)経由で利用できる古典ソルバーSCIPと結果を比較します。QAOA側とSCIP側はどちらも*同じ*`ommx.v1.Instance`を入力とするため、主な違いはアルゴリズムそのものです。そのため直接比較できます。
```
ksk-jij: [must]
`**` for `**Low Autocorrelation Binary Sequences (LABS)**` appears on ReadTheDocs:

Image

Please make it invisible.

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,427 @@
...
+plt.plot(cost_history, color="#2696EB")
+plt.xlabel("Iteration")
+plt.ylabel("Expectation value (normalized)")
+plt.title("QAOA Optimization Progress (LABS, n=5)")
+plt.show()
+
+# %% [markdown]
+# ### 最終サンプリング
+#
+# 最適化されたパラメータでショット数を増やして再度サンプリングします。その後、元の`ommx.v1.Instance`に対してデコードすることで、返ってくる`ommx.v1.SampleSet`が元のQUBO目的値を直接報告します。このQUBO定式化では、補助変数$z$が積$x_i x_{i+k+1}$を正しく表現しているサンプルではペナルティが発生せず、目的値がLABSエネルギー$E(\boldsymbol{s}) = \sum_k c_k^2$と一致します。一方、$z = x_i x_{i+k+1}$という暗黙の関係を破ったサンプルには、インスタンスに組み込まれたプレースホルダー$P$に比例する加算ペナルティが課されます。
```
ksk-jij: [must]
"報告" in "直接報告します。" is a bit weird. Could you make it a bit more natural in Japanese?

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
ksk-jij: I'm sorry what do you mean by "比較の*形*"? I didn't get the both, literally what "比較の形" means and the reason why "形" is italic.

ksk-jij: Thank you, but apparently that line is not reflected to the Japanese version.

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,427 @@
...
+print()
+print(f"QAOA hit rate on E* = {ref_E}: {hit_rate:.1%}  ({final_shots} shots)")
+
+# %% [markdown]
+# このベンチマークから読み取れることは大きく2点あります。
+#
+# 1. **最適値の到達:** SCIPとQAOAベストショットのどちらも参照最適値$E^\star = 2$に到達しており、QAOAは$n = 5$かつ$p = 3$レイヤーという軽い設定でも最適系列を*見つけられる*ことが分かります。
+# 2. **集中度:** QAOAの価値はサンプリング確率を低エネルギーのビット列に集中させる点にあります。上記のヒット率とヒストグラム左端の集中が、その性質を定量化したものです。
+#
+# 経過時間の列は比較の*形*を示すものです。SCIPはCPU上でQUBOを直接解くのに対し、QAOA側の計測値はステートベクタシミュレータ上での古典・量子最適化ループ全体を含みます。$n$が大きくなると、厳密な分枝限定法と、多項式深さのヒューリスティック回路との間に*定性的*なトレードオフが生じます。OMMX Quantum Benchmarksのようなデータセットは、この両面を評価するために有用です。
```
ksk-jij: [my opinion]
The sentence here is a bit difficult to understand what we want to let users know. It is difficult to articulate why I feel so, but will try below. Please feel free to ask me if I say nonsense or ununderstandable.

[SCIPはCPU上でQUBOを直接解くのに対し、QAOA側の計測値はステートベクタシミュレータ上での古典・量子最適化ループ全体を含みます。]
This sentence feels like we are comparing SCIP and QAOA on a fair stage, but it does not really? Like the sentence can be decomposed into
* SCIPはQUBOを直接解く
* QAOA(にかかる時間)は量子ベクトルのシミュレーションと最適化のループの両方を含む
Those feel like focusing on different thing? Additionally, "SCIPはCPU上でQUBOを直接解くのに対し" followed by the QAOA sentence feels like SCIP does not have any iterations. Branch-and-bound does have iteration process as well.

[$n$が大きくなると、厳密な分枝限定法と、多項式深さのヒューリスティック回路との間に*定性的*なトレードオフが生じます。]
We just mentioned about simulation, but I think this large $n$ cannot be simulated? In that case, the gap between this sentence and the former sentence may cause some mis-understanding.

[OMMX Quantum Benchmarksのようなデータセットは、この両面を評価するために有用です。]
Do you mean deterministic method (such as SCIP) and heuristic method (such as QAOA) by "両面"? Or quantum and classical? I feel either way (or possibly other way) we could improve the description here.

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
ksk-jij: [my opinion]
Maybe as a link to OMMX Quantum Benchmarks, the documentation page https://jij-inc.github.io/OmmxQuantumBenchmarks/ja/ is better? (Obviously English page for English version)

ksk-jij: [must]
The page for ommx-pyscipopt-adapter is 404. Maybe https://github.com/Jij-Inc/ommx/tree/main/python/ommx-pyscipopt-adapter or https://ommx-ja-book.readthedocs.io/ja/latest/user_guide/supported_ommx_adapters.html ?

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,427 @@
...
+import qamomile.circuit as qmc
+import qamomile.observable as qm_o
+from qamomile.optimization.binary_model import BinaryModel, VarType
+from qamomile.qiskit import QiskitTranspiler
+from qamomile.qiskit.transpiler import QiskitExecutor
+
+# %% [markdown]
+# ## OMMX Quantum Benchmarksとは?
+#
+# **OMMX**([Open Mathematical prograMming eXchange](https://jij-inc.github.io/ommx/en/introduction.html))は、数理最適化問題をツール間で受け渡すためのデータフォーマットです。`ommx.v1.Instance`には目的関数、制約条件、決定変数のメタデータ、必要に応じて参照解が格納されます。
```
ksk-jij: [question]
I might be not correct, but does `ommx.v1.Instance` itself could have solution (参照解)?

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,427 @@
...
+#
+# **OMMX Quantum Benchmarks**は、この`ommx.v1.Instance`形式で配布される最適化ベンチマークインスタンス集です。最初に提供されているデータセットは**QOBLIB**(Quantum Optimization Benchmarking Library)[arXiv:2504.03832](https://arxiv.org/abs/2504.03832)で、近年の量子最適化研究で使われている9つの問題ファミリーを収録しています。LABS、Market Split、Independent Set、Steiner Tree Packingなどが含まれます。
+#
+# 各ベンチマークインスタンスは`ommx.v1.Instance`として表現されるため、`QAOAConverter`を含む`ommx.v1.Instance`対応のQamomileワークフローは、追加のアダプタコードなしにこれらの問題を扱えます。同じ`Instance`は`ommx-pyscipopt-adapter`のような古典側のOMMXアダプタにも渡せるので、1つの問題定義を量子・古典の両方のワークフローで使えます。
```
ksk-jij: [my opinion]
I feel we don't really have consented piece of code for "アダプタコード" in Qamomile. Thus, it would be nice to say just "追加の実装なしに" or something similar would be nice?

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
ksk-jij: [my opinion]
What is placeholder $P$ in this case? It might be nice to say just ペナルティが加算されます?

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
```diff
@@ -0,0 +1,427 @@
...
+ref_E = int(reference_solution.objective)
+
+print(f"Shots:                {len(qaoa_summary)}")
+print(f"QAOA best objective:  {qaoa_best_E}")
+print(f"Reference E*:         {ref_E}")
+
+# %% [markdown]
+# ### 目的値の分布
+#
+# QAOAは一つの解ではなくビット列の分布を返します。下のヒストグラムは、最適化済みパラメータで取得した各ショットのQUBO目的値です。赤の点線は参照最適値$E^\star$を示します。その線上、またはすぐ右側にあるサンプルは、$x$の値がLABS和を最小化し、$z$の値が積を正しく符号化しています。線から大きく右に離れたサンプルは、$z$の不整合に対するペナルティを払っています。
```
ksk-jij: [my opinion]
QUBO目的"関数"値?

ksk-jij (changes requested):
@jij-yuichiro-nakano Thank you for addressing my comments. Please have a look at the comments I've left.

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Fixed — removed the `**...**` around the LABS phrase so the literal asterisks no longer render. In MyST/CommonMark the closing `**` here is preceded by `)` and immediately followed by a CJK character, so it is not a valid strong-emphasis delimiter and leaks as a literal `**`. The label `**目標:**` is unaffected (it has a trailing space). The paired notebook was re-synced.

jij-yuichiro-nakano: Fixed — the 404 link now points to the ReadTheDocs supported-adapters page: JA → `ommx-ja-book.readthedocs.io/ja/latest/user_guide/supported_ommx_adapters.html`, EN → the `ommx-en-book` equivalent. Applied to both languages and the paired notebooks.

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Fixed — reworded to 「返ってくる `ommx.v1.SampleSet` から元のQUBO目的関数値を直接読み取れます」, which is more natural than 「直接報告します」.

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Done — switched the link to the documentation page as suggested (JA → `OmmxQuantumBenchmarks/ja/`, EN → `/en/`) instead of the repository.

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Done — Qamomile has no user-written "adapter code" concept, so changed 「追加のアダプタコードなしに」 to 「追加の実装なしに」 (EN: "without writing any extra code").

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Done — $P$ was never defined and added little, so simplified to 「ペナルティが加算されます」 (EN: "incur an additive penalty").

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Done — changed 「目的値」 to 「目的関数値」 here and in the surrounding prose and the section heading for consistency (the sibling `qaoa_graph_partition` article also uses 「目的関数値の分布」). JA only; the EN "objective" wording is unchanged.

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Thank you, this was very helpful. I rewrote the whole paragraph to cover all three points:

- Fairness / different focus: it now states up front that the wall-time is only indicative and that a genuinely fair comparison would need a more careful methodology, rather than implying a level playing field. I also made the SCIP/QAOA sentence symmetric — 「SCIPの計測値はCPU上でのソルバーの実行時間であるのに対し、QAOA側の計測値は…古典・量子最適化ループ全体を含み、いずれも実行環境に依存します」 — so 「直接解く」 no longer implies SCIP has no iterations (branch-and-bound does iterate).
- Large n cannot be simulated: removed the 「$n$が大きくなると…厳密な分枝限定法と多項式深さの…トレードオフ」 sentence entirely, since it clashed with the simulation framing.
- 「両面」 ambiguity: replaced it with explicit wording — 「量子アルゴリズムと古典ソルバーを様々な指標のもとで容易に比較できる」.

Applied to both EN and JA.

docs/ja/integration/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Good catch — you are right. The reference solution is not stored inside `ommx.v1.Instance`; it is provided separately as an `ommx.v1.Solution`. I removed 「必要に応じて参照解が格納されます」 from the Instance description and added 「さらに、参照解も `ommx.v1.Solution` 形式で含まれており、ベンチマークの評価に使用できます」 (with an EN equivalent).

docs/ja/usage/ommx_quantum_benchmarks_qaoa.py
jij-yuichiro-nakano: Resolved, and sorry for the unclear phrasing — I removed 「比較の*形*」 entirely and rewrote the paragraph. It now reads 「なお、経過時間の値はあくまでも参考に留まります…」 with no 「比較の形」 and no italics. EN mirrors this ("the wall-time figures are only indicative").

jij-yuichiro-nakano: The SCIP intro is in the JA source — see the 「古典ベースライン: OMMXアダプタ経由のSCIP」 section: 「**SCIP**は分枝限定法ベースのMILP/QUBOソルバーで、厳密解を求めることができます。」 I have also reworded it for clarity and re-synced the paired notebook, so the latest build includes it. The earlier preview may have been stale.

ksk-jij (approved):
@jij-yuichiro-nakano Thank you for addressing my comments. Looks good to me :)

PR #334, opened 2026-04-13
Add HHL algorithm and Möttönen amplitude encoding

masa-whitestone:
## Summary
- Implement the HHL (Harrow-Hassidim-Lloyd) algorithm for solving linear systems Ax=b on a quantum computer
- Add Möttönen Y-rotation decomposition for real-amplitude state preparation (`amplitude_encoding`)
- Extract shared `get_size` utility from `qft.py` into `frontend/handle/utils.py`

## New Modules
- `qamomile/circuit/algorithm/hhl.py` — HHL circuit builder (QPE → reciprocal rotation → inverse QPE → post-selection)
- `qamomile/circuit/algorithm/mottonen_amplitude_encoding.py` — Möttönen decomposition for encoding real amplitude vectors into quantum states
- `qamomile/circuit/frontend/handle/utils.py` — `get_size()` utility for extracting array sizes from Vector handles

## Public API
```python
from qamomile.circuit.algorithm import hhl, reciprocal_rotation
from qamomile.circuit.algorithm import amplitude_encoding, MottonenAmplitudeEncoding
```

## Tests
- **28 unit tests** — eigenvalue decoding, circuit construction, parameter validation, block structure
- **8 end-to-end tests** — amplitude encoding → HHL → statevector verification (P(θ) and Rz(α) unitaries)
- **107 Qiskit transpile tests** — 5 fixed cases + 100 random seeds (fidelity > 0.999) + post-selection probability
- **CUDA-Q and QURI Parts** backend transpile tests (marker-gated)
- **Möttönen encoding tests** — correctness, edge cases, random vectors

## Test plan
- [ ] `uv run pytest tests/circuit/algorithm/test_hhl.py -v` — unit tests
- [ ] `uv run pytest tests/circuit/algorithm/test_hhl_e2e.py -v` — E2E tests
- [ ] `uv run pytest tests/circuit/algorithm/test_hhl_transpile.py -v` — Qiskit transpile tests
- [ ] `uv run pytest tests/circuit/test_mottonen_amplitude_encoding.py -v` — amplitude encoding tests
- [ ] `uv run pytest` — full regression

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #400, opened 2026-05-22
feat(algorithm): add HHL reciprocal rotation building block (n-qubit clock)

masa-whitestone:
## Summary

Adds the HHL eigenvalue-inversion building blocks specialised to a 2-qubit little-endian clock register, under `qamomile/circuit/algorithm/hhl_rotation.py`:

- **`reciprocal_rotation_2clock_le`** — applies a doubly-controlled `Ry(theta_raw)` to the ancilla for each non-zero clock basis state `|raw>` (`raw` in `{1, 2, 3}`); the `raw = 0` zero-eigenvalue bin is intentionally left untouched. The doubly-controlled `Ry` is decomposed into single-controlled `Ry` + `CX` (Sleator-Weinfurter) so it lowers on every backend rather than relying on a `controlled(..., num_controls=2)` operation.
- **`hhl_middle_block_2clock_le`** — the full middle block `IQFT -> reciprocal rotation -> QFT`, built with the explicit-count `IQFT(2)` / `QFT(2)` composite-gate classes so the qft/iqft stay correct when the block is used as a sub-kernel.

Both qkernels are exported from `qamomile.circuit.algorithm`.

## Tests

`tests/circuit/algorithm/test_hhl_rotation.py` runs cross-backend execution tests (Qiskit, QURI Parts, CUDA-Q — the latter two guarded by `importorskip` and the `quri_parts` / `cudaq` markers). Both the sampling and the expectation-value paths are exercised and checked against the analytic single-qubit rotation references `P(anc=1) = sin^2(theta/2)` and `<Z_anc> = cos(theta)`, parametrised over random seeds, clock bins and boundary angles (0, pi, 2pi).

Verified locally on Qiskit and QURI Parts; CUDA-Q is covered by the test matrix but skipped in the local (non-CUDA) environment.

## Notes

- The kernels are specialised to a fixed 2-qubit clock register (reflected in the `2clock` name), so tests parametrise over clock bins, seeds and boundary angles rather than register size.
- `/local-review` reports no remaining issues for the changed files; `ruff check`, `ruff format` and `zuban` pass.


qamomile/circuit/algorithm/hhl_rotation.py
ksk-jij, line 1: [must]
Could you generalise this to general $n$ qubits? This PR assumes only 2 qubits. What would be expected in this PR is to implement qkernel to perform reciprocal  such as

```python
@qmc.qkernel
def reciprocal_rotation(qubits: qmc.Vector[qmc.Qubit], ancilla: qmc.Qubit) -> tuple[qmc.Vector[qmc.Qubit], qmc.Qubit]:
   ...
```

Like one achieving the transform defined in the attached image below.

Image

ksk-jij, line 1: Also, although it is true that multiple number of control qubits is not supported on some quantum SDKs, but could you write it without decomposition for now? Which means you don't need to decompose multiple controlled gate by yourself.

ksk-jij, line 1: FYI, you can write`_ry = qmc.controlled(qmc.ry, num_controls=1)` or something like that.

ksk-jij, line 1: Also, the file name is a bit off. Could you name it just like reciprocal rotation or something?

masa-whitestone: Resolved in 0ea02caf. The kernel now takes `Vector[Qubit]` and applies the rotation for any concrete `n`. **As we agreed on Slack**, it is implemented as a plain Python helper rather than literally `@qmc.qkernel` because a qkernel body cannot currently express the per-`raw` enumeration — `UInt` lacks the bitwise operators and `arcsin` math the kernel-side form would need. Same pattern as `qamomile.circuit.algorithm.hhl.reciprocal_rotation` from #334. Signature otherwise matches your sketch, with an added `c: float` scaling parameter for the canonical HHL `theta = 2 * arcsin(c / raw)`.

masa-whitestone: Resolved in 0ea02caf. The Sleator-Weinfurter `CCRy` decomposition is removed; the n-controlled `Ry` is now emitted with `qmc.controlled(qmc.ry, num_controls=n)` and applied directly. Tests skip the QURI Parts backend for `n >= 2` with a docstring pointer to this thread.

masa-whitestone: Adopted in 0ea02caf. The `_ry_gate` qkernel wrapper is gone; `qmc.controlled(qmc.ry, num_controls=n)` is used directly.

masa-whitestone: Done in 0ea02caf. File renamed to `reciprocal_rotation.py`; the test file became `test_reciprocal_rotation.py` accordingly.

qamomile/circuit/algorithm/__init__.py
```diff
@@ -27,6 +27,10 @@
     qaoa_state,
     x_mixer,
 )
+from .reciprocal_rotation import (
+    hhl_middle_block,
```
ksk-jij: [my opinion]
We can omit this.

masa-whitestone: Done in 0270a4fe. `hhl_middle_block` removed from `reciprocal_rotation.py` (along with its `iqft` / `qft` imports), dropped from the `__init__.py` export, and the corresponding `TestHhlMiddleBlock` test class is gone. The module is now focused on the canonical `reciprocal_rotation` primitive.

ksk-jij:
I'll close this PR for now.

PR #408, opened 2026-05-25
Fix slice broadcast through inline (sub-kernel call) emit path

ksk-jij:
## Summary

Slice assignment broadcast (e.g. `qs[a:b] = qmc.h(qs[a:b])`) only worked when written directly in the entry kernel. When the same pattern lived in a sub-kernel and the sub was invoked via inline from another kernel, emit failed with `QubitIndexResolutionError: Computed qubit ID '<qs_slice_uuid>_0' not found in qubit_map`. The bug was present from the original slice-broadcast commit (1953c81e) — sub-kernel-via-inline was simply never covered by tests.

Root cause: `ValueSubstitutor.substitute_value` and `_resubstitute_fields` were rewriting `parent_array` / `slice_of` only when the field's own UUID appeared in the substitution map. For an element value inside a callee's broadcast loop the chain is `element.parent_array = sliced_av_clone` (a freshly-cloned UUID not in the map) and `sliced_av_clone.slice_of = qs_param_clone` (a UUID that *is* in the map). A shallow direct-lookup check missed the nested reference, so emit walked a stale `slice_of` pointing at the inner block's input parameter rather than the outer-allocated array.

Fix: substitute `parent_array` and `slice_of` by recursing through `substitute_value` so the slice chain is walked field-by-field and every transitively referenced callee dummy is rewritten consistently. The recursion is bounded by the depth of the call / slice graph, so this works for arbitrarily deep sub-kernel chains and arbitrarily deep nested slices.

## Test plan

- [x] New regression class `TestInlineCallSliceBroadcast` in `tests/circuit/test_slice_pattern_correctness.py` with 11 cases covering: concrete slice, symbolic slice over `n ∈ {3, 5, 7}`, slice-of-slice, multi-slice, slice + scalar mix, two-level inline, four-level inline, nested slice through two-level inline, and `VectorView` passed as a sub-kernel argument with further re-slicing inside.
- [x] Cross-backend execution (Qiskit / QuriParts / CUDA-Q) via the existing `_assert_cross_backend_matches` helper; statevector comparison against an analytic Qiskit reference for every backend that imports successfully.
- [x] Existing slice / inline / transpiler tests remain green: `tests/circuit/test_slice_pattern_correctness.py`, `tests/circuit/test_vector_slicing.py`, `tests/circuit/test_drawer_slice.py`, `tests/transpiler/` all pass (5935 passed, 465 skipped on `tests/circuit/ tests/transpiler/`).
- [x] `ruff check qamomile/ tests/` clean; `ruff format --check qamomile/ tests/` clean.
- [x] `zuban check qamomile/` introduces 0 new errors (the 81 reported errors are pre-existing on `main`, verified by re-running against the base branch).
- [x] `/local-review` clean (P0–P3 all zero).

### DIFF
--- a/qamomile/circuit/transpiler/passes/value_mapping.py
+++ b/qamomile/circuit/transpiler/passes/value_mapping.py
@@ -318,9 +318,9 @@ class ValueSubstitutor:
             fields are left unchanged.
         """
         new_parent_array = v.parent_array
-        if v.parent_array is not None and v.parent_array.uuid in self._value_map:
-            sub_pa = self._chase_transitive(v.parent_array)
-            if isinstance(sub_pa, ArrayValue):
+        if v.parent_array is not None:
+            sub_pa = self.substitute_value(v.parent_array)
+            if isinstance(sub_pa, ArrayValue) and sub_pa is not v.parent_array:
                 new_parent_array = sub_pa
 
         new_element_indices = v.element_indices
@@ -349,9 +349,12 @@ class ValueSubstitutor:
                 new_shape = tuple(new_shape_list)
 
             new_slice_of = v.slice_of
-            if v.slice_of is not None and v.slice_of.uuid in self._value_map:
-                sub_slice_of = self._chase_transitive(v.slice_of)
-                if isinstance(sub_slice_of, ArrayValue):
+            if v.slice_of is not None:
+                sub_slice_of = self.substitute_value(v.slice_of)
+                if (
+                    isinstance(sub_slice_of, ArrayValue)
+                    and sub_slice_of is not v.slice_of
+                ):
                     new_slice_of = sub_slice_of
             new_slice_start = v.slice_start
             if v.slice_start is not None and v.slice_start.uuid in self._value_map:
@@ -461,11 +464,22 @@ class ValueSubstitutor:
             new_shape: tuple[Value, ...] | None = None
             changed = False
 
-            # Check if this is an array element whose parent_array should be substituted
-            if v.parent_array is not None and v.parent_array.uuid in self._value_map:
-                new_parent = self._chase_transitive(v.parent_array)
-                if isinstance(new_parent, ArrayValue):
-                    new_parent_array = new_parent
+            # Substitute the parent_array.  ``v.parent_array`` itself may
+            # not be in the map (it can be a freshly-cloned slice-view
+            # ``ArrayValue`` whose own ``slice_of`` is the callee dummy
+            # being substituted), so a direct ``uuid in self._value_map``
+            # check would leave the parent_array's stale internal fields
+            # in place.  Recursing through ``substitute_value`` walks the
+            # parent_array's fields so the slice chain is rewritten
+            # consistently when a sub-kernel containing
+            # ``qs[a:b] = qmc.h(qs[a:b])`` is inlined.
+            if v.parent_array is not None:
+                sub_parent = self.substitute_value(v.parent_array)
+                if (
+                    isinstance(sub_parent, ArrayValue)
+                    and sub_parent is not v.parent_array
+                ):
+                    new_parent_array = sub_parent
                     changed = True
 
             # Also check element_indices for substitution
@@ -520,9 +534,16 @@ class ValueSubstitutor:
                 new_slice_start_v = v.slice_start
                 new_slice_step_v = v.slice_step
 
-                if v.slice_of is not None and v.slice_of.uuid in self._value_map:
-                    sub_slice_of = self._chase_transitive(v.slice_of)
-                    if isinstance(sub_slice_of, ArrayValue):
+                # ``slice_of`` may also point at a freshly-cloned sliced
+                # ``ArrayValue`` whose own slice metadata still references
+                # the callee dummy.  Recurse through ``substitute_value``
+                # so nested slices (slice-of-slice) get rewritten too.
+                if v.slice_of is not None:
+                    sub_slice_of = self.substitute_value(v.slice_of)
+                    if (
+                        isinstance(sub_slice_of, ArrayValue)
+                        and sub_slice_of is not v.slice_of
+                    ):
                         new_slice_of_v = sub_slice_of
                         slice_meta_changed = True
 
--- a/tests/circuit/test_slice_pattern_correctness.py
+++ b/tests/circuit/test_slice_pattern_correctness.py
@@ -366,6 +366,411 @@ class TestInlineSliceAssignBroadcast:
         assert int(est.gates.total) == 3
 
 
+# Module-scope sub-kernels used by ``TestInlineCallSliceBroadcast``.
+# They live at module scope (rather than as class attributes) for the
+# same name-resolution reason as ``_h_all``: the entry kernels reference
+# them by bare name, which is looked up via the function's
+# ``__globals__``.
+
+
+@qmc.qkernel
+def _slice_broadcast_inner(qs: qmc.Vector[qmc.Qubit]) -> qmc.Vector[qmc.Qubit]:
+    """Apply ``qs[0:2] = qmc.h(qs[0:2])`` inside a sub-kernel.
+
+    Args:
+        qs (qmc.Vector[qmc.Qubit]): Quantum register whose first two
+            qubits are broadcast-H'd in-place.
+
+    Returns:
+        qmc.Vector[qmc.Qubit]: The same register with the slice
+            assignment applied.
+    """
+    qs[0:2] = qmc.h(qs[0:2])
+    return qs
+
+
+@qmc.qkernel
+def _symbolic_slice_inner(qs: qmc.Vector[qmc.Qubit]) -> qmc.Vector[qmc.Qubit]:
+    """Apply ``qs[0:n-1] = qmc.h(qs[0:n-1])`` with symbolic bound.
+
+    Args:
+        qs (qmc.Vector[qmc.Qubit]): Quantum register; the slice covers
+            every element except the last.
+
+    Returns:
+        qmc.Vector[qmc.Qubit]: The same register with the symbolic
+            slice broadcast applied.
+    """
+    n = qs.shape[0]
+    qs[0 : n - 1] = qmc.h(qs[0 : n - 1])
+    return qs
+
+
+@qmc.qkernel
+def _nested_slice_inner(qs: qmc.Vector[qmc.Qubit]) -> qmc.Vector[qmc.Qubit]:
+    """Apply a slice-of-slice broadcast inside a sub-kernel.
+
+    Builds ``s1 = qs[0:4]`` (a view) and then broadcasts H on
+    ``s1[0:2]`` (a view of a view).  The two-level slice chain
+    stresses recursive ``slice_of`` rewriting through ``InlinePass``.
+
+    Args:
+        qs (qmc.Vector[qmc.Qubit]): Quantum register of length >= 4.
+
+    Returns:
+        qmc.Vector[qmc.Qubit]: The register with H applied to
+            ``qs[0]`` and ``qs[1]``.
+    """
+    s1 = qs[0:4]
+    s1[0:2] = qmc.h(s1[0:2])
+    qs[0:4] = s1
+    return qs
+
+
+@qmc.qkernel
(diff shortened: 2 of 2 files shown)


ksk-jij:
# fix-inline-slice-broadcast — work summary

## 1. Problem overview

Qamomile is a quantum programming language that lowers Python functions decorated with `@qmc.qkernel` into an IR (intermediate representation) and transpiles them for backends such as Qiskit, QuriParts, and CUDA-Q. This branch touches `InlinePass`, a transpiler pass whose role is to expand a callee kernel's body into the caller whenever one `@qmc.qkernel`-decorated function (hereafter "kernel") invokes another.

The bug on `main` is that writing a slice-assignment broadcast (e.g. `qs[0:2] = qmc.h(qs[0:2])`) **inside a sub-kernel and then invoking that sub-kernel from an entry kernel via inline** raises `QubitIndexResolutionError` at emit time. The very same syntax written directly inside the entry kernel (e.g. `TestInlineSliceAssignBroadcast` in `tests/circuit/test_slice_pattern_correctness.py`) transpiles fine. In other words, the error is not driven by how the user writes the code but by whether the pattern passes through even one inline call.

A minimal reproducer is the following.

```python
@qmc.qkernel
def inner(qs: qmc.Vector[qmc.Qubit]) -> qmc.Vector[qmc.Qubit]:
    qs[0:2] = qmc.h(qs[0:2])
    return qs

@qmc.qkernel
def entry() -> qmc.Vector[qmc.Bit]:
    qs = qmc.qubit_array(3, "qs")
    qs = inner(qs)
    return qmc.measure(qs)

QiskitTranspiler().transpile(entry)
# QubitIndexResolutionError: Failed to resolve qubit indices for gate 'H'.
#   Operand 0: 'qs[slice][i]' ... Computed qubit ID '<...>_0' not found in qubit_map.
```

The bug lives in the value substitution (`ValueSubstitutor`) performed by `InlinePass`. `InlinePass` clones every IR value (`Value` / `ArrayValue` etc.) appearing inside the sub-kernel's body with a fresh UUID and substitutes the sub-kernel's formal parameters with the caller's actual arguments. The problem is that when the slice-view representation (`ArrayValue.slice_of`) is embedded as a nested field, substitution never reaches that inner reference. Details follow in later sections. This branch fixes the root cause by making the `ValueSubstitutor` substitution logic recursive.

## 2. Frontend (user-facing code) changes

Nothing changes in the `@qmc.qkernel` syntax or API the user writes. Error messages and function signatures are unchanged. The only observable difference is that "code patterns that used to fail on main now transpile correctly on this branch."

An example of user code that fails on main and works on this branch:

```python
@qmc.qkernel
def inner(qs: qmc.Vector[qmc.Qubit]) -> qmc.Vector[qmc.Qubit]:
    qs[0:2] = qmc.h(qs[0:2])   # slice-broadcast assignment
    return qs

@qmc.qkernel
def entry() -> qmc.Vector[qmc.Bit]:
    qs = qmc.qubit_array(3, "qs")
    qs = inner(qs)              # sub-kernel reached via inline
    return qmc.measure(qs)
```

The expected behaviour is that H is applied only to `qs[0]` and `qs[1]` and `qs[2]` stays unchanged. As regression coverage, 9 test methods were added to a new `TestInlineCallSliceBroadcast` class starting at `tests/circuit/test_slice_pattern_correctness.py:630` (one of them is parametrized over `n ∈ {3, 5, 7}`, so the pytest case total comes to 11), and each verifies that the statevector matches an analytic reference on Qiskit, QuriParts, and CUDA-Q. The cases cover the following combinations:

- A sub-kernel containing a concrete slice (`qs[0:2]`) invoked through one level of inline.
- A sub-kernel containing a symbolic slice (`qs[0:n-1]`) invoked through one level of inline, parametrized over `n ∈ {3, 5, 7}`.
- A nested slice that takes a further slice of a slice view (`s1 = qs[0:4]; s1[0:2] = qmc.h(s1[0:2])`).
- Multiple non-overlapping slice broadcasts in the same sub-kernel (`qs[0:2] = ...; qs[2:4] = ...`).
- Slice broadcast and a scalar element gate (`qs[2] = qmc.h(qs[2])`) mixed in the same sub-kernel.
- Two-level and four-level inline chains (`entry → l3 → l2 → l1 → l0`) where only the innermost kernel performs the slice broadcast.
- A nested slice reached through two levels of inline.
- Passing a slice view (`VectorView`) as a sub-kernel argument and re-slicing it inside the sub-kernel to perform a broadcast.

"Inline survives at arbitrary depth" is exercised by the four-level chain test (`test_four_level_inline_transpile_and_statevector_match` at `tests/circuit/test_slice_pattern_correctness.py:754`). The recursion depth is not artificially capped, so the same logic works at any greater depth by design.

## 3. Backend (IR / passes) changes

The implementation diff sits in a single file (`qamomile/circuit/transpiler/passes/value_mapping.py`), with four changed locations. Both changed methods belong to the `ValueSubstitutor` class: `substitute_value` and `_resubstitute_fields`.

A recap of the relevant IR first. In Qamomile's IR a `Value` representing an array element (`q[i]`) holds metadata describing "which array and which index" via `parent_array` / `element_indices` (`qamomile/circuit/ir/value.py`). A slice view (`q[0:2]`) is represented as a fresh `ArrayValue` and uses the `slice_of` field to record "of which array this is a slice." An element of a slice view (`q[0:2][i]`) therefore has `parent_array` pointing to that slice view (an `ArrayValue` with `slice_of` set) — a nested reference graph. Emit-time qubit resolution starts from the element, follows `parent_array`, and then walks the `slice_of` chain down to the root to produce the final physical qubit index.

`InlinePass` (`qamomile/circuit/transpiler/passes/inline.py`) rewrites values in two phases during expansion. Phase one is `UUIDRemapper` cloning every `Value` in the sub-kernel's body with a fresh UUID; the clone rebuilds inner fields such as `slice_of` accordingly (`qamomile/circuit/transpiler/passes/value_mapping.py:137-182`). Phase two is `ValueSubstitutor` building a map "sub-kernel formal parameter (now a cloned UUID) → caller's actual argument" and rewriting each operation against it. The bug was in phase two.

On main, `substitute_value` (`qamomile/circuit/transpiler/passes/value_mapping.py:460`) handled the element value `v`'s `parent_array` substitution via only a direct `v.parent_array.uuid in self._value_map` lookup. When rewriting a slice-view element after inline, the reference graph has three layers: the outermost `v` is the element value, and its own `uuid` is not in the substitution map; `v.parent_array` is the slice view (`ArrayValue`), which holds a freshly-cloned UUID assigned by `InlinePass` and is also not in the substitution map; and finally `v.parent_array.slice_of` points at the sub-kernel's formal parameter `qs`, whose UUID **alone** is registered in the substitution map (`qs_dummy → caller's qs`). The direct lookup missed at both the outermost and middle layers, so the substitution that should have happened at the deepest layer never ran.

By the time emit ran, the element value's `parent_array.slice_of` still pointed at the sub-kernel's formal parameter (a clone UUID that should no longer be referenced from anywhere). `ResourceAllocator` walking the `slice_of` chain to the root then looked that UUID up in `qubit_map`, predictably failed to find it, and halted with `QubitIndexResolutionError` — the final crash site.

The fix changes how `parent_array` and `slice_of` are substituted: instead of a direct UUID lookup, the code now calls `substitute_value` recursively. The recursion descends one field at a time, and whenever a UUID is found in the substitution map at any layer, it is replaced at that point and the result propagates back up through the parent's field. The specific changed locations are:

- `qamomile/circuit/transpiler/passes/value_mapping.py:476-483` — `substitute_value`'s `parent_array` substitution. Instead of `v.parent_array.uuid in self._value_map`, it calls `self.substitute_value(v.parent_array)` and treats the result as substituted iff it is a different instance from the input.
- `qamomile/circuit/transpiler/passes/value_mapping.py:541-548` — `substitute_value`'s `slice_of` substitution. Same recursion-based rewrite. Nested slices (slice-of-slice) are resolved by the same logic as a side effect of the recursion.
- `qamomile/circuit/transpiler/passes/value_mapping.py:321-324` — `_resubstitute_fields`'s `parent_array` substitution. `_resubstitute_fields` is a helper used when a direct lookup of a `Value` returns a different `Value` whose own fields still reference UUIDs that need further substitution. The same recursion is applied here.
- `qamomile/circuit/transpiler/passes/value_mapping.py:351-358` — `_resubstitute_fields`'s `slice_of` substitution. Likewise recursive.

The recursion depth is bounded by "inline-chain depth + slice nesting depth." Both are finite (no cycles) in Qamomile's IR, so infinite recursion is impossible. The recursion terminates as soon as it reaches a non-sliced `ArrayValue` with no `parent_array`.

On performance, each `substitute_value` call now adds one extra function-call frame, but for values without slice metadata the recursive call immediately returns "the same instance" and exits, so the practical overhead is negligible. All tests under `tests/transpiler/` (869 cases) and `tests/circuit/` (5066 cases) pass with no measurable runtime increase.

## 4. Alternatives considered and rationale for the chosen approach

Three approaches were considered.

**A. Recursive value substitution inside `InlinePass` (chosen).**

Change `ValueSubstitutor`'s `parent_array` / `slice_of` substitution from "direct UUID lookup" to "recursive `substitute_value` call." This is the chosen approach.

Its advantages: it patches the exact site where the bug occurs; it stays consistent with `InlinePass`'s responsibility (rewriting formal parameters into actual arguments during sub-kernel expansion); and any UUID registered in the substitution map is guaranteed to be rewritten regardless of how deeply nested it is embedded. The same pattern naturally extends to future field additions beyond slices. The diff size is also small (four sites).

**B. Normalize `VectorView` at frontend IR-emission time so `parent_array` never carries it (rejected).**

The idea is to fold the `slice_of` chain all the way to the root when `VectorView` (the frontend representation of a slice view) is lowered to IR, so that the element value's `parent_array` always points at the original `ArrayValue` directly.

Rejected for two reasons. First, the `slice_of` chain itself encodes "how many layers of logical slice views are stacked," which is the basis for computing physical qubit indices at emit time via the affine map (`parent_index = slice_start + slice_step * view_local_index`). Folding it early only works when `slice_start` / `slice_step` are compile-time constants; symbolic slices (with `UInt` parameters) would still need dynamic expansion. Second, the frontend `VectorView` machinery is deeply intertwined with borrow tracking (which slots a view is currently borrowing), so changing it would force a re-evaluation of every existing slice-validation pass including `SliceBorrowCheckPass`. The bug is fundamentally "missed nested-reference substitution at the IR level," which is not the frontend's responsibility, so a cross-layer overhaul is overkill.

**C. Defensive emit-time fallback so stale `slice_of` is still resolvable (rejected).**

Add logic to `ResourceAllocator._resolve_root_qubit_address` that, when the `qubit_map` lookup misses, falls back to resolving the qubit through a different path (e.g. by `logical_id`) instead of the `slice_of` chain.

Rejected because that hides the symptom rather than fixing the cause. The post-`InlinePass` IR is supposed to satisfy the invariant "no value should still reference a sub-kernel-internal clone UUID," and accepting stale references at emit would blur that invariant. Several other transpiler passes (`analyze`, `classical_lowering`, `plan`, etc.) also implicitly assume that `InlinePass`'s output is internally consistent, so making emit alone defensive would leave the same class of bug latent elsewhere.

## 5. Known limitations

`_resubstitute_fields`'s substitution of `element_indices` / `shape` / `slice_start` / `slice_step` is still a shallow UUID lookup (`qamomile/circuit/transpiler/passes/value_mapping.py:326-336`, `:338-349`, `:359-368`). These fields are all scalar values (`UInt` etc.) in today's IR and do not themselves carry `parent_array` or `slice_of`, so the nested-substitution-miss problem fixed in this branch does not arise. If a future IR extension introduces "use a slice-derived value as an index or shape dimension," these fields will also need the same recursion. When: future IR extensions. Why: the current scalar assumption would break. Future fix: replace the direct UUID lookups with recursive `substitute_value` calls.

Similarly, `_has_referenced_field` (`qamomile/circuit/transpiler/passes/value_mapping.py:273-306`) is also a shallow UUID check. It is a helper that decides whether to call `_resubstitute_fields`, and the only case where "the nested field needs substitution but the shallow check misses it" arises today is via the `parent_array` / `slice_of` path, which this branch handles through the fall-through branch of `substitute_value` rather than through `_has_referenced_field`. The same follow-up applies: an IR extension that introduces the same nested-reference pattern via another path would need this helper updated too.

There are no other known limitations.

PR #409, opened 2026-05-26
Use dedicated symbols for controlled gates

ksk-jij:
## Summary

This PR updates the circuit drawer so controlled wrappers around simple built-in gates can reuse the drawer's dedicated gate symbols instead of always falling back to a generic controlled-U box. Controlled SWAP now renders as a Fredkin-style control dot plus SWAP markers, and controlled X/Z/CX/CZ/Toffoli wrappers also use their dedicated visual forms when the wrapped block is exactly one supported gate.

The IR and backend emission semantics are unchanged. The new behavior is carried as visualization metadata in the analyzer and consumed by the matplotlib renderer.

## Testing

- uv run pytest tests/circuit/test_drawer_controlled.py tests/circuit/test_drawer_slice.py tests/circuit/test_drawer_for_items.py -q

Type checking was intentionally skipped per maintainer instruction during local review.

### DIFF
--- a/qamomile/circuit/visualization/analyzer.py
+++ b/qamomile/circuit/visualization/analyzer.py
@@ -33,6 +33,7 @@ from qamomile.circuit.ir.operation.gate import (
     MeasureVectorOperation,
 )
 from qamomile.circuit.ir.operation.operation import QInitOperation
+from qamomile.circuit.ir.operation.return_operation import ReturnOperation
 from qamomile.circuit.ir.types.primitives import QubitType
 from qamomile.circuit.ir.value import ArrayValue, Value
 
@@ -1000,7 +1001,12 @@ class CircuitAnalyzer:
             u_name = getattr(op.block, "name", "U") or "U"
             power_val = self._resolve_controlled_u_power(op, param_values)
             label = f"{u_name}^{power_val}" if power_val > 1 else u_name
-            box_width = self._estimate_label_box_width(label)
+            controlled_gate_type = self._controlled_u_single_gate_type(op, power_val)
+            box_width = (
+                self.style.gate_width
+                if controlled_gate_type is not None
+                else self._estimate_label_box_width(label)
+            )
             # Control qubits first, then target qubits
             control_indices: list[int] = []
             for qval in op.control_operands:
@@ -1022,6 +1028,7 @@ class CircuitAnalyzer:
                 qubit_indices=control_indices + target_indices,
                 estimated_width=box_width,
                 kind=VGateKind.CONTROLLED_U_BOX,
+                gate_type=controlled_gate_type,
                 box_width=box_width,
                 control_count=len(control_indices),
             )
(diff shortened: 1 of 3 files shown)


ksk-jij:
## Branch Summary

### 1. Problem Overview

On `main`, drawing a normal `qmc.swap` gate uses the dedicated SWAP crossing symbol, but drawing a controlled SWAP through `qmc.controlled(qmc.swap)` or `qmc.controlled(_swap_gate)` falls back to a generic controlled-U box. The underlying wrapped block is still just a single SWAP gate, but that information was not preserved in the visualization path: `ControlledUOperation` was always converted into a generic `VGateKind.CONTROLLED_U_BOX`, with no route back to the renderer's existing dedicated symbols. This PR keeps the execution IR unchanged, but lets the visualization analyzer detect wrapped controlled-U blocks that contain exactly one supported gate plus an optional return, carry that `GateOperationType` into Visual IR, and let the renderer draw the existing CX/CZ/Toffoli/SWAP-style symbols instead of a box.

### 2. Frontend-Level Behavior

The user-facing qkernel code and `qmc.controlled` API do not change. The visible behavior changes only in `draw()` / `MatplotlibDrawer.draw_kernel()`: `qmc.controlled(qmc.swap)` now appears as a Fredkin-style symbol, with a control dot connected to SWAP crossing markers on the two target wires. Controlled wrappers around single X/Z/CX/CZ/Toffoli gates also use the dedicated symbol style instead of a generic box when the wrapped block is simple enough.

```python

@qmc.qkernel
def controlled_swap_example() -> tuple[qmc.Qubit, qmc.Qubit, qmc.Qubit]:
    q = qmc.qubit_array(3, "q")
    cswap = qmc.controlled(qmc.swap)
    return cswap(q[0], q[1], q[2])

fig = qmc.draw(controlled_swap_example)
```

Powered controlled-U operations such as `cswap(q[0], q[1], q[2], power=2)` remain boxed. Collapsing those into only a SWAP-style symbol would hide the exponent, so the generic box is kept when `power > 1`.

### 3. Backend / IR-Level Changes

There is no semantic change to the compiler IR, transpiler pipeline, or backend emission. The change is limited to circuit visualization: Visual IR construction in the analyzer and matplotlib rendering.

In `qamomile/circuit/visualization/analyzer.py:1000`, `ControlledUOperation` conversion now calls `_controlled_u_single_gate_type()` before creating the `VGate`. The helper at `qamomile/circuit/visualization/analyzer.py:1040` ignores `ReturnOperation`, checks whether the wrapped block contains exactly one `GateOperation`, and returns the wrapped `GateOperationType` only for `X`, `Z`, `CX`, `CZ`, `SWAP`, and `TOFFOLI`. That value is stored on `VGate.gate_type`, and dedicated-symbol controlled-U nodes use the normal gate width for layout.

In `qamomile/circuit/visualization/renderer.py:709`, the renderer tries `_draw_vcontrolled_u_special_gate()` before drawing the generic controlled-U target box. The helper at `qamomile/circuit/visualization/renderer.py:751` draws SWAP as explicit control dots plus SWAP X markers, draws X/CX/TOFFOLI as a multi-control target-X form, draws Z/CZ as a dot-only CZ-style form, and falls back to the existing controlled-U box when no dedicated symbol applies.

`tests/circuit/test_drawer_controlled.py:20` adds small qkernels for X/Z/CX/CZ/SWAP/Toffoli wrappers and their controlled forms. `tests/circuit/test_drawer_controlled.py:216` verifies that supported controlled-U nodes carry the expected `gate_type`, `control_count`, `qubit_indices`, and normal gate width. `tests/circuit/test_drawer_controlled.py:228` verifies that powered controlled SWAP stays boxed, and `tests/circuit/test_drawer_controlled.py:236` verifies that controlled SWAP renders without a target box and with one control dot.

### 4. Alternatives Considered

One alternative would be to lower `qmc.controlled(qmc.swap)` into a different IR operation, or to introduce a transpiler pass that rewrites controlled SWAP into a special gate. That would be the wrong layer for this issue: the problem is purely visual, not semantic, and concretizing the IR would go against Qamomile's design preference to keep IR abstract and push backend-specific realization later. This PR therefore keeps `ControlledUOperation` as the source of truth and only attaches a visualization hint in Visual IR for simple cases that can be drawn with existing symbols.

Another alternative would be to make the renderer inspect `ControlledUOperation.block` directly. That would blur the existing analyzer/renderer boundary: the analyzer is responsible for IR semantics and resolution, while the renderer should draw already-resolved Visual IR. This PR keeps that separation by doing the semantic detection in the analyzer and making the renderer branch only on `VGate.gate_type`.

### 5. Known Limitations

When the wrapped block contains multiple operations, or when it wraps a gate without dedicated controlled rendering in this PR, it still renders as a generic controlled-U box. This includes gates such as `H`, `P`, `RX`, `RY`, `RZ`, `CP`, and `RZZ`. The current scope is limited to X/Z/CX/CZ/Toffoli/SWAP-style gates because those either already have a dedicated symbol or naturally compose into one. A future extension could add phase and rotation symbols, but it should first define where parameter labels should go and how they should look with multiple controls.

When `power > 1`, the controlled-U remains boxed even if the wrapped gate is a single SWAP or X-style gate. This preserves the `pow=N` information in the diagram. A future layout enhancement could support exponent labels next to dedicated symbols, at which point powered cases could also avoid the generic box.

PR #410, opened 2026-05-26
Fix deep inline draw routing for bound data

ksk-jij:
## Summary

Fix deep inline circuit drawing so bound array data and bound dictionary data are forwarded through nested inlined qkernel calls. This prevents multi-qubit gates whose operands are resolved from matrices or dict keys from collapsing onto the first qubit row at higher inline depths.

Add regressions for a nested matrix-indexed CX and for qaoa_state with a bound quadratic dict, both drawn with inline=True, fold_loops=False, and deep inline_depth values.

## Tests

- uv run pytest tests/circuit/test_drawer_inline_depth.py tests/circuit/test_drawer_slice.py tests/circuit/algorithm/test_qaoa.py -q
- QAOA draw smoke with inline_depth=5

Type checking was skipped at the user request; full zuban currently fails on main with pre-existing errors.

### DIFF
--- a/qamomile/circuit/visualization/analyzer.py
+++ b/qamomile/circuit/visualization/analyzer.py
@@ -34,7 +34,7 @@ from qamomile.circuit.ir.operation.gate import (
 )
 from qamomile.circuit.ir.operation.operation import QInitOperation
 from qamomile.circuit.ir.types.primitives import QubitType
-from qamomile.circuit.ir.value import ArrayValue, Value
+from qamomile.circuit.ir.value import ArrayValue, DictValue, Value, ValueBase
 
 from .geometry import compute_border_padding
 from .style import CircuitStyle
@@ -635,7 +635,9 @@ class CircuitAnalyzer:
                 elif isinstance(op, ForItemsOperation):
                     dict_value = op.operands[0] if op.operands else None
                     materialized = (
-                        self._materialize_dict_entries(dict_value)
+                        self._materialize_dict_entries(
+                            dict_value, param_values, logical_id_remap
+                        )
                         if dict_value is not None
                         else None
                     )
@@ -1451,7 +1453,7 @@ class CircuitAnalyzer:
 
         # Materialize entries
         materialized = (
-            self._materialize_dict_entries(dict_value)
+            self._materialize_dict_entries(dict_value, param_values, logical_id_remap)
             if dict_value is not None
             else None
         )
@@ -1789,7 +1791,7 @@ class CircuitAnalyzer:
         def add_affected(operand: Value) -> None:
             """Resolve `operand` and merge its wire indices into `affected`."""
             indices = self._resolve_operand_to_affected_qubits(
-                operand, qubit_map, logical_id_remap
+                operand, qubit_map, logical_id_remap, param_values
             )
             if indices is not None:
                 affected.update(indices)
@@ -2597,6 +2599,10 @@ class CircuitAnalyzer:
             start = s_int + st_int * start
             step = st_int * step
             cur = cur.slice_of
+            assert cur is not None, (
+                "[FOR DEVELOPER] slice_of must point at a parent ArrayValue while "
+                "walking a slice chain."
+            )
         return cur, start, step
 
     def _compute_slice_view_wires(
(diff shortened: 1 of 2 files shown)


ksk-jij:
### 1. Problem Overview

On main, when circuit drawing expands nested qkernel calls with `inline=True` and a deep `inline_depth`, operand information for multi-qubit gates resolved from matrices or dictionaries was not correctly forwarded into intermediate inlined blocks. As a result, gates in the rendered drawing could appear collapsed onto the first qubit row instead of spanning their actual qubit rows. The root cause was that `CircuitAnalyzer` forwarded normal scalar parameters and qubit-wire logical IDs while mapping nested block formal inputs to actual inputs, but did not forward `_array_data_*` for bound array data or `_dict_data_*` for bound dictionary data to the child block formal-input logical IDs. This branch updates `_build_block_value_mappings` in `qamomile/circuit/visualization/analyzer.py:3076` so bound array and dictionary data are forwarded into the child visualization context, preserving gate operands derived from expressions such as `q[pairs[0, 0]]` or QAOA quadratic dictionary keys even under deep inline drawing.

### 2. Frontend-Level Behavior

There is no API change for user-written qkernels or `draw()`. The change is only in the rendered output. For example, a qkernel that chooses `CX` control and target qubits from a bound matrix, or a `qaoa_state` circuit where `quad={(1, 2): 1.0}` selects the target pair for an `RZZ`, now renders the multi-qubit gate across the correct qubit rows even when deeply inlined.

```python
fig = qaoa_state.draw(
    p=1,
    quad={(1, 2): 1.0},
    linear={},
    n=3,
    gammas=[1.0],
    betas=[1.0],
    inline=True,
    fold_loops=False,
    expand_composite=True,
    inline_depth=5,
)
```

In this example, QAOA is the Quantum Approximate Optimization Algorithm circuit helper, and the `quad` key `(1, 2)` represents the target qubit pair for a two-body interaction gate. On main, this pair information could be lost during deep inline drawing. In this branch, the regression at `tests/circuit/test_drawer_inline_depth.py:146` verifies that the `RZZ` gate still has `qubit_indices == [1, 2]` under the same conditions.

### 3. Backend / IR-Level Changes

`CircuitAnalyzer` is the visualization-layer analyzer that reads IR and builds qubit wire mappings plus visual IR. This branch does not change backend SDK emission or executable IR semantics; it only fixes nested inline context forwarding inside the visualization layer. `_build_block_value_mappings` at `qamomile/circuit/visualization/analyzer.py:3076` now accepts `Sequence[ValueBase]`, which allows it to handle IR values such as `DictValue` that are not subclasses of `Value`. In addition to the existing logical-ID remap and scalar forwarding, it now forwards bound array data from `_array_data_{actual.logical_id}` or the remapped actual logical ID into `_array_data_{dummy.logical_id}` when `ArrayValue.get_const_array()` is not directly available.

For dictionaries, `qamomile/circuit/visualization/analyzer.py:3254` forwards either `DictValue.metadata.dict_runtime.bound_data` or parent-context `_dict_data_*` into `_dict_data_{dummy.logical_id}`. `_materialize_dict_entries` at `qamomile/circuit/visualization/analyzer.py:3376` now receives `param_values` and `logical_id_remap`, allowing `ForItemsOperation` drawing and wire-chain construction to materialize forwarded bound dictionary data. Its callers were updated in the `build_qubit_map` `ForItemsOperation` path at `qamomile/circuit/visualization/analyzer.py:635` and in `_build_vfor_items` at `qamomile/circuit/visualization/analyzer.py:1455`.

The conservative affected-qubit fallback for loops and branches now passes `param_values` into `_resolve_operand_to_affected_qubits` at `qamomile/circuit/visualization/analyzer.py:1791`, so operands resolvable from bound data or loop parameters are not dropped when precise analysis falls back. For typing consistency, `_resolve_view_chain_to_root` also asserts that `cur.slice_of` is non-None after stepping through a slice chain at `qamomile/circuit/visualization/analyzer.py:2601`.

Regression tests were added in `tests/circuit/test_drawer_inline_depth.py:125` and `tests/circuit/test_drawer_inline_depth.py:146`. The first passes a bound matrix `pairs=[[1, 2]]` through helper qkernels and verifies that a deeply inlined `CX` remains on `[1, 2]`. The second verifies that an `RZZ` generated from the bound `quad` dictionary in `qaoa_state` also remains on `[1, 2]` under deep inline drawing, by inspecting the generated `VisualCircuit` directly.

### 4. Alternatives Not Chosen

One alternative was to expand unresolved multi-qubit gate operands to the whole qubit array in a later drawing fallback. That would avoid losing a gate visually, but it would not recover the exact target qubits; for example, a gate that should be on `[1, 2]` could be displayed as touching `[0, 1, 2]`. The chosen fix instead forwards the missing bound array and dictionary data into the nested inline context, preserving the information needed for exact operand resolution without changing the existing IR or draw API.

Another alternative was to lower bound dictionaries or arrays into per-element operations at the IR level before drawing. Qamomile’s design keeps IR as abstract as possible and delays concretization until the relevant stage. Since this issue was not an IR modeling problem but a visualization-context forwarding gap, keeping the fix inside `CircuitAnalyzer` is narrower and avoids side effects in backend emission or the compiler pipeline.

### 5. Known Limitations

When: if the IR later gains a way to represent an explicitly empty dictionary literal as `DictValue.entries=()` without runtime metadata. Why: `_materialize_dict_entries` currently treats empty IR entries as an unbound dictionary parameter placeholder, because kernel parameters are also constructed with `entries=()`. Future fix: once IR can distinguish an empty literal from an unbound placeholder, the check near `qamomile/circuit/visualization/analyzer.py:3397` should be extended so a known-empty dictionary is handled like runtime metadata and materialized as `[]`.

When: a deeply inlined operand depends on fully symbolic array or dictionary contents that are not represented by draw-time bindings or IR metadata. Why: this branch forwards existing bound data and logical-ID remaps into child contexts; it does not add a solver for unknown symbolic indices or symbolic dictionary contents. Future fix: if more precise drawing of symbolic contents becomes necessary, the natural direction would be to add a representation between the value resolver and visualization analyzer that can distinguish “unknown”, “affects all qubits”, and “affects a specific range”.

PR #411, opened 2026-05-26
Add Grover Adaptive Search (GAS) for Combinatorial Optimization

baptistechev:
## Summary

Adds Qamomile support for Grover Adaptive Search (GAS).

This includes:

- QFT-based primitives for encoding QUBO objectives into quantum phases
- A Grover operator kernel combining oracle, state-preparation inverse, and diffusion
- An adaptive outer loop demo notebook showing incumbent-guided iteration scaling

## Motivation

GAS solves QUBO problems by encoding the objective function into quantum phases via QFT and applying Grover iterations to amplify low-cost states. The adaptive outer loop progressively tightens the threshold, combining quantum amplitude amplification with classical feedback.

## Files

- `qamomile/circuit/algorithm/gas.py` — Qkernel for Grover Operator: QFT-encoding state preparation, hermitian conjugate and diffusion oracle.
- `qamomile/optimization/gas.py` — Build the GAS circuit from a QUBO/HUBO instance.

Tests
- `tests/circuit/algorithm/gas.py` — Tests for the qkernel
- `tests/optimization/gas.py` — Tests for the converter

Docs
- `docs/en/algorithm/grover_adaptive_search.ipynb` — GAS demo notebook for portfolio problem 
- `docs/en/algorithm/grover_adaptive_search.py` — GAS demo notebook for portfolio problem
- `docs/ja/algorithm/grover_adaptive_search.ipynb` — GAS demo notebook for portfolio problem
- `docs/ja/algorithm/grover_adaptive_search.py` — GAS demo notebook for portfolio problem

## Questions for reviewers

1. Should I implement the classical part in the library or only the example in the notebook ? -> we decided only in the notebook for now
2. Do we replace the HUBO qkernel factory by indexing register ? -> we decided the functionality won't be ready yet
3. Should the GASConverter implement the MathematicalProblemConverter interface ? It uses a BinaryModel rather than a spin model and it does not have any Hamiltonian. They have in common a large part of the init and the transpile function.... 

## Implementation notes

**About GASConverter**

- `GASConverter` wraps a QUBO/HUBO model and compiles it into an executable GAS circuit via any registered backend transpiler
- `transpile()` binds all coefficients at compile time and appends measurements via a local `measure_grover_algorithm` kernel
- When the coefficients are real numbers, the `transpile` function call a method `approximate_real_valued_model` which compute the quantize model whose coefficients are rational and approximate the original model (method of appendix B.1. of the CBPO GAS paper)

**About QFT encoding**

- Three primitives encode the QUBO function $f(x) = c_0 + \sum_i l_i x_i + \sum_{i<j} q_{ij} X_i x_j$ into output-register phases:
  - `zero_degree_qft_encoding` — encodes the constant term
  - `first_degree_qft_encoding` — encodes linear coefficients, controlled on input qubit $i$
  - `second_degree_qft_encoding` — encodes quadratic coefficients, controlled on qubits $i, j$
- `apply_function_preparation_qubo` chains these into a full $A_y$: Hadamard superposition → phase encoding → IQFT
- Its adjoint `apply_function_preparation_qubo_dagger` reverses the sequence with negated angles

**About the Grover operator**

- `grover_operator` implements one Grover step: oracle (Z on the last output qubit) → $A_y^\dagger$ → diffusion ($(n-1)$-controlled Z sandwiched by X) → $A_y$
- `grover_algorithm` wraps this in a compile-time loop for a fixed iteration count
- The adaptive outer loop (in the notebook) samples the circuit, checks whether the candidate improves the incumbent $y$, and scales the iteration count by $\lambda$ when no improvement is found

**Notebook tutorial: GAS for portfolio**

The notebook presents a solution to portfolio problem using GAS and demonstrates the benefit of the adaptive outer loop. It follows a similar structure as the existing QAOA graph partitioning and AOA graph coloring tutorials: problem modeling using JIJmodeling -> Converter -> component visualisation -> transpiling to qiskit -> solving the problem. 

## Status

- [x] Implement QFT encoding primitives
- [x] Implement Grover operator and algorithm kernels
- [x] Implement `GASConverter` for QUBO
- [x] Add the HUBO transpiler path
- [x] Documentation / tutorial notebook
- [x] Unit tests
- [x] Support for Continuous variables

## How to test

Example:

```python
from qamomile.optimization.gas import GASConverter

converter = GASConverter(instance)
converter.transpile(transpiler, num_iterations=3)
```

Run tests (once added):

```bash
uv run pytest tests/circuit/algorithm/test_gas.py
uv run pytest tests/optimization/test_gas.py
```


qamomile/optimization/gas.py
```diff
@@ -0,0 +1,991 @@
...
+            list[float]: ``values`` with every non-integer entry rounded to the
+                finest meaningful decimal precision inferred from non-integer
+                entries. Returned unchanged (aside from float coercion) when
+                every value is integer-valued.
+
+        """
+        non_integer_values = []
+        for v in values:
+            fv = float(v)
+            if not np.isclose(fv, round(fv), atol=1e-12):
```
jij-yuichiro-nakano: **Preserve large fractional coefficients**

The default relative tolerance in np.isclose treats 100000.4 as close to 100000, causing the converter to silently store 100000.0 before the intended quantization warning. Please remove this heuristic correction or explicitly use rtol=0.0 wherever only absolute tolerance is intended, including the other comparisons in this helper. Add a regression test that verifies large fractional coefficients are preserved until explicit quantization.

qamomile/optimization/gas.py
```diff
@@ -0,0 +1,991 @@
...
+            no quantization was applied. Thresholds expressed in the original
+            objective scale must be multiplied by this before being compared
+            against the circuit's arithmetic register.
+    """
+
+    _MAX_EXACT_INT_IN_FLOAT_BITS = 54
+
+    def __post_init__(self) -> None:
+        """Derive and cache the BINARY model from the parent spin model."""
+        self.binary_model = self.spin_model.change_vartype(VarType.BINARY)
```
jij-yuichiro-nakano: **Avoid the BINARY–SPIN round trip**

BINARY and OMMX-HUBO inputs have already been converted to SPIN before this conversion back to BINARY. A degree-d BINARY monomial expands into 2^d SPIN terms, making this round trip unnecessarily expensive for high-degree inputs and introducing the numerical noise that the precision correction attempts to remove. Please retain the normalized BINARY model directly and convert only actual SPIN inputs. Add regression coverage for high-degree BINARY models.

qamomile/circuit/algorithm/gas.py
```diff
@@ -0,0 +1,387 @@
...
+
+    Args:
+        q_input (qmc.Vector[qmc.Qubit]): Input register to reflect around the uniform superposition.
+
+    Returns:
+        qmc.Vector[qmc.Qubit]: Updated input register.
+
+    """
+    n = q_input.shape[0]
+    controlled_z = qmc.control(qmc.z, num_controls=n - 1)
```
jij-yuichiro-nakano: **Handle a single-qubit diffuser**

For a valid single-variable QUBO, n - 1 becomes zero and transpilation raises ValueError because num_controls must be positive. Please apply Z directly for n == 1 and retain the controlled-Z path for larger registers. Add transpile-and-execute tests for register sizes including 1, 2, 3, and 5 across Qiskit, QuriParts, and CUDA-Q.

qamomile/optimization/gas.py
```diff
@@ -0,0 +1,991 @@
...
+                "This may affect the accuracy of the results. "
+                "Use approximate_real_coefficients=False to disable this approximation "
+                "and use the original real coefficients."
+            )
+        else:
+            self.effective_model = self.binary_model
+            self.quantization_scale = 1.0
+
+        if output_bits is None:
+            output_bits = self._required_output_bits(self.effective_model)
```
jij-yuichiro-nakano: **Validate the threshold-adjusted output range**

Manual output_bits values bypass range validation, while automatic sizing does not account for the threshold here. Overflow can reverse the sign used by the oracle: for f = 2 - x0 - x1 - x0*x1, three bits produced the unique optimum in all 128 shots, but forcing two bits distributed shots across all states. Please compute bounds for the actual quantized f(x) - y, including y * quantization_scale, use them for automatic sizing, and reject insufficient manual widths. Cover zero, negative, and insufficient widths as well as thresholds outside the model range.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,395 @@
...
+MatplotlibDrawer(block).draw(fold_loops=False)
+
+# %% [markdown]
+# ### Inspecting the Building Blocks
+#
+# The idea of the Grover used for optimization is to, given a reference solution $y$, mark all the elements such that their functional value is lower than $y$, i.e. $f(x) < y$. This is achieved by the preparation oracle $A_y$ and the marker operator $O_y$. Because the preparation operator, prepare all input whose functional value is smaller than $y$ with a negative value, we can identify them by their Most Significant Bit (MSB) that turn to 1 in complementary two representation. Thus the marker operator is simply a $Z$ gate. Finally, the diffusion operator amplifies the amplitude of the marked states.
+#
+# The Grover Ansatz is made of the following component three :
+#
+# - $A_y$ the oracle that prepare the state $\sum \ket{x, y - f(x)}$ : Built from QFT encoding
```
jij-yuichiro-nakano: **Correct the encoded objective sign**

The implementation encodes f(x) - y so that the sign bit marks f(x) < y. The expression y - f(x) reverses that condition and contradicts the surrounding explanation. Please correct this expression in both language versions and regenerate the corresponding notebooks from the source files.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,395 @@
...
+
+    return x, y
+
+# %% [markdown]
+# ## Running the full pipeline
+#
+# We call the hybrid classical-quantum algorithm using Grover Search.
+#
+# The parameter `lamb` handle the speed at which the sample space grow for the number of iterations.
+# The parameters `max_iter` and `max_no_improvement` fix the stopping criterion
```
jij-yuichiro-nakano: **Describe the actual stopping API**

grover_adaptive_search accepts max_no_improvement but has no max_iter parameter. Please remove the reference to max_iter so the explanation matches the callable API. Apply the same correction to the Japanese source and synchronize both notebooks.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,395 @@
...
+        iters
+    )
+    return qmc.measure(q_output), qmc.measure(q_input)
+
+
+block = transpiler.to_block(
+    sampling_grover_algorithm,
+    bindings={
+        "n" : converter.binary_model.num_bits,      # Number of input qubits = number of binary variables
+        "m" : 9, #Number of output qubits, computed by the converter
```
jij-yuichiro-nakano: **Clarify the visualization register width**

The value 9 is hard-coded here rather than obtained from the converter's width calculation. Please describe it as an explicit visualization choice, or obtain the value through the converter if that is the intended example. Update the equivalent Japanese explanation and synchronize the notebooks.

tests/circuit/algorithm/test_gas.py
```diff
@@ -0,0 +1,330 @@
...
+    The expectation of Z on |0> is exactly +1. This analytic reference is trivial
+    (<Z>_{|0>} = 1) and exercises the estimator path independently from the sampler
+    path on every backend.
+    """
+    tr = make_transpiler()
+    H = qm_o.Z(0)
+    exe = tr.transpile(_wrap_qft_encoding_expval, bindings={"coef": 0.0, "H": H})
+    result = exe.run(tr.executor()).result()
+
+    np.testing.assert_allclose(result, 1.0, atol=1e-6)
```
jij-yuichiro-nakano: **Specify numerical tolerances explicitly**

This assertion specifies atol but retains the default relative tolerance. Please specify both atol and rtol according to the intended numerical contract; for this analytic reference, use rtol=0.0. Apply the same review to the numerical assertions in tests/optimization/test_gas.py, including the epsilon comparisons that currently specify neither tolerance.

tests/circuit/algorithm/test_gas.py
```diff
@@ -0,0 +1,330 @@
...
+    """qft_encoding with coef=0 leaves |0> unchanged; Z-observable expval is +1.
+
+    With coef=0, the phase gate applies zero rotation and the state remains |0>.
+    The expectation of Z on |0> is exactly +1. This analytic reference is trivial
+    (<Z>_{|0>} = 1) and exercises the estimator path independently from the sampler
+    path on every backend.
+    """
+    tr = make_transpiler()
+    H = qm_o.Z(0)
+    exe = tr.transpile(_wrap_qft_encoding_expval, bindings={"coef": 0.0, "H": H})
```
jij-yuichiro-nakano: **Exercise meaningful estimator behavior**

This test binds coef=0 and measures Z after applying phase encoding to |0⟩. That expectation remains +1 even for a nonzero coefficient, so the test cannot detect incorrect phase encoding and does not exercise Grover evolution. Please add an expectation-value test for nontrivial function preparation or a Grover state with an analytically known result, and execute it on every supported backend. Also verify both registers in the preparation/dagger tests, since both should be restored deterministically.

tests/circuit/algorithm/test_gas.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Module-level qkernels (reused across tests)
+# ---------------------------------------------------------------------------
+
+
+@qmc.qkernel
+def _wrap_qft_encoding(n: qmc.UInt, coef: qmc.Float) -> qmc.Vector[qmc.Bit]:
+    """Apply qft_encoding on a fresh register and measure it."""
```
jij-yuichiro-nakano: **Document reusable test helpers**

This reusable qkernel wrapper has arguments and a return value but only a one-line docstring. Please add Google-style Args and Returns sections with types, and apply the same convention to the other wrappers, fixtures, nested helpers, and dispatch fakes identified in the local review. Ordinary test functions can retain concise docstrings.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,395 @@
...
+# ---
+# tags: [algorithm, optimization, oracle-based]
+# ---
+#
+# # Grover Adaptive Search for Combinatorial Polynomial Binary Optimization
+#
+# This tutorial demonstrates how to solve the **portfolio problem**
+# using the Grover Adaptive Search (GAS) algorithm with Qamomile.
+#
+# This tutorial explains how to use the Grover Adaptive Search Method for Constrained Polynomial Binary Optimization from the paper
```
jij-yuichiro-nakano: **Remove documentation trailing whitespace**

This line is one of the 14 trailing-whitespace findings reported by git diff --check origin/main...HEAD. Please clean the affected English and Japanese source lines, synchronize notebook cell sources where needed, and rerun the whitespace check.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,395 @@
...
+# tags: [algorithm, optimization, oracle-based]
+# ---
+#
+# # Grover Adaptive Search for Combinatorial Polynomial Binary Optimization
+#
+# This tutorial demonstrates how to solve the **portfolio problem**
+# using the Grover Adaptive Search (GAS) algorithm with Qamomile.
+#
+# This tutorial explains how to use the Grover Adaptive Search Method for Constrained Polynomial Binary Optimization from the paper 
+# > Gilliam, Austin, et al. Quantum, 5, p428 (2021) https://doi.org/10.22331/q-2021-04-08-428.
```
jij-yuichiro-nakano: **Use MyST citation syntax**

Please format this academic reference using MyST citation syntax and the documentation's bibliography mechanism. Apply the same convention to academic citations throughout the tutorial.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,395 @@
...
+#
+# 1. Formulate the problem with [JijModeling](https://jij-inc-jijmodeling-tutorials-en.readthedocs-hosted.com/en/latest/introduction.html).
+# 2. Create an instance with concrete data.
+# 3. Use `GASConverter` to build the Grover Adaptive Search circuit.
+# 4. Run the circuit inside a classical layer that control the number of iterations. Sample the circuit and keep the best solutions until a stopping criterion is reached.
+
+# %%
+# Install the latest Qamomile through pip!
+# # !pip install "qamomile[qiskit,visualization]"
```
jij-yuichiro-nakano, line 40: **Consolidate imports near the beginning**

Please follow the current skeleton.py template and collect the tutorial's library imports near the beginning of the page. Move imports from later cells into that initial import section.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,395 @@
...
+
+# %% [markdown]
+# ## Portfolio Problem Formulation
```
jij-yuichiro-nakano: **Follow the current tutorial section structure**

Please align the tutorial's section organization and heading hierarchy with the current skeleton.py template. Apply the structure consistently across the page.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,395 @@
...
+
+# %%
+from qamomile.circuit.algorithm.gas import apply_function_preparation_qubo
+
+apply_function_preparation_qubo.draw(
+    q_output = 9, 
+    q_input = 9, 
+    y = 0, 
+    linear = converter.binary_model.linear, 
+    quad = converter.binary_model.quad,
```
jij-yuichiro-nakano: **Assert computed results throughout the tutorial**

Please add assertions to verify the computed results in this cell and every other cell that produces calculation results. Check expected values or meaningful invariants so that executing the tutorial detects incorrect results. For numerical or sampled results, use explicit tolerances or robust properties appropriate to the calculation.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,395 @@
...
+    f"GAS returned objective {y}, but the true optimum is {brute_force_y}"
+)
+print("\nGAS matched the brute-force optimum.")
+
+# %% [markdown]
+# The assertion above is what makes this page self-checking: if a regression
+# ever degrades the oracle or the decoding, the notebook fails instead of
+# quietly rendering a worse solution.
+#
+# ## Summary
```
jij-yuichiro-nakano, line 475: **Simplify the summary**

Please simplify this summary to match the current skeleton.py template. Keep it focused on the main takeaways and avoid repeating the detailed explanations already covered in the tutorial.

jij-yuichiro-nakano (changes requested):
@baptistechev 
Thanks for making this pull request!
I recommend requesting changes before merging. The local-review identified correctness issues that can silently change the objective function, reject a valid single-variable problem, or mark incorrect states. The inline comments below describe the requested fixes, regression coverage, and documentation updates.

qamomile/optimization/gas.py
baptistechev: Fixed in d0d73896.

Confirmed the bug: `np.isclose(100000.4, 100000, atol=1e-12)` is `True`, because the default `rtol=1e-5` allows a difference of ~1.0 at that magnitude, so the value was snapped to an integer before `transpile()` could see it as fractional.

I kept the heuristic and made every comparison absolute-only instead of removing it, since it still serves its stated purpose for genuinely noisy decimals. `rtol=0.0` is now explicit at all four `np.isclose` calls in `_align_precision` and in `approximate_real_valued_model`'s zero check. I also fixed the same latent bug outside the helper: the `has_non_integer` gate in `transpile()` had the identical comparison, so a large fractional coefficient would have skipped quantization entirely.

Regression tests added in `tests/optimization/test_gas.py`: `test_align_precision_preserves_large_fractional_values` (the helper returns `100000.4` untouched) and `test_large_fractional_coefficient_survives_construction` (the coefficient survives `__post_init__` and still triggers the quantization warning, with `quantization_scale > 1.0`). Every numeric assertion in that file now passes an explicit `rtol`.

qamomile/optimization/gas.py
baptistechev: Fixed in d0d73896.

The round trip could only be removed in the shared base: `normalize_problem_input` was already building the BINARY model (`BinaryModel.from_hubo(hubo, constant)`) and discarding it in the same expression that derived the SPIN model. It now returns both, the base converter stores it as `self.binary_model`, and `GASConverter.__post_init__` precision-aligns that instead of calling `spin_model.change_vartype(BINARY)`. A `BinaryModel` input pays only for a copy on the BINARY side; only a genuine SPIN input is actually converted, and in one direction.

The other unpack site (`PCEConverter`, which deliberately does not inherit from the base) was updated too. Full unit suite is green, so QAOA/QRAO/FQAOA/PCE are unaffected.

Regression coverage: `test_high_degree_binary_model_is_not_round_tripped` in `tests/optimization/test_gas.py` is parametrized over degrees 3-6 and asserts the coefficients arrive bit-for-bit identical; `test_spin_input_is_converted_to_binary_once` checks the SPIN input path still produces matching energies. At the converter level, `test_hubo_ommx_instance_exposes_matching_binary_model` in `tests/optimization/test_converter_ommx_sampleset.py` asserts the BINARY and SPIN models agree on every assignment for an OMMX HUBO instance.

One thing worth recording: decoding still routes through `spin_model.decode_from_sampleresult`, which length-checks against `spin_model.num_bits`, while the circuit width comes from the BINARY model. The tests assert `binary_model.num_bits == spin_model.num_bits` so a divergence would be caught rather than silently mis-decoded.

qamomile/circuit/algorithm/gas.py
baptistechev: Fixed in d0d73896.

Confirmed: `qamomile/circuit/frontend/operation/control.py:337` raises `ValueError: num_controls must be >= 1`, so a single-variable QUBO could not be transpiled at all. A one-qubit register now applies a bare Z, which is the same operator the general path produces: X Z X = -Z, and the multi-qubit identity carries that same global sign for every n (for n = 2, X X CZ X X = diag(-1, 1, 1, 1)).

One implementation note, because the obvious fix does not work. Writing the branch inside the qkernel body fails: the DSL transformer rewrites the `if` into an `IfOperation` and `emit_if` traces *both* arms, so the multi-controlled arm still builds `qmc.control(qmc.z, num_controls=0)` and raises during constant folding, before the dead branch can be pruned. The branch therefore lives in a plain helper that the transformer does not descend into, resolving the width from `Vector.shape[0]`'s constant value the same way `algorithm/trotter.py` does.

The HUBO path had a character-identical copy of the diffuser; rather than fix it twice I deleted it and routed `hubo_grover_operator` through the shared `diffusion_op` (6d708736).

Tests: `test_diffusion_op_transpiles_for_every_register_width` covers n = 1, 2, 3, 5 on Qiskit, QuriParts and CUDA-Q, and does not just smoke-test - it pins the distribution against the analytic amplitudes of H^n D H^n on the all-zero state (`1 - 2/N` on the zero string, `-2/N` elsewhere). `test_single_variable_qubo_transpiles_and_samples` covers the end-to-end converter path.

qamomile/optimization/gas.py
baptistechev: Fixed in d0d73896.

Sizing is now computed from the actual encoded quantity. `_required_output_bits(model, y_scaled)` takes the threshold, and `transpile()` passes `y * quantization_scale`, so the width is the smallest m with `2**(m-1) > max(|f_min - y|, |f_max - y|)` rather than one derived from the bare span. A manual `output_bits` that cannot represent that range now raises `ValueError` naming both the encoded range and the required width; the same check rejects zero and negative values, since the requirement is never below 2. `_objective_bounds` was factored out so the sizing and the error message share one definition.

I also added the public `GASConverter.required_output_bits(y)` so callers can ask for the width without reaching into a private method - the tutorials now use it (see the thread on the hard-coded 9).

Worth flagging, because it changed two existing tests. Quantization rescales coefficients by `2**(q-1)`, so real-valued models genuinely need very wide registers: `test_hubo_degree4_term_transpiles_and_samples` was passing `output_bits=5` for an encoded range of roughly [-4e13, 7e13], i.e. it needed 48 qubits and had been silently wrapping around the whole time. That test and `test_hubo_random_coefficients_sample_valid_bitstrings` now pass `approximate_real_coefficients=False`, which keeps the register simulatable and is appropriate since both assert structural validity only. The strictness is doing real work here rather than merely being pedantic.

Negative coverage added: `test_transpile_rejects_non_positive_output_bits` (0, -1, 1), `test_transpile_rejects_insufficient_output_bits` (exactly one bit short, plus the sufficient width accepted), `test_transpile_widens_the_register_for_out_of_range_thresholds` (thresholds above f_max and below f_min, asserting the two's-complement interval actually contains the range), and `test_required_output_bits_accounts_for_the_threshold` for an objective whose range never approaches zero.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in d0d73896. Both language sources now read `\sum_x \ket{x, f(x) - y}`, which matches the implementation and the surrounding MSB explanation. Both notebooks were regenerated from the sources and re-executed with outputs.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in d0d73896. `grover_adaptive_search` only takes `max_no_improvement`, so the sentence now mentions just that parameter and `lamb`. Applied to both language sources with the notebooks regenerated.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in d0d73896. The value now genuinely comes from the converter: I added the public `GASConverter.required_output_bits(y)` (the same sizing `transpile()` uses, see the output-range thread) and both tutorials call `converter.required_output_bits(y=0)`, print it, and reuse it for the drawing calls. For this instance it reports 9, and the accompanying prose now explains why the width is not a free parameter - it has to hold `f(x) - y` or the two's-complement encoding wraps and flips the oracle's sign bit.

tests/circuit/algorithm/test_gas.py
baptistechev: Fixed in d0d73896. The analytic assertion now uses `atol=1e-6, rtol=0.0`, and I swept the numerical assertions in `tests/optimization/test_gas.py` as well - the `_detect_eps` comparisons that specified neither tolerance now pass `atol=1e-12, rtol=0.0`, as do the `approximate_real_valued_model` and `quantization_scale` assertions. Every `np.isclose` / `assert_allclose` in both GAS test files now states both tolerances explicitly.

tests/circuit/algorithm/test_gas.py
baptistechev: Fixed in d0d73896.

You were right that the old test could not fail: `<Z>` on |0> is +1 for any coefficient, since a phase gate does nothing to |0>. The wrapper now applies H first, so the register is |+> and the encoded phase is observable: on one qubit `qft_encoding` applies P(pi * coef), giving `<X> = cos(pi * coef)`. It is parametrized over coef = 0, 1/2, 1 and 1/3 with expected values 1, 0, -1 and 1/2, checked on all three backends, so a wrong angle or a dropped phase gate now shows up.

Both registers are verified in the preparation/dagger tests as well. `test_apply_then_dagger_restores_both_registers` (renamed, circuit layer) and `test_hubo_prep_dagger_restores_state` (optimization layer) now measure the output and input registers and assert a single deterministic all-zero outcome across all shots, rather than only checking the input register.

On a full Grover state with an analytically known result: the closest thing I added is `test_diffusion_op_transpiles_for_every_register_width`, which pins the exact post-diffusion distribution rather than just sampling it. The existing `test_qubo_grover_finds_unique_optimal_state` already asserts P = 1 for the unique optimum after one iteration. Both are sampler-path; happy to add an estimator-path Grover case if you would like that specifically.

tests/circuit/algorithm/test_gas.py
baptistechev: Fixed in d0d73896. All six reusable qkernel wrappers in `tests/circuit/algorithm/test_gas.py` now carry Google-style `Args:` and `Returns:` sections with types, as do the two new ones. In `tests/optimization/test_gas.py` the same applies to the nested `wrap_const` / `wrap_term` / `wrap_prep_then_dagger` kernels, the `_all_integer` helper, and the `_fake_qubo` / `_fake_hubo` dispatch fakes. I also lifted the `_CaptureTranspiler` stand-in, which had been defined twice inside test bodies, to module level. Ordinary test functions keep their concise one-line docstrings as you suggested.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in d0d73896. `git diff --check origin/main...HEAD` is now silent - all 14 findings are gone from both language sources, and the notebooks were regenerated from the cleaned sources rather than patched.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in d0d73896. The reference is now `{cite:p}` with the DOI key `10.22331/q-2021-04-08-428`, matching the convention already used in `docs/en/algorithm/pce_maxcut.py` and `qsci.py`, and it appears in the page intro rather than as a hand-written blockquote. It is the only academic citation on the page.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in d0d73896. Both tutorials now have a single import cell right after the pip cell, following `docs/skeletons/algorithm.py`. The imports that were previously scattered across the converter cell, the visualization cells, the classical-loop cell and the brute-force cell (jijmodeling, numpy, itertools, random, typing, the Qamomile circuit and GAS symbols, the drawer, the transpiler and AerSimulator) are all collected there. An unused `import ommx` was dropped in the process.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in d0d73896. Both pages now follow `docs/skeletons/algorithm.py`: intro, pip cell, imports, then Background, Problem Settings, Algorithm, Implementation and Result, with the sub-sections (circuit visualization, building blocks, classical layer) nested under Implementation as H3s. The old flat sequence of H2s and the trailing 'Verifying the Result' section are gone - verification now lives under Result where the skeleton puts it.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in d0d73896. Every cell that computes something now asserts. Concretely: the data cell checks the shapes of mu and Sigma and that the covariance matrix is symmetric; the instance cell checks the decision-variable count and that the empty portfolio evaluates to exactly 0; the converter cell checks `binary_model.num_bits`; the sizing cell checks the reported width; the visualization cell checks the inlined block is non-empty; the GAS cell checks the returned assignment is binary, of the right length, and that the reported objective actually matches re-evaluating that assignment; and the final cell keeps the brute-force comparison. Numerical comparisons use `np.isclose` with explicit `atol` and `rtol=0.0`. Both notebooks re-execute cleanly, and `pytest -m docs -k grover_adaptive_search` passes for both languages.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in d0d73896. The summary is now three bullets in the skeleton's 'In this notebook, we:' form, covering the hybrid loop, what `GASConverter` provides, and the register sizing plus brute-force check. The phase-encoding, higher-order and threshold-scaling bullets were dropped - they repeat the Algorithm and Implementation sections - and the separate paragraph about the page being self-checking went with them, since the assertions now speak for themselves.

tests/circuit/algorithm/test_gas.py
jij-yuichiro-nakano: Thanks for improving the phase-encoding expectation test and checking that both registers are restored by preparation followed by its dagger.

The remaining part of this request is an estimator-path test for function preparation or a Grover state. The existing sampler tests provide useful coverage, but the new expectation test still exercises only standalone phase encoding.

Please add a test using the existing QUBO `f(x) = 2 - x0 - x1 - x0*x1` with threshold zero. With a three-qubit output register, the expected Z value of its sign bit is `0.5` before Grover iterations and `-1.0` after one iteration. Please verify these values through `qmc.expval` on Qiskit, QuriParts, and CUDA-Q, with explicit tolerances.

This would complete the remaining estimator coverage requested in the original review.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,438 @@
...
+# ## Background
+#
+# GAS is a hybrid loop, not a single circuit. The quantum half answers one
+# fixed question at a time, and a classical layer drives the threshold down:
+#
+# 1. Formulate the problem with [JijModeling](https://jij-inc-jijmodeling-tutorials-en.readthedocs-hosted.com/en/latest/introduction.html).
+# 2. Create an instance with concrete data.
+# 3. Use `GASConverter` to build the Grover circuit for the current threshold.
+# 4. Sample it, keep the best candidate, and repeat until a stopping criterion
+#    is reached.
```
jij-yuichiro-nakano: **Move the tutorial overview into the introduction**

This passage describes the tutorial workflow—modeling, constructing an instance, compiling, and sampling—rather than providing background on the algorithm. Please integrate it into the opening abstract/introduction, keeping the overview concise and avoiding duplication. Reserve Background for the concepts readers need to understand GAS.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,438 @@
...
+    apply_function_preparation_qubo,
+    diffusion_op,
+    grover_algorithm,
+)
+from qamomile.circuit.visualization import MatplotlibDrawer
+from qamomile.optimization.gas import GASConverter
+from qamomile.qiskit import QiskitTranspiler
+
+# %% [markdown]
+# ## Background
```
jij-yuichiro-nakano, line 62: **Explain the motivation for Grover Adaptive Search**

Please add a brief introduction to Grover search in the Background section: an oracle marks target states, and amplitude amplification increases their measurement probability. Explain that a fixed search predicate alone does not define an objective-minimization procedure. GAS extends this into optimization by marking states with `f(x) < y`, finding an improving candidate, and updating the threshold to that candidate's objective value.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,438 @@
...
+# The remaining question is how many times to apply the Grover operator. GAS
+# answers it by sampling the iteration count from a range that grows slowly
+# whenever a round brings no improvement.
+
+# %% [markdown]
+# ## Implementation
+#
+# `GASConverter` takes an OMMX instance and converts the problem into a
+# QUBO/HUBO in the BINARY domain. Its `transpile()` method builds, for a given
+# transpiler, the Grover circuit used by the adaptive search.
```
jij-yuichiro-nakano: **Attribute the state-preparation implementation to the paper**

Please explicitly state here that the implementation of `A_y` follows the QFT-based construction in Gilliam, Woerner, and Gonciulea, *Grover Adaptive Search for Constrained Polynomial Binary Optimization*. Use the existing MyST citation key `10.22331/q-2021-04-08-428`. The introduction cites the paper generally, but does not identify it as the source of this implementation.

docs/en/algorithm/grover_adaptive_search.py
```diff
@@ -0,0 +1,438 @@
...
+#   from QFT phase encoding. Because the register holds $f(x) - y$ in two's
+#   complement, the marked inputs are exactly those whose Most Significant Bit
+#   (MSB) is $1$.
+# - $O_y$, the marker: a single $Z$ on that MSB.
+# - $D$, the diffusion operator, which amplifies the amplitude of the marked
+#   states. It is a single multi-controlled-$Z$ sandwiched between $X$ layers.
+#
+# The remaining question is how many times to apply the Grover operator. GAS
+# answers it by sampling the iteration count from a range that grows slowly
+# whenever a round brings no improvement.
```
jij-yuichiro-nakano: **Connect the operators to the complete search procedure**

Please expand this explanation to show how the components work together. `A_y` associates each candidate with `f(x) - y`; `O_y` reverses the phase of candidates with negative encoded values. Explain how the sequence `O_y`, `A_y†`, `D`, and `A_y` amplifies improving candidates, followed by measurement and a classical threshold update. Clarify that `y` is an objective-value threshold, not a solution, and that phase marking alone does not change measurement probabilities.

jij-yuichiro-nakano (changes requested):
Thanks for the updates. The tutorial still needs a clearer introduction to Grover search and a more complete explanation of how GAS performs optimization. Please address the four comments below.

tests/circuit/algorithm/test_gas.py
baptistechev: Added in 113fc5bd, with exactly the reference you specified.

`test_grover_sign_bit_expval_matches_analytic_value` in `tests/circuit/algorithm/test_gas.py` uses f(x) = 2 - x0 - x1 - x0*x1 with threshold 0 and a three-qubit output register, and estimates Z on the sign bit `q_output[2]` through `qmc.expval`. It asserts 0.5 after state preparation and -1.0 after one Grover iteration, on Qiskit, QuriParts and CUDA-Q, with `atol=1e-6, rtol=0.0`. Both values reproduce: the preparation leaves f = 2, 1, 1, -1 in uniform superposition so exactly one of four inputs sets the sign bit, and one iteration on one marked state out of four reaches sin^2(3*arcsin(1/2)) = 1 exactly.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in 113fc5bd. The four-step list is gone from Background and is now a single sentence in the opening abstract: model with JijModeling, build the Grover circuit for the current threshold, and sample inside a classical loop that keeps the best candidate until it stops improving. The JijModeling link moved with it, so nothing is duplicated and the section is a net reduction in length. Applied to both language sources.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in 113fc5bd. Background is now conceptual only, in two short paragraphs: Grover search takes an oracle that marks target states and uses amplitude amplification to raise their measurement probability, which answers a decision problem, since a fixed predicate says which states qualify rather than which is best; GAS makes the predicate movable, marking f(x) < y, sampling an improving candidate, lowering y to that candidate's objective value, and repeating. Applied to both language sources.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in 113fc5bd. The `A_y` bullet now reads "built from QFT phase encoding following {cite:p}`10.22331/q-2021-04-08-428`", so the construction is attributed at the point where it is described rather than only in the introduction. Same key as the intro citation, applied to both language sources.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Fixed in 113fc5bd. The Algorithm section now opens by stating that y is a threshold on the objective value, not a candidate solution, and a short paragraph after the operator bullets traces the procedure: one iteration applies O_y, then A_y-dagger, D, A_y; the phase flip on its own leaves every measurement probability unchanged, and the reflection A_y D A_y-dagger is what converts it into amplitude; measuring the input register returns an improving candidate, and the classical layer lowers y to its objective value. I kept it to one paragraph and did not restate the per-operator roles already given in the bullets. Applied to both language sources.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Done, as of 517dee98. (A reply I posted here earlier was deleted, so this restates the resolution against the current version of the page.)

The four-step overview is no longer in Background. It now sits in the introduction, immediately after the abstract:

> This page solves an unconstrained **portfolio selection** problem with Qamomile's `GASConverter`.
>
> The tutorial is built on the following structure:
>
> 1. Formulate the problem with JijModeling.
> 2. Create an instance with concrete data.
> 3. Use `GASConverter` to build the Grover circuit for the current threshold.
> 4. Sample it, keep the best candidate, and repeat until a stopping criterion is reached. GAS is a hybrid loop, not a single circuit.

Background is now reserved for the concepts needed to understand GAS, and carries nothing about the tutorial's own steps, so there is no duplication between the two. Applied to both language sources, with the notebooks regenerated from them.

docs/en/algorithm/grover_adaptive_search.py
baptistechev: Done, as of 517dee98. (A reply I posted here earlier was deleted, so this restates the resolution against the current version of the page.)

The Algorithm section now covers each point you listed.

It opens by separating the threshold from a solution: "Here $y$ is a threshold on the objective value, not a candidate solution: GAS marks every input with $f(x) < y$."

The first two bullets state what the operators do rather than only how they are built:

> - $A_y$, the preparation operator. It associates each input with $f(x) - y$ by building the quantum dictionary state $\sum_x \ket{x, f(x) - y}$. That state is built from QFT phase encoding following {cite:p}`10.22331/q-2021-04-08-428`. Because the register holds $f(x) - y$ in two's complement, the marked inputs are exactly those whose Most Significant Bit (MSB) is $1$.
> - $O_y$, the marker operator. It reverses the phase of candidates with negative encoded values, using a single $Z$ on that MSB.

And a closing paragraph ties the sequence to the full procedure, including the point that marking alone is not enough:

> One iteration applies $O_y$, then $A_y^\dagger$, $D$, $A_y$. The phase flip on its own leaves every measurement probability unchanged; the reflection $A_y D A_y^\dagger$ is what converts it into amplitude. Measuring the input register then returns an improving candidate, and the classical layer lowers $y$ to its objective value.

Applied to both language sources, with the notebooks regenerated from them.

jij-yuichiro-nakano (approved):

PR #412, opened 2026-05-26
Make local-review portable and expose skills under .codex/

ksk-jij:
## Summary

- Make the `local-review` skill body portable by dropping `$ARGUMENTS` / `argument-hint` / `TARGET` plumbing and hardcoding `main` as the diff base. The skill is invoked against `main` in practice anyway, and removing the placeholder lets other AI tools (which do not expand `$ARGUMENTS`) read the same SKILL.md unchanged.
- Add `.codex/skills` as a symlink to `.claude/skills` so Codex CLI auto-discovers the same skill set. Codex uses the same SKILL.md format and progressive-disclosure semantics as Claude Code, so a single source of truth under `.claude/skills/` now serves both tools.

## Test plan

- [ ] `/local-review` still runs end-to-end and reports against `main` (manual smoke test in Claude Code)
- [ ] `.codex/skills/local-review/SKILL.md` resolves to the same file as `.claude/skills/local-review/SKILL.md` (`ls -la .codex/skills/local-review/SKILL.md`)
- [ ] Codex CLI lists the four skills (`local-review`, `translate`, `python-simplifier`, `generate_release_note`) via its skill discovery

### DIFF
--- a/.agents/skills
+++ b/.agents/skills
@@ -1,0 +1,1 @@
+../.claude/skills
--- a/.agents/skills
+++ b/.agents/skills
@@ -1,1 +1,0 @@
-../.claude/skills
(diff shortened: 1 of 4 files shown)


ksk-jij:
I've checked if codex recognises those skills on my local. It worked.

PR #420, opened 2026-05-29
docs(tutorial): assert deterministic values in "Vector Slicing"

masa-whitestone:
## Summary

Third of the per-tutorial assert series (after #418, #419). Tutorial 03 has a different shape from the prior two: most code cells only `.draw()` (no measured output, no resource estimate printed), so the only deterministic values displayed are the **error-type prints** in the four `VectorView` anti-pattern examples.

Each block already uses a typed `except XxxError as e:` and an `else: raise AssertionError(...)` guard, so the error path is positively verified. This PR adds an explicit `assert type(e).__name__ == "<ExpectedType>"` after the existing `print` statements so the test pins the **exact concrete type** rather than also accepting future subclasses — same pattern as #418's `bad_rebind`.

## Asserts added

| Error block | Asserted type |
|---|---|
| `direct_parent_access` | `QubitBorrowConflictError` |
| `forgot_return` | `UnreturnedBorrowError` |
| `overlapping_views` | `QubitBorrowConflictError` |
| `invalid_nested_return` | `AffineTypeError` |

No other deterministic values to guard in this tutorial; all the non-error cells just call `.draw()`.

## Test plan

- [x] `uv run pytest tests/docs/test_tutorials.py -m docs -k "03_vector_slicing" -v` passes for both `en/` and `ja/`.
- [x] `.ipynb` re-synced via `jupytext --to ipynb --update`.
- [x] EN and JA parity-locked.


ksk-jij:
Thank you, but what this PR adds has already covered in else, so will close this PR.

PR #429, opened 2026-05-29
docs(algorithm): assert deterministic invariants in "Möttönen Amplitude Encoding"

masa-whitestone:
## Summary

Continues the systematic effort to make `docs/{en,ja}/{tutorial,algorithm}/`
articles double as a regression check by asserting every deterministic
printed value. The Möttönen amplitude-encoding tutorial was already mostly
covered (fidelities, gate counts, runtime-parameter counts, sampling
deviations, expval); this PR closes the two remaining gaps:

- **§1 target distribution.** `target (norm)` for `a = (1, 2, 3, 4)` is
  pure numpy normalization and equals `a / sqrt(30)` exactly — now asserted
  directly. The phase-invariant per-basis probability `|sv|^2` is also
  asserted against `[1, 4, 9, 16] / 30` as defense-in-depth alongside the
  existing fidelity assert.
- **§4.1 directing error message.** The tutorial explicitly claims that
  the `ValueError` raised when `amps` is left as a runtime parameter
  points the reader at `amplitude_encoding_from_angles`. That claim is
  now asserted: the message must contain that identifier verbatim.

EN and JA receive parallel changes. `.ipynb` files are re-synced via
`jupytext --to ipynb --update`.

## Test plan

- [x] `uv run pytest tests/docs/test_tutorials.py -m docs -k "mottonen_amplitude_encoding" -v` passes locally for both EN and JA.

### DIFF
--- a/docs/en/algorithm/mottonen_amplitude_encoding.ipynb
+++ b/docs/en/algorithm/mottonen_amplitude_encoding.ipynb
@@ -179,9 +179,14 @@
     "print(f\"prepared      = {np.round(sv, 4)}\")\n",
     "print(f\"target (norm) = {np.round(expected, 4)}\")\n",
     "print(f\"fidelity      = {fidelity(sv, expected):.6f}\")\n",
+    "# Target is purely deterministic from amps_real / ||amps_real|| = a / sqrt(30).\n",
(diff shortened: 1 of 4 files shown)


ksk-jij:
What this PR adds has already mostly covered in the original doc, so I'll close this PR.

PR #439, opened 2026-05-29
fix: QuriParts controlled gate parameter context

### LINKED ISSUE #416: QuriParts emitter: runtime parameters after a controlled gate fail with "parameter does not belong to this circuit"
## Summary

On the QuriParts backend, any `@qkernel` that uses a controlled gate (`qmc.control(...)`) **before** a gate driven by a runtime parameter fails to transpile with:

```
ValueError: The given parameter Parameter(name=<p>) does not belong to this LinearMappedParametricQuantumCircuit.
```

This was discovered while adding cross-backend execution tests for `FQAOAConverter` (#415): FQAOA sampling with runtime `gammas`/`betas` fails on QuriParts because `fqaoa_state` applies Givens rotations (`qmc.control(qmc.ry)`) before the parametric cost layer. QAOA is unaffected because it uses no controlled gates. Qiskit and CUDA-Q are unaffected.

## Problem

The failure is **not** FQAOA-specific. A minimal reproduction with a single controlled-RY (compile-time bound) followed by a runtime-parametric RZZ triggers the same error:

```python
@qmc.qkernel
def ctrl_then_param(theta: qmc.Float, gamma: qmc.Float) -> qmc.Vector[qmc.Bit]:
    q = qmc.qubit_array(2, "q")
    q[0] = qmc.h(q[0])
    cry = qmc.control(qmc.ry)
    q[0], q[1] = cry(q[0], q[1], angle=theta)   # controlled gate
    q[0], q[1] = qmc.rzz(q[0], q[1], gamma)     # runtime-parametric gate
    return qmc.measure(q)

QuriPartsTranspiler().transpile(
    ctrl_then_param, bindings={"theta": 0.3}, parameters=["gamma"]
)
# -> ValueError: The given parameter Parameter(name=gamma) does not
#    belong to this LinearMappedParametricQuantumCircuit.
```

For comparison, the same kernel **without** the controlled gate transpiles and samples correctly. The bug only manifests when a controlled gate precedes a runtime-parametric gate in the same kernel.

## Cause

When emitting a controlled gate, `blockvalue_to_gate` (`qamomile/circuit/transpiler/passes/emit_support/controlled_emission.py:648`) builds the controlled-U body as a **sub-circuit** by calling `emit_pass._emitter.create_circuit(...)`.

The QuriParts emitter's `create_circuit` (`qamomile/quri_parts/emitter.py:128-142`) mutates **shared emitter state** and never restores it:

```python
def create_circuit(self, num_qubits, num_clbits):
    circuit = qp_c.LinearMappedUnboundParametricQuantumCircuit(num_qubits)
    self._current_circuit = circuit   # switches to the sub-circuit
    self._param_map.clear()           # clears the parameter map
    return circuit
```

After the controlled gate's sub-circuit is built and converted to a gate, `_current_circuit` still points at the sub-circuit and `_param_map` has been cleared. The next runtime-parametric gate then:

1. `create_parameter(name)` registers the parameter against the **sub-circuit** (`self._current_circuit.add_parameter(name)`), then
2. `emit_rzz` (and friends) add the parametric gate to the **main** circuit, whose `_check_param_exist` rejects the parameter because it belongs to a different circuit instance.

Root cause: `create_circuit` is reused for sub-circuit construction but clobbers the parent emit context (`_current_circuit`, `_param_map`) without save/restore. This is a QuriParts backend emitter state-management bug — the IR and the qkernel are valid (Qiskit/CUDA-Q emit them correctly).

## Proposed fix

Save and restore the emitter's `_current_circuit` and `_param_map` around the sub-circuit construction so that controlled-gate emission does not leak state into the parent circuit. Concretely, in `blockvalue_to_gate` (or by giving the emitter a dedicated sub-circuit-construction entry point that does not touch shared state):

```python
saved_circuit = emit_pass._emitter._current_circuit
saved_param_map = emit_pass._emitter._param_map
try:
    sub_circuit = emit_pass._emitter.create_circuit(qubit_count, 0)
    emit_pass._emitter._emit_operations(sub_circuit, ...)
    gate = emit_pass._emitter.circuit_to_gate(sub_circuit, "U")
finally:
    emit_pass._emitter._current_circuit = saved_circuit
    emit_pass._emitter._param_map = saved_param_map
return gate
```

This fixes every kernel that uses a controlled gate before a runtime parameter, not just FQAOA.

A regression test should cover:
- the minimal `controlled-gate -> runtime-parametric-gate` kernel above (sampling + expval) on QuriParts, and
- FQAOA sampling with runtime `gammas`/`betas` on QuriParts (currently worked around in `TestCrossBackendSampling` in #415 by detecting and omitting this specific failure).

jij-yuichiro-nakano:
## Summary

Fixes #416.

- Preserve QURI Parts emitter circuit state while controlled-gate emission probes a wrapped block as a sub-circuit.
- Restore emit-pass parameter metadata when block-to-gate probing falls back to gate-by-gate decomposition.
- Add QURI Parts regressions for a controlled gate followed by a runtime-parametric gate, including both sampling and expectation-value paths.
- Add FQAOA runtime `gammas` / `betas` sampling coverage on QURI Parts.

## Testing

- `uv run pytest tests/transpiler/backends/test_quri_parts_symbolic.py::TestSymbolicEndToEnd::test_controlled_gate_before_runtime_param_sample_and_expval tests/optimization/test_fqaoa.py::test_fqaoa_quri_parts_runtime_parameter_sampling -m quri_parts -q`

## Notes

- `uv run zuban qamomile/` currently fails before type-checking with `error: unrecognized subcommand qamomile/`; the installed zuban CLI expects a subcommand such as `check`.


ksk-jij (commented):
@jij-yuichiro-nakano Hi, thank you for working on this bug! I read through the changes and checked what this PR is trying to fix. The `@contextmanager` approach is interesting, but I wonder whether we need to update the common standard emission path for this. The bug is currently specific to QURI Parts, and I think the current QURI Parts `create_circuit()` implementation is stateful in a risky way, unlike Qiskit's emitter.

Longer term, I think we should make QURI Parts `create_circuit()` less stateful, but that would be a larger refactor. For this PR, could we instead follow the CUDA-Q approach and override `_blockvalue_to_gate()` in the QURI Parts transpiler so it returns `None` directly and skips sub-circuit probing?

My understanding is that this should fix the bug because QURI Parts `circuit_to_gate()` always returns `None` anyway, so the current probe can never succeed. Skipping it should avoid corrupting the parent circuit's parameter state and go straight to the existing fallback path. Could you check whether this actually fixes the regression?

jij-yuichiro-nakano:
@ksk-jij 

Thank you for your comment! I understand your intention now. I updated the implementation so that it is completed within the QURI Parts emit pass. 

Changes made:
- Merged the latest `origin/main` into `fix-issue-416`.
- Kept the fix backend-local to QURI Parts by relying on the QURI Parts emit-pass override for block-to-gate probing and the backend controlled fallback path from latest main.
- Removed the previously added generic controlled-emission context preservation and the QURI Parts emitter context manager from this branch.
- Added an explicit `rtol=0.0` to the QURI Parts symbolic expval regression assertion.

Verification run:
- `uv run pytest -m quri_parts tests/transpiler/backends/test_quri_parts_symbolic.py tests/optimization/test_fqaoa.py -q`

All checks passed. The QURI Parts targeted pytest run reported `21 passed, 5 deselected`.

Commit: `d4876f91`

ksk-jij (approved):
@jij-yuichiro-nakano Thank you for checking my comments :) I to be fair didn't mean to fix this issue in the current main, but apparently what I was going to ask you to do was in there; however, the test case is nice. Looks good to me :)

PR #441, opened 2026-06-01
Support seed specification for the qulacs sampler via QURI Parts

ksk-jij:
## Summary

`QuriPartsExecutor` exposed no way to seed the qulacs simulator used under QURI Parts, so sampling was always non-deterministic and experiments (paper benchmarks, tutorials) could not be reproduced from Qamomile alone. The QURI Parts integration tutorial explicitly noted this gap.

This adds an optional `seed` to the QURI Parts execution surface:

- `QuriPartsExecutor.__init__` and `QuriPartsTranspiler.executor` accept `seed: int | None = None`.
- When a seed is set, a seeded qulacs vector sampler is built. The installed QURI Parts version's high-level `create_qulacs_vector_sampler()` does not accept a seed, so a small helper reproduces QURI Parts' own qulacs state-vector sampling path (`convert_circuit(circuit).update_quantum_state(state)` then `state.sampling(shots, seed)`, using the seeded `qulacs.QuantumState.sampling(count, random_seed)` overload). Two `sample(...)` calls with the same seed and circuit then return identical shot counts.
- When `seed is None`, behavior is unchanged (non-deterministic). The seed is ignored when a custom `sampler=` is supplied, since an arbitrary sampler has no standard seed interface; this is documented on both `__init__` and `executor`.
- The seeded path always uses `QuantumState.sampling` rather than QURI Parts' (unseedable) multinomial fast-path at very large shot counts; the distribution stays statistically identical. This is noted in the helper docstring.

The EN and JA QURI Parts integration tutorials now construct a seeded executor and the closing reproducibility note is corrected accordingly.

## Test plan

- New `tests/quri_parts/test_seed.py` (guarded by `pytest.importorskip` and the `quri_parts` marker) asserts:
  - same seed → identical counts, across two executors and reused on one executor, parametrized over `n in {1, 2, 3}` and seeds `{0, 1, 42}` / `{0, 7, 123}`;
  - different fixed seeds (0 vs 1) → different counts (deterministic, not probabilistic);
  - an unseeded executor still samples correctly;
  - the constructor path (`QuriPartsExecutor(seed=...)`) matches the factory path (`transpiler.executor(seed=...)`).
- `uv run ruff check qamomile/ tests/` — clean.
- `uv run ruff format --check qamomile/ tests/` — clean.
- `uv run zuban check qamomile/` — 0 errors in changed files (pre-existing errors elsewhere match the `main` baseline).
- `uv run pytest tests/` — 8126 passed, 284 skipped.
- `uv run pytest tests/quri_parts/ -m quri_parts` — 70 passed (includes 15 new seed tests).
- `uv run pytest -m docs -k "en/integration/quri_parts"` — passed.


ksk-jij:
## Implementation summary

**What changed:** `QuriPartsExecutor` previously exposed no way to seed the qulacs simulator used under QURI Parts, so sampling was always non-deterministic and experiments (paper benchmarks, tutorials) could not be reproduced from Qamomile alone. This PR adds an optional `seed` to the QURI Parts execution surface so that qulacs sampling can be made reproducible.

**`qamomile/quri_parts/transpiler.py`:** `QuriPartsExecutor.__init__` gains a `seed: int | None = None` argument stored on `self._seed`. The lazy `sampler` property branches on the seed: when `seed is None` it keeps the previous behavior (`create_qulacs_vector_sampler()`), and when a seed is set it builds a seeded sampler instead. A new module-level helper `_create_seeded_qulacs_vector_sampler(seed)` is introduced because the installed QURI Parts version's high-level `create_qulacs_vector_sampler()` does not accept a seed; the helper reproduces QURI Parts' own qulacs state-vector sampling path (`convert_circuit(circuit).update_quantum_state(state)` then `state.sampling(shots, seed)`, using the seeded `qulacs.QuantumState.sampling(count, random_seed)` overload) and returns a `collections.Counter`. A docstring note records that this path does not switch to QURI Parts' (unseedable) multinomial fast-path at very large shot counts, so the distribution stays statistically identical. `execute()` is unchanged in shape (it still calls `self.sampler(circuit, shots)`), with the seed captured inside the cached sampler closure so repeated calls on one executor reuse the same seed. `QuriPartsTranspiler.executor` gains a matching `seed: int | None = None` argument and forwards it, so the seed is reachable through the public factory as well as the constructor. The seed is only applied to the default qulacs sampler; it is ignored when a custom `sampler=` is supplied, since an arbitrary sampler has no standard seed interface, and this is documented on both `__init__` and `executor`.

**`tests/quri_parts/test_seed.py` (new):** Regression tests, guarded by `pytest.importorskip` and the `quri_parts` marker, sample a uniform-superposition kernel and assert that the same seed (across two executors and reused on one executor) yields identical counts (parametrized over `n in {1, 2, 3}` and seeds `{0, 1, 42}` / `{0, 7, 123}`), that different fixed seeds (0 vs 1) yield different counts (deterministic, not probabilistic), that an unseeded executor still samples correctly, and that the `QuriPartsExecutor(seed=...)` constructor path matches the `transpiler.executor(seed=...)` factory path.

**Docs (`docs/en` and `docs/ja` integration/quri_parts_support.py):** The QAOA example now constructs `QuriPartsExecutor(seed=42)` so the tutorial is reproducible, with a comment explaining the seed, and the closing reproducibility note is corrected from "the Qulacs sampler does not seed its RNG" to an explanation that passing a seed makes sampling reproducible while dropping it restores non-deterministic behavior. EN and JA are kept in parity.

**Scope note:** The seed support is intentionally kept QURI Parts-specific. Promoting a `seed` contract to the base `Transpiler.executor()` / `QuantumExecutor` (so Qiskit Aer's `seed_simulator` and CUDA-Q's global `set_random_seed` would be honored under a shared API) is a cross-backend change with backend-specific semantics — CUDA-Q's seed is process-global, and real-hardware executors have no seed concept — and is deliberately left out of this PR.

**Verification:** `ruff check` and `ruff format --check` on `qamomile/` and `tests/` are clean; `zuban check qamomile/` introduces no new errors versus the `main` baseline; the default unit suite passes (8126 passed, 284 skipped); the `quri_parts`-marked suite passes (70 passed, including 15 new seed tests); the EN docs tutorial test passes; and all CI checks on this PR are green, with the Copilot review reporting no comments.

---

docs/ja/integration/quri_parts_support.py
```diff
@@ -164,7 +164,10 @@ def qaoa_ansatz(
 from qamomile.quri_parts import QuriPartsExecutor, QuriPartsTranspiler
 
 transpiler = QuriPartsTranspiler()
-executor = QuriPartsExecutor()
+# `seed` を渡すと Qulacs sampler が再現可能になります。同じ seed と回路で `sample(...)` を
+# 2 回呼ぶと、まったく同じショットカウントが得られます。非決定的なサンプリングにしたい場合は
+# この引数を省略する（または `seed=None` を渡す）だけです。
```
ksk-jij: [must]
In Japanese, we should not artificially insert breaks since they cause wired space in Japanese.

ksk-jij (commented):
Please update it.

docs/ja/integration/quri_parts_support.py
ksk-jij: Fixed in 160cacf: collapsed the seed comment so each Japanese sentence is on a single line, removing the artificial mid-sentence breaks. The JA docs test still passes.

---

CLAUDE.md
```diff
@@ -436,7 +436,12 @@ Never include **bare** `@username` or `@org/team` mention tokens in normal GitHu
...
+
 This rule targets **editorial** content — prose humans read. URLs that are functional metadata consumed by tooling rather than read by people (e.g., `$schema` references in JSON / YAML config files, dependency URLs in lockfiles, IDE / editor schema hints) are out of scope and may be added when the tool requires them.
 
 - ✅ "Implements the Trotter circuit (see #337 for the design discussion)."
 - ❌ "Implements the Trotter circuit (see `<external-url>`)." (the URL is shown as a `<external-url>` placeholder rather than a real address so this example itself doesn't render as a clickable external link; outside this kind of placeholder, never paste a literal external URL in prose unless the user explicitly provided it)
+- ✅ "Implements the stochastic-differential-equation quantum algorithm."
+- ❌ "Implements the algorithm from arXiv:2507.06198." (a bare external article ID — drop it unless the user supplied the reference)
+- ❌ "Closes the qulacs-seed gap (BACKLOG-8115)." (an internal backlog ID — describe the work in prose instead and leave the tracker key out of the public record)
```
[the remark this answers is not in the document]
ksk-jij: Good catch — fixed in cefec00. The example bullets now use placeholders (`BACKLOG-XXXX`, `arXiv:XXXX.XXXXX`) instead of the real-looking IDs, so the guideline's own examples no longer embed concrete identifiers. The prose still notes that a real arXiv ID is acceptable in actual writing (arXiv is the explicitly allowed exception); only private/internal tracker keys must stay out.

---

PR #443, opened 2026-06-02
chore: enable zuban CI on qamomile/ with zero static-type findings

ksk-jij:
## Summary

- Add a Zuban GitHub Actions workflow that runs `uv run zuban check qamomile/` on every push and pull request, mirroring the existing Ruff workflow, and fix CLAUDE.md to document the correct `check` subcommand.
- Clear every static-type finding under `qamomile/` so the new workflow passes on its first run. Most fixes widen public signatures to match runtime contracts that were already in place (`expval` accepts a single `Qubit` handle for 1-qubit observables; `HermitianMatrix.__mul__` / `__truediv__` accept `np.number`; `ValueSubstitutor.__init__` widens its `value_map` to `Mapping[str, ValueBase]` for covariance), lift `num_controls` onto `ControlledUOperation` so generic-typed callers see a typed slot, drop the narrower `_borrowed_indices` redeclarations in `Vector` / `Matrix` / `Tensor` (the parent's wider Union already matches runtime), collapse the QiskitExecutor V1/V2 Estimator dispatch onto one `cast(Any, self._estimator).run` lookup, and add localised assertions / casts / `# type: ignore` annotations where zuban's Generic / Protocol narrowing falls short of frontend-side invariants. A small docs sweep brings the example notebooks in line: an explicit `dict[tuple[int, ...], float]` annotation on the `from_hubo` example, an `isinstance` narrowing for the `BinarySampleSet | SampleSet` decode result in `qaoa_graph_partition`, `dict[str, Any]` on a list-then-ndarray dict in `hamiltonian_simulation`, and a typo-fix on `griffe.Expression` → `griffe.Expr` in the API-doc generator.
- Keep the qkernel parameter contract strict: kernel parameter annotations stay restricted to a single Qamomile handle (`UInt`, `Float`, `Bit`, `Vector[...]`, ...) or the corresponding Python primitive (`int`, `float`, `bool`) — primitive-and-Handle Unions such as `int | UInt` continue to be rejected by `handle_type_map`. The runtime literal auto-promotion via `_promote_literal_to_handle` is unchanged, so sub-kernel call sites can still pass raw `0` / `0.5` literals; the docs example in `08_reuse_patterns.py` keeps a `# type: ignore[arg-type]` with a comment that documents the deliberate gap between the static signature and the runtime convenience.

## Test plan

- [ ] \`uv run zuban check qamomile/\` reports \`Success: no issues found\`.
- [ ] \`uv run ruff check qamomile/ tests/\` and \`uv run ruff format --check qamomile/ tests/\` pass.
- [ ] \`uv run pytest tests\` is green (7905 passed locally on this branch).
- [ ] \`uv run pytest -m docs tests/docs -k \"08_reuse_patterns or qaoa_graph_partition or hamiltonian_simulation or binary_model\"\` passes on the updated docs examples.
- [ ] The new \`.github/workflows/zuban.yml\` job runs and stays green on this PR.


ksk-jij:
# Detailed change summary

This is a structured walkthrough of the diff for reviewers who only know `main`. It mirrors the internal working summary for this branch.

## 1. Problem

`zuban` is a mypy-compatible, Rust-based static type checker. `CLAUDE.md` documented the dev command as `uv run zuban qamomile/`, but the actual `zuban` CLI requires the `check` subcommand, so following the documented command fails outright with an "unrecognized subcommand" error. Even when run correctly as `uv run zuban check qamomile/`, `main` reports roughly 80 type errors under `qamomile/`.

Most of those errors are not bugs but static-analysis artifacts: signatures narrower than the implementation actually accepts, false positives from Generic + `isinstance` narrowing, and `TYPE_CHECKING` imports that cannot resolve optional dependencies. The Python runtime and the existing test suite all pass on `main`.

Behavior on `main`: `zuban` errors out at the CLI level, or reports dozens of type errors when invoked correctly; CI does not type-check at all.

Behavior on this branch: `uv run zuban check qamomile/ docs/` completes with zero errors (254 source files) — not just `qamomile/` but `docs/` too. Ruff and pytest keep passing. A new GitHub Actions workflow `.github/workflows/zuban.yml` runs `uv run zuban check qamomile/` on every push / PR. `docs/` is made type-clean (section 3.13), but the CI gate is intentionally scoped to `qamomile/` only (sections 3.14 / 5.1).

## 2. Frontend (user-facing) changes

The user-visible API changes are limited to widening signatures so they honestly express inputs the runtime already accepted. New ways of accepting input are added; existing call sites keep working unchanged.

### 2.1 `expval` accepts a single `Qubit`

`qm.expval(qubits, hamiltonian)` originally declared `qubits` as `Vector[Qubit] | tuple[Qubit, ...]`. The implementation body (the `else` branch in `qamomile/circuit/frontend/operation/expval.py`) already handled both a `Vector` and a bare single `Qubit`, so the runtime worked — but passing a `Qubit` was a static type error. The branch widens the signature to `Qubit | Vector[Qubit] | tuple[Qubit, ...]` (and updates the docstring), so a 1-qubit observable can be measured with `qmc.expval(q, H)` where `q` is a single `Qubit`, and the type checker is happy.

### 2.2 `HermitianMatrix.__mul__` / `__truediv__` accept `np.number`

The signatures were `scalar: int | float | complex`, but the body already gated on `isinstance(scalar, (int, float, complex, np.number))`, so NumPy scalars (`np.float64`, etc.) worked at runtime while being statically rejected. The branch widens the signatures to `int | float | complex | np.number`; the implementation is untouched.

### 2.3 Type annotation added to the `from_hubo` docs example

`BinaryModel.from_hubo(hubo: dict[tuple[int, ...], float], ...)` is unchanged. Python infers a mixed-arity dict literal like `{(0,): 1.0, (0, 1): 0.5, (0, 1, 2): 0.1}` as the narrow union `dict[tuple[int] | tuple[int, int] | tuple[int, int, int], float]`, and because `dict` is invariant in its key type, passing it to a function expecting `dict[tuple[int, ...], float]` is statically rejected. The docs examples (`docs/en/usage/binary_model.py`, `docs/ja/usage/binary_model.py`) annotate the literal as `hubo: dict[tuple[int, ...], float] = {...}`, aligning Python's inference with the function signature without touching the library.

### 2.4 qkernel parameter annotations stay a single handle or a single primitive (no Unions)

The qkernel decorator accepts a kernel parameter declared either as a Qamomile handle (`UInt` / `Float` / `Bit`) or as the corresponding Python primitive (`int` / `float` / `bool`); the latter resolves through the `TYPE_MAPPING` in `qamomile/circuit/frontend/func_to_block.py` and the tuple-membership checks in `_create_bound_input`. A "primitive-and-handle" union such as `int | qmc.UInt` is rejected at runtime by `handle_type_map` with an "Unsupported type annotation" error — this is the same as `main` and is deliberately preserved.

This is an intentional constraint following Qamomile's design stance that a kernel's parameter contract is expressed as either a single handle or a single primitive. Silently accepting `int | UInt` would open a loophole where users could write a Union even though the contract should be one or the other. The runtime literal auto-promotion (`_promote_literal_to_handle`) is preserved, so a sub-kernel call site can still pass raw `0` / `0.5` literals; the docs `08_reuse_patterns.py` example keeps a `# type: ignore[arg-type]` with a comment documenting the deliberate gap between the static signature and the runtime convenience.

### 2.5 `UInt` implements `__index__` (satisfies `SupportsIndex`)

`UInt` in `qamomile/circuit/frontend/handle/primitives.py` gains `__index__()`. `__index__` declares "this object is usable losslessly as a Python `int`" and satisfies `typing.SupportsIndex`.

The motivation is the slice-with-handle pattern `q[0 : n - 1]` where `n: qmc.UInt`. Python processes this in two stages: it first constructs `slice(0, n - 1)`, then calls `Vector.__getitem__(slice_obj)`. The type checker was rejecting the first stage — the typeshed `slice.__init__` signature requires each component to be `SupportsIndex | None`, so a `UInt` bound triggers "Slice index must be an integer, SupportsIndex or None". Implementing `__index__()` makes the slice construction type-check.

Runtime semantics: a constant `UInt` returns its underlying `int`; a symbolic `UInt` raises `TypeError` with a descriptive message (symbolic values cannot collapse to a concrete `int`; `Vector` slicing is dispatched directly and never goes through `__index__`). Crucially this does not change `Vector` slicing — Python does not call `__index__` while building `slice(0, n-1)` (it just stores the components), and `Vector.__getitem__(slice)` dispatches on the type of `slice.start` / `slice.stop` without going through `__index__`. Adding `__index__` does not change `isinstance(x, int)` (that inspects the class MRO, not protocol satisfaction). Verified by review that the change introduces no behavior change across `qamomile/`, `tests/`, and `docs/`.

## 3. Backend (IR / passes / internal helpers) changes

### 3.1 Drop the subclass `_borrowed_indices` redeclarations

`ArrayBase` declares `_borrowed_indices: dict[tuple[str, ...], "tuple[UInt, ...] | ArrayBase[T]"]`. On `main`, `Vector` / `Matrix` / `Tensor` each redeclared the same field with a narrower `dict[tuple[str, ...], tuple[UInt, ...]]` (missing the `ArrayBase[T]` member), but the code actually stores `ArrayBase[T]` values there, so the narrower declaration was a type lie. The branch deletes the three subclass redeclarations so they inherit the parent's wider Union; the runtime is unchanged and zuban stops flagging the inconsistency.

### 3.2 Lift `num_controls` onto `ControlledUOperation`

`ControlledUOperation` gains `num_controls: int | Value = 1`. Previously only `ConcreteControlledU` and `SymbolicControlledU` declared the field, so generic code typed against `ControlledUOperation` (e.g. `qamomile/circuit/estimator/_engine.py`) could not see `op.num_controls`. Each concrete subclass still redeclares the field with its narrow type, so the runtime is unchanged. The base `= 1` is a dead default: dataclass field-ordering rules (the parent `Operation`'s `operands` / `results` both have defaults) force a default here, and reordering ahead of `power` would require `@dataclass(kw_only=True)`, which is overkill. The docstring documents that the base is never instantiated directly and `= 1` is a slot reservation matching `ConcreteControlledU`'s single-control default.

### 3.3 `_emit_binop`: `hasattr` to `isinstance`

`_emit_binop` updates a scalar result handle's `init_value` after folding a constant BinOp. `main` gated this with `hasattr(result, "init_value")`, which zuban cannot use for narrowing, so the subsequent `type(result.init_value)(folded)` was a type error. The branch switches to `isinstance(result, (UInt, Float, Bit))`. The fold-failure fallback keeps the original `result.init_value = folded` with a `# type: ignore[assignment]`. Behavior is equivalent (see section 6.3 for why the two gates capture the same set).

### 3.4 `ValueSubstitutor.__init__` takes `Mapping[str, ValueBase]`

The `value_map` parameter widens from `dict[str, ValueBase]` to `Mapping[str, ValueBase]`. `dict` is invariant in its value type but `Mapping` is covariant, so a `dict[str, Value]` caller (the phi resolution in `compile_time_if_lowering.py`) type-checks. Only read APIs are used, so the weakening is runtime-safe. `compile_time_if_lowering.py` also gains an `assert isinstance(sub_r, Value)` to convey to zuban a narrowing the implementation already guarantees.

### 3.5 qiskit Estimator V1/V2 dual-call cleanup

`QiskitExecutor`'s expectation-value path dispatches over the Estimator V1 (`run(circuits, observables, parameter_values)`) and V2 (`run([(circuit, observable, params)])`) APIs via try/except. `main` needed long `# type: ignore[arg-type,call-arg,list-item]` chains on each `.run(...)` call. The branch lifts `estimator_run = cast(Any, self._estimator).run` once before the try and calls it on both paths. The `cast(Any, ...)` is a runtime no-op; both V1 and V2 estimators always have `.run`, so moving the attribute lookup out of the try changes nothing (see section 6.4). An `assert kind is not None` is also added after the `RuntimeOpKind.NOT` branch (the `__post_init__` already rejects `None`).

### 3.6 `# type: ignore` on VectorView branches

In `array.py` / `cast.py` / `pauli_evolve.py`, several `isinstance(self, VectorView)` / `isinstance(source, VectorView)` branches access `_slice_*` attributes. The `Vector`-to-`VectorView` isinstance narrowing interacts poorly with the Generic + dataclass MRO analysis and zuban reports "unreachable", so each line gets a `# type: ignore[unreachable]` or `[attr-defined]`. No runtime change.

### 3.7 `# type: ignore[import-not-found]` on `quri_parts` / `qulacs` imports

The `TYPE_CHECKING` blocks and in-function lazy imports across `qamomile/quri_parts/{transpiler,observable,emitter}.py` get `# type: ignore[import-not-found]`. `quri-parts-core` etc. ship a `py.typed` marker (typed package), so zuban tries to analyze the import and reports "import-not-found" when the optional dependency is absent; `cudaq` / `qbraid` are untyped and skipped. This keeps the imports optional without adding `quri-parts-qulacs` to dev deps. The lazy `import qulacs` and `from quri_parts.qulacs.circuit import convert_circuit` inside `_create_seeded_qulacs_vector_sampler` get the same ignore.

### 3.8 Defensive assertions / casts / unreachable ignores in emit

Where zuban cannot narrow along the IR path, assertions / casts encode invariants the construction side guarantees: `assert isinstance(input_array, ArrayValue)` / `assert isinstance(result_array, ArrayValue)` in `pauli_evolve_emission.py` (the frontend `pauli_evolve()` always feeds an `ArrayValue`); `cast(SliceArrayOperation, ...)` twice in `constant_fold.py` (the call site is the SliceArrayOperation-only path); `# type: ignore[unreachable]` on the defensive `case _: return None` tails in `eval_utils.py`; and the same on the defensive tails in `trotter.py`.

### 3.9 Protocol-narrowing suppression in `analyzer.py`

`qamomile/circuit/visualization/analyzer.py` accesses `ValueBase` Protocol members (`v.name or ""`, `v.logical_id in param_values`, `formal.type`) with `# type: ignore[unreachable,index,operator,attr-defined]`. The `ValueBase` Protocol declares `name: str` (non-None), an abstraction that does not match the runtime fact that `name` can be the empty string.

### 3.10 Local annotations in docs build scripts

`docs/scripts/build_doc_tags.py` annotates the yaml-frontmatter result as `fm: dict[str, Any]` / `raw_tags: Any` (the latter intentionally `Any` so the following `isinstance` defensive check is not seen as always-true). `docs/{en,ja}/algorithm/hamiltonian_simulation.py` annotates `errors: dict[str, Any]` (the variable is reassigned from dict-of-list to dict-of-ndarray). `docs/api_gen/annotations.py` fixes a typo: the argument type was `str | griffe.Expression | None`, but `griffe` does not export `Expression`; the correct name is `Expr`. This was a latent defect that never fired because the annotation stays a string at runtime.

### 3.11 Union narrowing in the `qaoa_graph_partition` docs

`QAOAConverter.decode(...)` returns `BinarySampleSet | ommx.v1.SampleSet` depending on the converter source. The docs examples build from an OMMX `Instance`, so the result is always `ommx.v1.SampleSet`, but accessing `.summary` / `.best_feasible` on the union is flagged. An `assert isinstance(sample_set, ommx.v1.SampleSet)` (plus an in-cell `import ommx.v1`) narrows it.

### 3.12 Restored `# type: ignore` in the `08_reuse_patterns` tutorial

Per section 2.4, the qkernel parameter annotation stays a single handle or a single primitive. The `rotate_first` in `docs/{en,ja}/tutorial/08_reuse_patterns.py` keeps `idx: qmc.UInt`, `angle: qmc.Float`, with a `# type: ignore[arg-type]` on the literal call site and a comment explaining the deliberate gap. The paired `.ipynb` is re-synced via `jupytext --update` (source cells only; outputs preserved, since the change is runtime-equivalent).

### 3.13 All `docs/` zuban errors cleared (CI gate stays on `qamomile/` only)

This branch also brings `docs/` to zero zuban errors (`uv run zuban check qamomile/ docs/` reports Success for 254 source files). Roughly:

- Library-side findings surfaced only when `docs/` is also analyzed: the sympy `Expr.subs` calls in `qamomile/circuit/estimator/_loop_executor.py` and `resource_estimator.py` widen the substitution-dict annotation to `dict[Any, Any]` to satisfy sympy's invariant `Mapping[Basic | complex, Expr | complex]` signature.
- Docs-side library-stub gaps narrowed via assertions: assert the `remove_final_measurements(inplace=False)` result is non-None before `Statevector.from_instruction` (`09_hermitian_decomposition`, `mottonen_amplitude_encoding`); narrow `decode`'s `BinarySampleSet | SampleSet` via `assert isinstance(...)` in `pce_maxcut`; narrow OpenFermion `MolecularData.fci_energy` (`float | None`) in `vqe_for_hydrogen`; narrow `nn.Sequential.__getitem__` to `nn.Module` and assert each parameter's `.grad` is non-None in `hybrid_qnn`; narrow `griffe.load` results (`Object | Alias`) to `griffe.Module` and handle `Path | list[Path]` filepaths in `api_gen/*`.
- True stub gaps via localized `# type: ignore`: the qiskit-aer monkey-patch `QuantumCircuit.save_statevector` (`[attr-defined]`); the lazy `quri_parts.*` imports in `integration/quri_parts_support.py` (`[import-not-found]`, mirroring `qamomile/quri_parts/`); the `nn.Module`-narrowed layer `.weight` access in `hybrid_qnn` (`[union-attr]`) and the `Tensor.abs()` chain zuban misreads as "Tensor not callable" (`[operator]`); the `Docstring.parse(config.docstring_style)` call in `api_gen/docstrings.py` (`[arg-type]`) and the `module.filepath is None` branch in `api_gen/discovery.py` (`[unreachable]`); the `theta: qmc.Float = math.pi / 2` default in `04_controlled_gates` (`[assignment]`, a runtime-supported gap — see section 2.4).
- Other small fixes in `04_controlled_gates`: `expect_error(exc_type: type)` is tightened to `type[BaseException]` so `except exc_type` type-checks; an `n = q.shape[0]` rebinding (`q.shape[0]` is `int | UInt`) that widened the parameter's narrow `qmc.UInt` is renamed to `n_shape` to keep the signature contract intact.

The paired `.ipynb` notebooks are re-synced for the 16 affected files via `jupytext --update` (source cells only; outputs preserved). All these docs fixes are runtime-equivalent. Unit tests (7905) and docs tests (129) stay green.

### 3.14 CI workflow added (gate scoped to `qamomile/`)

`.github/workflows/zuban.yml` (new) mirrors `ruff.yml`: Python 3.11 + uv, `uv sync --frozen --group dev`, then `uv run zuban check qamomile/`. Triggers: `workflow_dispatch` / push to `main` / pull_request (`opened`, `reopened`, `synchronize`). `docs/` is type-clean (section 3.13) but the gate is intentionally scoped to `qamomile/` — docs articles are edited frequently, and letting an unrelated docs edit break the type-check job would defeat the gate's primary purpose of protecting library code. The workflow has an inline comment noting how to flip the gate to `qamomile/ docs/` once docs-side type discipline is established.

## 4. Alternatives considered and why the chosen approach won

### 4.1 The `from_hubo` dict-literal invariance

Candidates: (A) weaken the signature to `Mapping[tuple[int, ...], float]`; (B) use a `TypeVar` bound `dict[_KT, float]`; (C) leave the library unchanged and annotate the docs literal as `dict[tuple[int, ...], float]`; (D) force a `cast()`. (A) does not help because `Mapping` is also invariant in the key type (verified). (B) works via the bound but introduces a type variable into a signature whose result type is unchanged, adding cosmetic complexity. (C) chosen: the library signature stays a plain readable dict, and the docs example teaches the natural Python idiom of annotating a mixed-arity dict literal. Call sites that build a `dict[tuple[int, ...], float]` first need no change.

### 4.2 Whether the qkernel decorator should accept Union annotations

Candidates: (A) extend the decorator type-key dispatch so `int | UInt` reduces to its handle member, removing the call-site `# type: ignore[arg-type]`; (B) keep the `# type: ignore[arg-type]` on call-site literals and restrict the declaration to a single handle or single primitive; (C) wrap literals explicitly at the call site (`rotate_first(q, qmc.uint(0), ...)`), losing the runtime auto-promotion. (B) chosen: Qamomile's stance is that a kernel parameter contract is one handle or one primitive (both accepted by `handle_type_map`), and silently accepting a mixed Union splits user-code style into "handle camp", "primitive camp", and "Union camp", which is harder to review. The runtime literal convenience stays; the gap is absorbed by a call-site ignore. (A) was prototyped and works (~10 lines reducing the Union in `handle_type_map`) but conflicts with the stance, so it was rejected. (C) erases the "literals accepted directly" showcase.

### 4.3 dev dependency vs `# type: ignore` for quri-parts

Candidates: (A) add `quri-parts-core` etc. to dev deps so zuban resolves the import; (B) add `# type: ignore[import-not-found]` to every `qamomile/quri_parts/*.py` import. (A) creates an asymmetry with the other optional backends (cudaq / qbraid): zuban tries to analyze quri-parts only because it is a typed package, which is a zuban-behavior artifact unrelated to design symmetry, and bloating dev deps for one backend has weak motivation. (B) chosen, applied to both `TYPE_CHECKING` and in-function imports; the same pattern works if `cudaq` / `qbraid` become typed later.

### 4.4 `_emit_binop` fold-failure fallback

Candidates: (A) keep `result.init_value = folded` with `# type: ignore[assignment]`, preserving `main`'s runtime behavior; (B) make the fold-failure branch a `pass` (leaving a stale value). (B) is near-dead defensive code with no practical difference, but changing `main`'s semantics is out of scope for a zuban PR. (A) chosen.

### 4.5 qiskit Estimator V1/V2 dual-call

Candidates: (A) keep the multi-flag `# type: ignore` chain on each `self._estimator.run(...)`; (B) lift `estimator_run = cast(Any, self._estimator).run` before the try. (A) is a long, hard-to-read ignore chain. (B) moves the `.run` attribute lookup out of the try (both V1 and V2 estimators always have `.run`, so no AttributeError in practice), and the exceptions caught inside the try (the V2 signature `TypeError`, the V2 result `.data.evs` `AttributeError`) still fall back to V1. Readability wins; (B) chosen.

### 4.6 `ControlledUOperation.num_controls` representation

Candidates: (A) base dataclass field `num_controls: int | Value = 1` overridden by subclasses; (B) a base `@property` raising `NotImplementedError`; (C) drop from the base and narrow via `isinstance`; (D) reorder before `power` with `@dataclass(kw_only=True)`. (B) collides with dataclass `__init__` generation and the property setter absence (`self.num_controls = ...` hits the missing setter and `AttributeError`s), breaking subclass instantiation (76 tests fail). (C) does not resolve the zuban error on the generic `op.num_controls` access. (D) requires `kw_only=True` because of the parent's default-bearing `operands` / `results`; technically possible since all construction sites are kwargs, but a wide change for the sole gain of dropping the base `= 1`. (A) chosen, with the dead-default nature documented.

### 4.7 How to clear the `UInt` slice-bound type error

Candidates: (A) implement `__index__()` on `UInt` to satisfy `SupportsIndex`; (B) add `# type: ignore[misc]` to each slice call site in docs; (C) change `Vector.__getitem__`'s slice overload. (A) chosen: `UInt` represents an integer value, so implementing `__index__` is natural in the type system and a net gain (constant `UInt` becomes usable with `range()` / `int()` / list indexing — section 2.5). One method clears the "Slice index must be an integer" findings (76 of them on the origin/main baseline of `uv run zuban check qamomile/ docs/`, almost all the `q[0 : n - 1]` patterns in `04_controlled_gates.py`). (B) scatters identical ignores throughout the docs. (C) touches the slice path with high blast radius for a type-only gain.

### 4.8 Widen the CI gate to `docs/` or keep it on `qamomile/`

Candidates: (A) run `uv run zuban check qamomile/ docs/` in CI; (B) gate on `qamomile/` only, with docs cleaned in this PR and a comment for re-enabling. (B) chosen: docs are edited frequently, and a single unrelated docs edit (e.g. a new tutorial whose external-library types do not line up) breaking the type-check job would defeat the gate's primary purpose of protecting library code, and would invert into docs fixes blocking library PRs. Cleaning docs now and scoping the gate to `qamomile/`, with a one-line path to widen it later, defers the migration safely.

## 5. Known limitations

### 5.1 `docs/` is type-clean but not on the CI gate

`docs/` is cleared to zero zuban errors (section 3.13), but the gate covers `qamomile/` only (section 3.14). When a future docs edit introduces a docs-only zuban error, CI passes (the gate is `qamomile/`-only); docs cleanliness is only verified by running `uv run zuban check qamomile/ docs/` locally. Future fix: once docs-side discipline is established, flip the workflow's command per its inline comment. Because docs are already clean in this PR, that flip is a one-line change.

### 5.2 Sub-kernel literal call sites need a call-site `# type: ignore[arg-type]`

Per sections 2.4 / 4.2, the qkernel parameter annotation stays a single handle or single primitive. The runtime `_promote_literal_to_handle` auto-wraps `int -> UInt`, `float -> Float`, `bool -> Bit`, but declaring the kernel parameter as a handle (`idx: qmc.UInt`) and passing an `int` literal at the call site is flagged. Docs that want to call with literals use a call-site `# type: ignore[arg-type]` to mark the deliberate gap — a positive choice aligned with not blurring the handle contract.

### 5.3 Many `# type: ignore` for Generic-Protocol / VectorView narrowing

`array.py` / `cast.py` / `pauli_evolve.py` / `analyzer.py` / `binary_model/model.py` / `binary_model/expr.py` / `hermitian.py` / `trotter.py` carry `# type: ignore[unreachable]` `[attr-defined]` `[index]` `[operator]` from zuban's weak Generic narrowing / Protocol attribute resolution. These are not Qamomile logic bugs. Future fix: prune them as zuban's narrowing improves. Not registered in LIMITATIONS.md (internal style level, no user impact).

### 5.4 The `_emit_binop` fold-failure fallback stays type-violating

`qamomile/circuit/frontend/handle/handle.py`'s `result.init_value = folded` writes a raw `folded` back when `type(result.init_value)(folded)` raises, which can mismatch the declared type; suppressed with `# type: ignore[assignment]`. The branch effectively never fires given `evaluate_binop_values`'s return type (`float | int`). Future fix: tighten that return type and make the fallback a `pass` or `raise`; it carries a behavior change so it is separate from this zuban PR.

## 6. Changes that could alter behavior, and why they do not

This PR is dominated by type-check fixes, and no change alters the current runtime behavior. Some changes nonetheless touch runtime branches / attributes / exceptions, so each is recorded as a pair: why it could change behavior, and why it does not in the current code. Every "does not change" claim is backed both by all 7905 unit tests / 129 docs tests staying green and by the structural argument below.

### 6.1 Widened `expval` / `HermitianMatrix` signatures (sections 2.1 / 2.2)

Could change: widening the declared parameter type changes the accepted-input set. Does not change: both implementation acceptance checks (`expval`'s `else` branch; `HermitianMatrix`'s `isinstance(scalar, (int, float, complex, np.number))`) already accepted the wider input on `main`. The branch only widens the signature to match the implementation; not a single branch is changed. Inputs that passed still take the same path, and inputs that were rejected are still rejected in the same place.

### 6.2 `UInt.__index__` (section 2.5)

Could change: a new dunder changes how `UInt` behaves in Python integer contexts (`range()` / `int()` / list/numpy indexing / `slice()` construction). Does not change (current paths): the codebase has no execution path that passes a `UInt` handle directly to a Python integer builtin — if it did, it would already have raised `TypeError` on `main` (where `UInt` had neither `__int__` nor `__index__`), and all tests pass. The `int(b)` / `int(bit)` sites that look relevant (`qamomile/linalg/subspace.py:227-230`, `qamomile/circuit/transpiler/classical_executor.py:247`) operate on already-narrowed Python int / numpy int or measured bit values, not `UInt` handles. `Vector` slicing dispatches on `slice.start` / `slice.stop` by type and never goes through `__index__` (Python does not call `__index__` while building `slice(0, n-1)`).

Two directional changes exist, both preserving existing normal-path and error-path behavior: passing a constant `UInt` to a Python integer builtin now works instead of raising `TypeError` (a capability addition, not a change to a previously working result — and no existing path uses it); passing a symbolic `UInt` still raises `TypeError` with the same exception type, only the message text changes (and again no existing path uses it). `isinstance(uint, int)` is unaffected (it inspects the MRO, not protocol satisfaction). Review confirmed no behavior change across `qamomile/`, `tests/`, `docs/`.

### 6.3 `_emit_binop` `hasattr` to `isinstance` (section 3.3)

Could change: switching the gate from `hasattr(result, "init_value")` to `isinstance(result, (UInt, Float, Bit))` could capture a different set. Does not change: the `Handle` subclasses with an `init_value` field are exactly `UInt` (`int`) / `Float` (`float`) / `Bit` (`bool`); neither the `Handle` base nor the others (`Qubit` / `QFixed` / `Tuple` / `Dict` / `Observable` / `ArrayBase`) have it. So the two gates capture the identical set. `_emit_binop`'s `result` is either a `_make_result()`-derived `UInt` / `Float` / `Bit`, or a non-handle `Value` (both gates false), so the truth value is always the same.

### 6.4 qiskit Estimator V1/V2 `.run` lift (sections 3.5 / 4.5)

Could change: lifting `estimator_run = cast(Any, self._estimator).run` outside the try could change behavior in cases where the `.run` attribute access itself raises. Does not change: `self._estimator` is a `StatevectorEstimator` (V2), an `Estimator` (V1 / Aer), or a user-supplied expectation estimator, all of which have `.run`. In the normal path the attribute access never raises, so the result is identical inside or outside the try. Even with an invalid object lacking `.run`, `main` does "V2 attempt -> AttributeError -> catch -> V1 attempt -> AttributeError propagates" while the branch does "`.run` access -> AttributeError propagates immediately" — both end with `AttributeError` propagating (only the attempt count differs). The observed final result is the same in both normal and error paths.

### 6.5 Defensive assertions (library and docs, sections 3.4 / 3.8 / 3.13)

Could change: the added `assert isinstance(...)` / `assert ... is not None` raise `AssertionError` when false, whereas `main` had no assertion and proceeded. Does not change: each assertion encodes an invariant the construction side guarantees (frontend building the IR / `__post_init__` / SDK call contract) and never fires while it holds. Concretely: `PauliEvolveOp.qubits` is always an `ArrayValue` from `pauli_evolve()`; `RuntimeClassicalExpr.kind` is non-None per `__post_init__`; SliceArrayOperation's `substitute_value` always returns `Value`; on the docs side, `remove_final_measurements(inplace=False)` always returns a new circuit, `run_fci=True` always populates `fci_energy`, a `BinaryModel`-derived `decode` always returns a `BinarySampleSet`, `.grad` is non-None after `backward()`, and a subpackage `griffe.load` always returns a `Module`. These holding is backed by passing unit tests and docs tests (129) that actually execute the articles. If an invariant ever broke, `main` would fail one line later as an `AttributeError` / `TypeError`, while this PR fails closer to the cause as an `AssertionError` — same "passes while it holds" behavior.

### 6.6 Pure runtime no-ops (most of 3.1 / 3.2 / 3.6 through 3.12)

`# type: ignore[...]` additions, `cast(SliceArrayOperation, ...)` (runtime no-op), `dict` to `Mapping` parameter weakening (read-only), `dict[Any, Any]` annotations on `subs_dict` etc., dropping the subclass `_borrowed_indices` redeclarations (the parent's `default_factory=dict` is inherited, so the constructed object is identical), the `ControlledUOperation.num_controls` base addition (always overridden by the concrete subclass), and the docs `n = q.shape[0]` to `n_shape` rename (still using `q.shape[0]`, only the local name changes) — none of these change any runtime value, branch, or exception; they are purely static-type / naming-level changes.

PR #448, opened 2026-06-03
Fix tuple expval metadata through inline

ksk-jij:
## Summary

Fix tuple-form `qmc.expval((q[i],), observable)` so inlined sub-kernels keep the caller register's root metadata through InlinePass.

Add root parent UUID/index metadata to tuple-packed expval operands, remap it during value cloning/substitution, preserve it through canonicalization and IR serialization, and consult it before direct UUID fallback during emit-time qubit-map construction.

Add cross-backend regressions covering inlined vector/scalar qkernels, native scalar gates, broadcast and sliced view updates, and composite gate round trips.

## Tests

- `RUFF_CACHE_DIR=/private/tmp/qamomile-ruff-cache UV_CACHE_DIR=/private/tmp/qamomile-uv-cache uv run ruff check qamomile/ tests/`
- `RUFF_CACHE_DIR=/private/tmp/qamomile-ruff-cache UV_CACHE_DIR=/private/tmp/qamomile-uv-cache uv run ruff format --check qamomile/ tests/`
- `UV_CACHE_DIR=/private/tmp/qamomile-uv-cache uv run zuban check qamomile/`
- `PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS='-p no:cacheprovider' UV_CACHE_DIR=/private/tmp/qamomile-uv-cache uv run pytest tests/circuit/test_expval_tuple_frontend_patterns.py -m '' -q`
- `PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS='-p no:cacheprovider' UV_CACHE_DIR=/private/tmp/qamomile-uv-cache uv run pytest tests/circuit/test_vector_slicing.py -m '' -q`
- `PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS='-p no:cacheprovider' UV_CACHE_DIR=/private/tmp/qamomile-uv-cache uv run pytest tests/circuit/test_expval.py tests/circuit/ir/test_canonical.py tests/circuit/ir/test_serialize.py -q`


ksk-jij:
## English Summary

This branch fixes stale tuple-form `expval` metadata across `InlinePass`. On `origin/main`, tuple-form `expval((q[i],), obs)` records each tuple element's root array address in `ArrayRuntimeMetadata.element_parent_uuids` and `element_parent_indices`, then emit falls back to `QubitAddress(root_uuid, index)` when a flat element UUID is not registered. However, when `InlinePass` substituted a sub-kernel array parameter or scalar `Qubit` parameter with the caller-side register element, normal value fields such as `parent_array`, `element_indices`, and slice metadata were rewritten, but the UUID-bearing metadata inside `ArrayRuntimeMetadata` was not. As a result, inlined helpers could keep callee-side root UUIDs or the standalone sentinel `("", -1)`, leaving the compiled expval `qubit_map` empty and silently binding the observable to the wrong physical qubit.

The frontend API is unchanged. Existing user code such as `helper(q, obs) -> qmc.expval((q[1],), obs)` now keeps the caller-side element identity after inlining, so `qm_o.Z(0)` is remapped to the tuple slot's actual physical qubit instead of falling back to physical qubit 0. The scalar helper form is covered too: if a helper traces `expval((q,), obs)` as a standalone scalar but the caller passes `caller_vec[i]`, inline substitution now promotes the sentinel metadata to the caller element's root `(array_uuid, index)` whenever that root address is statically resolvable.

The implementation adds `remap_value_metadata_references()` in `qamomile/circuit/ir/value.py` and uses it from both `UUIDRemapper._clone_metadata()` and canonicalization so clone and canonical UUID rewriting share the same metadata-reference list. The helper remaps `CastMetadata`, `QFixedMetadata`, and `ArrayRuntimeMetadata` UUID/logical-id references while preserving the empty parent UUID sentinel as a non-Value UUID.

`ValueSubstitutor` now rewrites UUID-bearing metadata through the same value map used for regular fields. It updates cast source/qubit references, QFixed qubit references, array-runtime element UUIDs/logical IDs, and parent UUIDs. For sliced parents, it folds constant `slice_start` / `slice_step` chains into root-space indices. For scalar helper tuple metadata, it also calls `resolve_root_qubit_address()` on the substituted caller element and replaces `("", -1)` with the real root address when possible.

Regression coverage was added in `tests/circuit/test_expval_vector_element.py` for vector helpers, scalar helpers with gates, scalar helpers without gates, two-element scalar tuple helpers, broadcast/view/composite patterns, and seeded random-angle `RY` cases. The tests assert both the expectation value and the compiled expval `qubit_map`, across the existing backend fixture matrix. Additional IR-level tests cover metadata clone/substitution for `CastMetadata`, `QFixedMetadata`, and array-runtime metadata through canonicalization and serialization.

The branch also documents the remaining symbolic-root limitation in `LIMITATIONS.md`. Current `ArrayRuntimeMetadata` can only encode parent addresses as `(array_uuid: str, index: int)`, so it cannot represent a symbolic root index such as `(q.uuid, j)` for a caller-side element like `q[j:j+1][0]` where `j` is a `qmc.range(...)` loop variable. This gap is pinned by `tests/transpiler/test_value_mapping.py::TestArrayRuntimeMetadataSymbolicRootLimitations::test_scalar_inline_metadata_cannot_promote_symbolic_slice_parent`, marked `xfail(strict=True)`, and should be flipped once metadata can carry symbolic affine root expressions.

Validated locally with `ruff check`, `ruff format --check`, `zuban check qamomile/`, `pytest tests/transpiler/test_value_mapping.py -q`, and the full `tests/circuit/test_expval_vector_element.py -m '' -q` regression matrix.

PR #449, opened 2026-06-03
Fix whole-vector expval qubit remapping

ksk-jij:
## Summary
Fix whole-Vector expval observable remapping so Pauli indices bind through the compiled quantum segment's qubit map instead of falling through to identity when the register starts at a nonzero physical qubit.

Keep tuple-form element UUID lookup intact, then fall back to root array address enumeration for whole vectors and views so backend execution still performs a single runtime remap.

Add cross-backend regression coverage for offset whole vectors across native gates, sub-kernel calls, composite gates, whole-vector broadcast, view broadcast, and randomized RY preparations.

## Tests
- `uv run pytest -m "" tests/circuit/test_expval_whole_vector_mapping.py -q`
- `uv run pytest -m "" tests/circuit/test_expval.py tests/circuit/test_vector_slicing.py -k expval -q`


ksk-jij:
### 1. Problem Overview

On the latest `main`, Vector-element forms such as `expval((q[i],), obs)` and `expval(q[i], obs)` already fall back to the root array `(array_uuid, index)` address when the element UUID itself is not present in the quantum segment's `qubit_map`. However, the offset whole-register form `qmc.expval(q, obs)` with a whole `Vector[Qubit]` still had a gap: observable Pauli indices could bind directly to physical qubit `k` instead of the logical register slot `q[k]`. A typical case is allocating an ancilla first and then allocating `q = qmc.qubit_array(2, "q")`. In that layout, physical qubit 0 is the ancilla while `q[0]` and `q[1]` are physical qubits 1 and 2, so `expval(q, Z(0))` could observe the ancilla instead of `q[0]`, and `expval(q, Z(1))` could observe `q[0]` instead of `q[1]`.

The root cause was that a whole `Vector[Qubit]` does not carry explicit element UUIDs like a tuple-packed operand, but the non-view path in `EmitPass._build_qubit_map()` tried element UUID lookup first and did not send an unresolved root array through the same root-address enumeration used for views. Whole vectors created by `qubit_array(...)` are registered by resource allocation as `QubitAddress(array_uuid, i)`, so the emit pass must resolve the shape length and map each logical Pauli index through `QubitAddress(root_av.uuid, start + step * i)`.

This branch keeps the Vector-element fallback from `main`, and only when the tuple / Vector-element lookup produces an empty map does it continue into the shared root-address enumeration for whole vectors and views. This remaps observable index `i` to the physical qubit corresponding to frontend register slot `q[i]`. The implementation is in `_build_qubit_map()` at `qamomile/circuit/transpiler/passes/emit.py:296`; the Vector-element fallback is at `qamomile/circuit/transpiler/passes/emit.py:393` through `qamomile/circuit/transpiler/passes/emit.py:407`, and the whole-Vector fallback is at `qamomile/circuit/transpiler/passes/emit.py:409` through `qamomile/circuit/transpiler/passes/emit.py:434`.

### 2. Frontend-Level Change

For user-written `qmc.expval(q, obs)`, the meaning no longer depends on whether the register starts at physical qubit 0. Even if an ancilla or another register is allocated before `q`, `Z(0)` observes `q[0]` and `Z(1)` observes `q[1]`, matching the natural frontend semantics.

```python
import qamomile.observable as qm_o

@qmc.qkernel
def energy(obs: qmc.Observable) -> qmc.Float:
    anc = qmc.qubit("anc")
    anc = qmc.x(anc)
    q = qmc.qubit_array(2, "q")
    q[1] = qmc.x(q[1])
    return qmc.expval(q, obs)

# After this branch: qm_o.Z(0) maps to q[0], not anc; qm_o.Z(1) maps to q[1].
```

This does not change how users call `expval` or construct Hamiltonians. `qmc.expval(q, obs)` now produces logical-slot expectation values for registers prepared through native gates, sub-kernel calls, CompositeGate operations, whole-Vector broadcast, and view broadcast. These frontend forms are covered in the integrated expval remapping tests in `tests/circuit/test_expval_vector_element.py`, with qkernels at `tests/circuit/test_expval_vector_element.py:396`, `tests/circuit/test_expval_vector_element.py:413`, `tests/circuit/test_expval_vector_element.py:430`, `tests/circuit/test_expval_vector_element.py:449`, and `tests/circuit/test_expval_vector_element.py:465`.

### 3. Backend / IR-Level Change

No IR operation or public API was added. The change is limited to construction of the observable-remapping qubit map in the emit pass. `EmitPass._build_qubit_map()` first preserves the `main` behavior: it directly looks up tuple-packed `element_uuids`, and for missed Vector elements it falls back to root `(uuid, index)` metadata captured at trace time. If that resolves anything, it returns the mapping and preserves tuple / Vector-element behavior.

The branch adds the next step for a root `ArrayValue` whose lookup remains empty: run the same root-address enumeration used for views. It resolves the operand length from `shape[0]`, then maps each logical Pauli index `i` through `QubitAddress(root_av.uuid, start + step * i)` into the compiled quantum segment's physical qubit map. This matches the existing `ResourceAllocator` contract for `QInitOperation`, which registers whole arrays as `QubitAddress(result.uuid, i)`. The IR is not expanded into per-qubit operations, and the Hamiltonian is not pre-remapped at compile time; the branch continues to use the existing single runtime remap where `ProgramOrchestrator` applies `hamiltonian.remap_qubits(...)` once based on `CompiledExpvalSegment.qubit_map`.

The tests were folded into the `main`-side expval remapping file, `tests/circuit/test_expval_vector_element.py`. The backend fixture at `tests/circuit/test_expval_vector_element.py:39` runs the same tests across Qiskit, QURI Parts, and CUDA-Q, and the helper at `tests/circuit/test_expval_vector_element.py:653` asserts the compiled expval segment's `qubit_map` directly. Deterministic coverage starts at `tests/circuit/test_expval_vector_element.py:678`; seeded random RY coverage for native / sub-kernel / CompositeGate forms starts at `tests/circuit/test_expval_vector_element.py:698`; whole-broadcast coverage starts at `tests/circuit/test_expval_vector_element.py:721`; and view-broadcast coverage starts at `tests/circuit/test_expval_vector_element.py:742`.

### 4. Alternatives Not Taken

The branch does not pre-remap the Hamiltonian to physical qubit indices at compile time. The existing design stores `CompiledExpvalSegment.qubit_map` and has `ProgramOrchestrator` apply `hamiltonian.remap_qubits(...)` once at runtime, which avoids the double-remap problem for views. This branch follows that design and only fixes how the map is built.

The branch also does not expand a whole `Vector[Qubit]` into per-element operands at the IR level. Qamomile's IR design keeps concepts such as Vector measurement and composite operations abstract for as long as possible, leaving backend-specific concretization to emit time. The issue here was not the abstraction level of the IR, but a missing emit-time map resolution path, so fixing `_build_qubit_map()` with a root-address fallback is the smallest design-aligned change.

The branch does not keep a separate whole-Vector-only test file. Since the latest `main` already has `tests/circuit/test_expval_vector_element.py` for expval observable remapping, the whole-Vector offset regressions were added there as well. That keeps Vector element, bare Vector element, whole Vector, sliced view, and standalone tuple remapping invariants in one place.

### 5. Known Limitations

None.

PR #457, opened 2026-06-08
Fix emit value resolver array elements

ksk-jij:
## Summary

Fix emit-side value resolution for bound Vector elements, including VectorView slice elements, so loop bounds, helper-kernel arguments, and controlled-QPE sub-block parameters resolve to concrete values during emission.

## Details
- Enable array-element lookup for integer resolution and block-parameter binding.
- Resolve sliced VectorView elements by composing the slice affine map back to the root bound container.
- Prefer const_array metadata over explicit bindings when both are present, while keeping runtime-parameter arrays unresolved.
- Add frontend execution regressions for bound Vector elements, sliced VectorView elements, and controlled-QPE theta binding.

## Validation
- uv run zuban check qamomile/
- uv run pytest -m "" tests/circuit/test_frontend_cross_backend_execution.py -k "vector-view-value-element-helper or vector_view_qpe_controlled_u_parameter_sample" -q -rs

### DIFF
--- a/qamomile/circuit/transpiler/passes/emit_support/value_resolver.py
+++ b/qamomile/circuit/transpiler/passes/emit_support/value_resolver.py
@@ -379,7 +379,7 @@ class ValueResolver:
         param operand at the call site must resolve to a value to seed the
         callee's parameter bindings.
         """
-        return self.lookup_in_bindings(operand, bindings)
+        return self.lookup_in_bindings(operand, bindings, index_array=True)
 
     def bind_block_params(
         self,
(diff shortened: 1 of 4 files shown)


ksk-jij:
### English branch summary

### 1. Problem overview

On main, the emit-side `ValueResolver` had inconsistent paths for resolving array elements, so elements of `Vector[UInt]` and `Vector[Float]` could fail to resolve back to concrete values from `bindings` or `const_array`. In particular, `resolve_int_value()` called `lookup_in_bindings()` without enabling array indexing, so loop bounds such as `qmc.range(bounds[0])` could remain unresolved. `resolve_operand_for_binding()` also did not use array-element lookup, so parameter bindings such as `theta=view[0]` passed into controlled-U or composite-gate inner blocks could fail to materialize; in Qiskit this could leave an unbound `theta` inside `PGate(theta)` and fail during numeric conversion.

This branch aligns integer resolution, block parameter binding, and raw bound-value resolution around the same array-element resolution path in `qamomile/circuit/transpiler/passes/emit_support/value_resolver.py:371`, `qamomile/circuit/transpiler/passes/emit_support/value_resolver.py:442`, and `qamomile/circuit/transpiler/passes/emit_support/value_resolver.py:580`. It also handles Qamomile `VectorView` values, which are represented as sliced `ArrayValue` instances with `slice_of`, `slice_start`, and `slice_step`, by resolving `view[i]` back to the root container at `root_index = slice_start + slice_step * i` before extracting the concrete value.

### 2. Frontend-level behavior

For frontend qkernel code, bound `Vector` elements and sliced `VectorView` elements can now be used reliably as compile-time bound values. For example, passing `view[0]` from `gammas[1:3]` into a QPE controlled-U helper could leave the inner `phase_gate` parameter `theta` unbound on main, but this branch resolves it as `gammas[1] == math.pi / 2`, and the sampled phase is deterministically `0.25`.

```python
@qmc.qkernel
def phase_gate(q: qmc.Qubit, theta: float) -> qmc.Qubit:
    return qmc.p(q, theta)

@qmc.qkernel
def kernel(gammas: qmc.Vector[qmc.Float]) -> qmc.Float:
    counting = qmc.qubit_array(3, name="counting")
    target = qmc.qubit(name="target")
    target = qmc.x(target)
    view = gammas[1:3]
    phase = qmc.qpe(target, counting, phase_gate, theta=view[0])
    return qmc.measure(phase)
```

This behavior is covered by regression tests in `tests/circuit/test_frontend_cross_backend_execution.py:319` and `tests/circuit/test_frontend_cross_backend_execution.py:1088`. The Qiskit path executes through `sample()` and asserts that every shot decodes to `0.25`; the QuriParts and CUDA-Q paths use the backend fixture so they execute the same deterministic assertion when the optional SDK dependency is available, and skip via `importorskip` when it is not. The ordinary `VectorView` element helper-kernel case is also covered through execution assertions in `tests/circuit/test_frontend_cross_backend_execution.py:296` and `tests/transpiler/test_emit_base.py:1230`.

### 3. Backend and IR-level behavior

`ValueResolver.resolve_operand_for_binding()` now uses `lookup_in_bindings(..., index_array=True)`, so controlled-U and QPE inner-block parameter binding can resolve operands such as `gammas[i]` and `view[0]` to concrete values. This lets callers such as `bind_block_params()` in controlled emission keep using the shared emit-side resolver path instead of carrying special-case array-element logic.

`ValueResolver._index_into_array()` now first calls `_resolve_array_element_location()` to convert the immediate parent and local indices into a root `ArrayValue` plus root indices, then calls `_resolve_array_container()` to retrieve the concrete container. `_resolve_array_element_location()` is at `qamomile/circuit/transpiler/passes/emit_support/value_resolver.py:495`; for ordinary elements it preserves the index as-is, and for sliced `ArrayValue` objects it walks the `slice_of` chain and composes the affine map. Symbolic element indices or symbolic slice bounds still return `None` and flow into the existing unresolved/fallback behavior.

`_resolve_array_container()` was added at `qamomile/circuit/transpiler/passes/emit_support/value_resolver.py:552`. It prefers the parent array's `const_array` metadata, then falls back to `bindings[parent.name]`, then `bindings[parent.uuid]`. If the root parent is a runtime parameter, `_index_into_array()` stops resolution so placeholder-like concrete data is not accidentally indexed at emit time.

### 4. Alternatives considered

One option was to special-case `view[0]` at callers such as `controlled_emission.py` or `composite_gate_emission.py` while constructing local bindings. That would leave the same underlying issue spread across loop bounds, qubit indices, gate angles, QPE/composite inner-block parameters, and other emit paths, with a high risk of lookup-order drift between resolvers. This branch instead centralizes root-container resolution and slice affine composition in `ValueResolver`, so callers can keep using `resolve_int_value()`, `resolve_classical_value()`, and `resolve_operand_for_binding()`.

Another option was to rewrite `VectorView` elements into root array elements earlier in the IR pipeline. That would lower the abstraction level earlier than necessary. Qamomile's design keeps sliced `ArrayValue` meaning in the IR until emit time and resolves concrete addresses or bound values only when backend emission needs them, so this fix stays within the emit-side resolver.

### 5. Known limitations

When: future IR support for multidimensional partial slices where a slice affine map should apply to dimensions beyond the leading index. Why: the current Qamomile frontend `VectorView` is a one-dimensional `Vector` slice, so `_resolve_array_element_location()` composes the slice affine map over the leading index and treats remaining indices as ordinary nested-container indexing. Future fix: a PR that introduces multidimensional partial slices should add dimension-aware slice metadata and extend the resolver to compose affine maps per sliced dimension.

When: QuriParts or CUDA-Q is not installed in the local development environment. Why: the cross-backend regressions follow the backend fixture's `pytest.importorskip()` behavior, so missing optional SDKs are skipped. In local verification for this branch, Qiskit passed through `sample()` and `run()`, while QuriParts and CUDA-Q were skipped because their SDK dependencies were not installed. Future fix: CI or any environment with those backend SDKs installed will run the same deterministic assertions against QuriParts and CUDA-Q.

PR #458, opened 2026-06-08
Fix QPE fallback phase array resolution

ksk-jij:
## Summary

Fix QPE composite fallback phase extraction so it resolves bound Vector elements, frozen const_array elements, and sliced VectorView elements back to the root container before emitting manual controlled-phase kickback gates.

The fallback now preserves unresolved symbolic indices, symbolic slice bounds, and runtime parameter arrays as unresolved instead of baking placeholder values. This keeps the existing no-concrete-phase fallback behavior while covering concrete array element operands that previously returned None.

## Tests

- `uv run pytest tests/circuit/test_composite_gate.py tests/circuit/test_qpe.py -q -rs`

## Notes

QURI Parts and CUDA-Q execution tests are guarded with `importorskip` and skipped locally because those SDKs are not installed. The Qiskit frontend regression executes sampling and asserts the deterministic decoded phase and total shots.


ksk-jij:
### Branch Summary

This branch strengthens regressions around QPE phase parameters that come from `Vector[Float]` elements or sliced `VectorView` elements. In Qamomile, an expression such as `gammas[1:3][0]` is represented in IR as an element of a sliced `ArrayValue` carrying `slice_of`, `slice_start`, and `slice_step`, rather than as a separate `VectorView` value type. Resolving that element therefore requires mapping the view-local index back to the root container index with `root_index = slice_start + slice_step * local_index`. The code change fixes phase extraction for block-free QPE composite fallback emission, and the tests also cover the public `qmc.qpe(..., theta=gammas[1:3][0])` path that now works after the latest shared resolver changes from `main`.

### User-Facing Coverage

The user-facing API remains ordinary `qmc.qpe(...)`. This branch adds a qkernel-style regression where a sliced `Vector` element is passed as `theta`, then verifies Qiskit execution through `sample()` and asserts the deterministic decoded phase `0.25`. The implementation fix for the public `qmc.qpe(...)` controlled-U path comes from the shared resolver work already merged into `main`; this branch adds regression coverage so that behavior does not regress.

### Backend and IR Changes

`extract_phase_from_params()` is now treated explicitly as a QPE manual fallback helper. It validates the gate type and raises `EmitError` when called with a non-QPE `CompositeGateOperation` or with a non-composite operation. The QPE/QFT/IQFT-specific helpers `emit_qpe_manual()`, `emit_qft_with_strategy()`, and `emit_iqft_with_strategy()` also validate their expected `CompositeGateType` at entry.

Phase extraction now scans only `CompositeGateOperation.parameters`, not arbitrary operands. The block-free QPE fallback can represent a single phase parameter, so it returns a phase only when one concrete numeric parameter is resolved. If multiple concrete numeric parameters resolve, it raises `EmitError` instead of guessing. Array element, `const_array`, sliced `VectorView` element, and runtime parameter array behavior is delegated to the shared `ValueResolver.resolve_classical_value()` path; composite emission no longer carries a separate array root-indexing fallback.

### Tests

Direct unit tests cover bound array elements, `const_array` elements, bound `VectorView` elements, literal `VectorView` elements, symbolic slice bounds, runtime parameter arrays, symbolic indices, rejection of multiple concrete phase parameters, and mismatched QPE/QFT/IQFT helper calls. Frontend execution tests cover both the block-free QPE fallback shim and the public `qmc.qpe(...)` path. Qiskit executes both through `sample()` and asserts the deterministic result and total shots. QuriParts and CUDA-Q tests are guarded with `importorskip`, so they run when those SDKs are available.

### Scope and Known Limits

This branch is intentionally scoped to composite fallback phase extraction plus regression coverage for the public QPE path. It does not reimplement shared array-element resolution inside composite emission; that behavior now belongs to the shared resolver. In the local environment, QuriParts and CUDA-Q execution tests are skipped because those SDKs are not installed. On older `main` revisions without the shared resolver controlled-U binding fix, the public `qmc.qpe(..., theta=gammas[1:3][0])` regression would still fail in controlled-U binding before reaching backend sampling.

### Verification

- `uv run pytest tests/circuit/test_composite_gate.py tests/circuit/test_qpe.py -q -rs` passed with `111 passed, 4 skipped`.
- `uv run ruff check qamomile/ tests/` passed.
- `uv run ruff format --check qamomile/ tests/` passed.
- `uv run zuban check qamomile/` passed.
- Claude checked the Japanese branch summary against the implementation diff after the latest updates and reported no divergence.

PR #459, opened 2026-06-09
Fix normalize_problem_input to use to_hubo() for ommx.v1.Instance inputs

baptistechev:
Previously the `ommx.v1.Instance` branch in normalize_problem_input called `to_qubo()`, which raises a `RuntimeError` for any instance whose objective contains higher-order (degree-3+) terms. This silently broke `QAOAConverter` and other converters when called with a HUBO `ommx.v1.Instance` — the error message gave no hint that the converter itself supports HUBO.

Switch to `to_hubo()` + `BinaryModel.from_hubo()`. 
The two paths are equivalent for purely quadratic instances (QUBO diagonal (i,i) entries map identically to linear (i,) tuples in from_hubo), so no existing behavior changes. 

For higher-order instances the new path succeeds and passes a correctly populated spin_model.higher through to converters that support HUBO. 

Converters that reject HUBO (`QRACConverterBase`, `FQAOAConverter`) already guard via `spin_model.higher` in `__post_init__ / degree()` in `__init__` and now raise their own clear `ValueError` rather than OMMX's opaque `RuntimeError`.

Added a test for HUBO `ommx.v1.Instance` on different converters

### DIFF
--- a/qamomile/optimization/converter.py
+++ b/qamomile/optimization/converter.py
@@ -20,25 +20,30 @@ def normalize_problem_input(
     :class:`MathematicalProblemConverter` base class and
     :class:`~qamomile.optimization.pce.PCEConverter` (which deliberately
     does not inherit from that base) both delegate to this helper so the
-    OMMX deep-copy / ``to_qubo`` semantics live in exactly one place.
+    OMMX deep-copy / ``to_hubo`` semantics live in exactly one place.
 
     For OMMX inputs, the instance is deep-copied via a bytes round-trip
-    before ``to_qubo`` is called: ``to_qubo`` mutates the instance it is
+    before ``to_hubo`` is called: ``to_hubo`` mutates the instance it is
     called on (it appends slack decision variables for non-binary vars
     and absorbs constraints into the objective via the penalty method).
     Mutating the caller's instance silently is surprising; copying first
     leaves the caller's view untouched. The deep copy retains
     original-constraint metadata internally, so downstream
     ``evaluate_samples`` reports feasibility against the user's
-    *original* constraints, and slack bits added by ``to_qubo`` are
+    *original* constraints, and slack bits added by ``to_hubo`` are
     reconstructed back into the original decision variables (e.g.,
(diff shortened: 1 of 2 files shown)


tests/optimization/test_converter_ommx_sampleset.py
```diff
@@ -533,3 +533,70 @@ def test_qrao_decode_empty_list_returns_empty_result():
...
+    # caught independently of the BinaryModel conversion code.
+    # The objective x0*x1*x2 + x1*x3 - x0 must produce exactly one cubic key.
+    hubo_dict, _ = ommx.v1.Instance.from_bytes(instance.to_bytes()).to_hubo()
+    cubic_keys = [k for k in hubo_dict if len(k) == 3]
+    assert len(cubic_keys) == 1, f"expected exactly one cubic HUBO term, got {cubic_keys}"
+    # Use a set comparison so index ordering within the tuple does not matter.
+    assert set(cubic_keys[0]) == {0, 1, 2}, f"unexpected cubic variables {cubic_keys[0]}"
+    assert hubo_dict[cubic_keys[0]] == pytest.approx(1.0), (
+        f"x0*x1*x2 coefficient must be 1.0, got {hubo_dict[cubic_keys[0]]}"
+    )
```
ksk-jij: [my opinion]
These lines verify `ommx` behaivour.  We may want to verify `converter.spin_model`.

qamomile/optimization/converter.py
ksk-jij: [my opinion]
Should this be also changed into `to_hubo`?

ksk-jij (commented):
@baptistechev Thank you for the PR! Almost perfect to me! I've left two small comments. Could you check them for me?

tests/optimization/test_converter_ommx_sampleset.py
baptistechev: Right ! Let's me change that :smiley:

ksk-jij (approved):
@baptistechev Thank you for addressing my comments! Looks good to me :)

PR #464, opened 2026-06-15
docs(integration): add Qiskit integration tutorial

jij-yuichiro-nakano:
## Summary
Add a new bilingual Qiskit integration tutorial and align the existing QURI Parts tutorial with the same structure and tone.

## What changed
- Added English and Japanese Qiskit integration tutorials with executed notebooks.
- Updated the QURI Parts integration tutorials to match the Qiskit tutorial structure, import layout, comments, and wording.
- Updated the English and Japanese integration index pages to list the new Qiskit tutorial and align article titles.

## Test plan
- `./docs/build.sh page-build docs/ja/integration/qiskit_support.py docs/en/integration/qiskit_support.py docs/ja/integration/quri_parts_support.py docs/en/integration/quri_parts_support.py`
- `uv run pytest -m docs -q`


docs/ja/integration/qiskit_support.py
```diff
@@ -0,0 +1,579 @@
...
+    with_labels=True,
+    node_color="white",
+    node_size=700,
+    edgecolors="black",
+)
+plt.title(f"MaxCut graph: {num_nodes} nodes, {G.number_of_edges()} edges")
+plt.show()
+
+# %% [markdown]
+# ## `@qkernel`によるQAOAアンザッツの構築
```
ksk-jij, line 84: [my opinion]
Is there any reason why this section in this notebook and one in quri-parts are bit different? Like this notebook
- has additional comment `# QAOA回路を量子カーネルとして定義します。`,
- uses `qaoa_state` in `qaoa_ansatz` wheres quri parts directly uses `superposition`, `mixer_layer` and `cost_layer`, and
- has `qaoa_energy` wheres quri-parts does not.

docs/ja/integration/qiskit_support.py
```diff
@@ -0,0 +1,579 @@
...
+plt.tight_layout()
+plt.show()
+
+# %% [markdown]
+# 上の`AerSimulator`は固定の`seed_simulator`で構築しているため、同じ回路列に対して再現可能なサンプリング列が得られます。
+# この5ノードグラフでは、$H_C$の基底状態エネルギー付近まで収束するはずです。
+# ここで得た最適パラメータ(`opt_gammas`、`opt_betas`)を、以降の例でも使います。
+
+# %% [markdown]
+# ## `run()`による期待値計算
```
ksk-jij, line 322: [question]
Is there any reason why this section does not exist in quri-parts?

docs/ja/integration/qiskit_support.py
```diff
@@ -0,0 +1,579 @@
...
+).result()
+
+print(f"Executable.run() expectation: {energy_via_run:+.10f}")
+print(f"sample mean energy          : {res.fun:+.4f}")
+assert np.isfinite(energy_via_run)
+
+# %% [markdown]
+# QamomileのAPIだけで扱う場合は、`ExecutableProgram.run(...)`を使うのがおすすめです。
+# Qiskit回路を自分で扱いたい場合には`executor.estimate(...)`も使えますが、その場合はQiskitのパラメータ順や回路のバインド状態をユーザー側で管理する必要があります。
+# `QiskitExecutor`はデフォルトで`StatevectorEstimator`を生成し、現在のV2 estimatorインターフェースと古いV1インターフェースの両方を自動的に扱います。
```
ksk-jij: [my opinion]
`現在のV2 estimatorインターフェースと古いV1インターフェースの両方を自動的に扱います`
I think it would be nice to clarify when each of them is used.

docs/ja/integration/qiskit_support.py
```diff
@@ -0,0 +1,579 @@
...
+
+# %% [markdown]
+# ## Qiskit固有の高度な機能
```
ksk-jij: [my opinion]
"固有" might lead misunderstanding since CUDA-Q also allows at least some mid-measurement as far as I remember (could be also for?), but "固有" feels like those are really only for Qiskit. How about just "Qiskitの高度な機能".

docs/ja/integration/qiskit_support.py
ksk-jij, line 1: [my opinion]
Sorry that I didn't mention this before, but could you use 量子SDK instead of バックエンド since バックエンド could make users confused if they are real hardware or something. Indeed, in this notebook, `# ## 他のQiskitバックエンドの利用` section, which means another simulator in Qiskit. This could make people think we have another Executor or Transpiler but with Qiskit or something. (We have QBraidExecutor for Qiskit but it is not what we mention in this notebook.)

This comment is applicable to quri-parts and both english versions as well.

ksk-jij (commented):
@jij-yuichiro-nakano Thank you for the PR! I've left comments! Please have a look!

About my comments on the difference between quri-parts and qiskit, I thought it might be nice to use mostly the same structure except for ones each SDK does not support. Otherwise users might think it is not supported.

docs/ja/integration/qiskit_support.py
jij-yuichiro-nakano: Thanks for your suggestion!
I applied in e03bf430: Reworded the Qiskit and QURI Parts docs in both languages to describe them as quantum SDK integrations, and changed Qiskit execution-target wording so it is not conflated with Qamomile integrations.

docs/ja/integration/qiskit_support.py
jij-yuichiro-nakano: Thanks for your suggestion!
Fixed in e03bf430: Clarified that QiskitExecutor uses StatevectorEstimator and the V2 primitive interface by default when available, and falls back to the V1 run signature for custom or older estimators that do not accept the V2 call.

docs/ja/integration/qiskit_support.py
jij-yuichiro-nakano: Thanks for your suggestion!
Fixed in e03bf430: Renamed the advanced-feature section in both languages and removed wording that implied the features are exclusive to Qiskit.

docs/ja/integration/qiskit_support.py
jij-yuichiro-nakano: Thank you for pointing this out. I revisited this after adding the matching QURI Parts `ExecutableProgram.run(...)` section in 0c0d81b. Both pages now use the same structure: `qaoa_state` prepares the unmeasured state, `qaoa_ansatz` handles sampling, and `qaoa_energy` returns `qmc.expval(...)` for `run()`. I kept these helpers because the `run()` explanation still needs an expval qkernel, and Qiskit cannot estimate a measured circuit.

docs/ja/integration/qiskit_support.py
jij-yuichiro-nakano: Thank you for your question!
There was no particular reason why an explanation of run() was not included in the QURI Parts tutorial.
I fixed in 0c0d81b: Added the QURI Parts `ExecutableProgram.run(...)` expectation-value section, introduced the matching `qmc.expval(...)` qkernel, and moved the direct `estimate_expectation(...)` material under an advanced QURI Parts features section.

jij-yuichiro-nakano:
@ksk-jij 
Thank you for reviewing the code! 
I have addressed your comments and questions and updated the content accordingly. I would appreciate it if you could take another look when you have time.

docs/ja/integration/qiskit_support.py
```diff
@@ -0,0 +1,580 @@
...
+# :::{tip}
+# Qamomileの回転ゲートは$e^{-i\theta/2}$という規約に従います。
+# そのため、$1/2$係数の扱いはコスト層とミキサー層で少し異なります。
+# ミキサー層では`rx`に$2\beta$を渡すので、$1/2$が打ち消され、教科書通りの$e^{-i\beta X}$になります。
+# 一方、コスト層では`rzz`に$J_{ij} \cdot \gamma$を渡すため、$1/2$は残ります。
+# この係数の違いは変分パラメータ$\gamma$に吸収しています。つまり、ここで使う$\gamma$は教科書のQAOAの$\gamma$の2倍に相当します。
+# :::
+
+# %%
+# QAOA回路を量子カーネルとして定義します。
```
ksk-jij: [my opinion]
Sorry for asking this small stuff, but could you remove this and the corresponding comment in the English version if you have no particular reason?

ksk-jij (commented):
@jij-yuichiro-nakano Thank you for the update! I'm afraid of leaving this tiny comment, but I'd appreciate it if you could address that.

PR #469, opened 2026-06-19
Add algorithm/option_pricing

github-nakasho:
# 概要

Qamomileのドキュメントに、オプションプライシングのNISQ計算論文の実装を追加しました。


docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+# Install the latest Qamomile through pip! 
+# # !pip install qamomile
+
+# %%
+import numpy as np
+from scipy.optimize import minimize_scalar
+import matplotlib.pyplot as plt
+
+import qamomile.circuit as qmc
+from qamomile.circuit.algorithm.state_preparation import amplitude_encoding
```
jij-yuichiro-nakano: In the most recent major update, the location of `amplitude_encoding` has changed. Please update your import as follows:

```python
from qamomile.circuit.stdlib import amplitude_encoding

```

jij-yuichiro-nakano (changes requested):
Thanks for PR! I left some comments and please check them.

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
+# ---
```
jij-yuichiro-nakano, line 1: The purpose of the tutorial is to introduce an example of implementing a quantum algorithm using Qamomile. Since the theme of this tutorial is the application of QAE to option pricing, I think QAE should be included in the title. In addition, since Qamomile is being used, the fact that it involves a quantum computer is self-evident.

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+
+# %% [markdown]
+# ---
+# tags: [algorithm, finance, simulation]
+# ---
+#
+# # 量子コンピュータを用いたオプションプライシング
+#
+# ゲート型量子コンピュータ上で実行する量子振幅推定は、様々な応用が提案されています。
+# その中でも、金融工学のオプションプライシングは、その最たる例です。
```
jij-yuichiro-nakano: I thought "代表的な応用例の一つ" would be more appropriate than "最たる例".

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+# これまで、オプションプライシングには古典モンテカルロ法が用いられてきましたが、量子振幅推定はそれと比較して2次高速化をもたらすことが知られています。
```
jij-yuichiro-nakano: "2次高速化" feels like an ambiguous expression. Please clarify what the acceleration refers to (e.g., time, number of qubits).

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+import matplotlib.pyplot as plt
+
+import qamomile.circuit as qmc
+from qamomile.circuit.algorithm.state_preparation import amplitude_encoding
+from qamomile.qiskit import QiskitTranspiler
+
+# %% [markdown]
+# ## 背景
+#
+# ### 問題: オプションプライシング
```
jij-yuichiro-nakano, line 52: I feel that the explanation of option pricing is insufficient. In particular, the description of the method uses terms such as “アセット”, “オプション”, and “ペイオフ”. However, there are no definitions for these terms. At a minimum, please add an explanation of the problem definition.

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+#
+# 金融リスクの計算は、大きな需要があります。
+# モンテカルロ (MC) 計算はその中心的な計算手法であり、Value at Risk (VaR) の推定や店頭デリバティブのプライシング決定など、幅広く用いられています。
```
jij-yuichiro-nakano: I thought "価格決定" would be more appropriate than "プライシング決定".

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+# しかし収束が悪く、その誤差は $\varepsilon = \mathcal{O} (M^{-1/2})$ で減衰するという欠点があります。
+# ここで $M$ はサンプル数です。  
+# 対象となる確率分布が量子状態として準備される場合、量子振幅推定 (QAE) により、期待値などの統計量を $\mathcal{O} (M^{-1})$ で推定できることが知られています。
+# これは先ほどの古典MCに対し、二次の高速化をもたらすことがわかります。  
+# [Stamatopoulos et al. (2020)](https://quantum-journal.org/papers/q-2020-07-06-291/)では、先行研究のプライシング手法を拡張し、NISQ実機での計算を可能にする量子回路実装を示しました。
```
jij-yuichiro-nakano: The claim "NISQ実機での計算を可能にする量子回路実装を示しました" is too strong. Even with the introduced method, the quantum circuit would make practical application on NISQ devices extremely difficult. On the other hand, this method already holds sufficient value as research on QAE for option pricing. Therefore, I think it would be fine to remove the description regarding NISQ devices.

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+# モンテカルロプライシングの量子アルゴリズムの理論的枠組みを示した先行研究として、[Rebentrost et al. (2018)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.98.022321)があります。
+# この論文は、プライシング計算において、古典に対し2次高速化が得られることを理論的に示しました。
+# そして、特にヨーロピアンオプションとアジアンオプションに対して数値シミュレーションを行い、その有効性を示しました。
+# しかし、その成果は理論的な枠組みの提示と数値シミュレーションに留まり、実機での実行は示されていません。  
+# もう一つの重要な先行研究として、[Woerner & Egger (2019)](https://www.nature.com/articles/s41534-019-0130-6)があります。
+# これはプライシングに対してではなく、VaR や Conditional Value at Risk (CVaR) に QAE を適用した研究です。
+# 先ほどの[Rebentrost et al. (2018)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.98.022321)のペイオフ計算を改良し、必要量子ビット数・ゲート数の大幅な削減に成功しました。  
+# そこで[Stamatopoulos et al. (2020)](https://quantum-journal.org/papers/q-2020-07-06-291/)では、[Rebentrost et al. (2018)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.98.022321)のペイオフ計算手法をオプションプライシングに拡張し、NISQフレンドリーな実装を示しました。
+#
+# ## アルゴリズム: NISQでのオプションプライシング計算
```
jij-yuichiro-nakano: In contrast to the problem-setting discussion, I felt there was too much explanation of the proposed method. The purpose of this tutorial is to provide an implementation example using Qamomile, and the theoretical explanation only needs to support understanding of the implementation. The algorithm section should be made simpler.

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+# a 
+# = \sum_{i=0}^{2^n - 1} f(S_i) p_i 
+# = \mathbb{E} [f(S)] \tag{3}
+# $$
+#
+# となることがわかります。
+# ここで $\mathbb{E}[f(S)]$ は、オプション $S$ のペイオフ $f$ の平均値です。
+# すなわち、式(2)のような状態を作る演算子 $\mathcal{A}$ さえ構築できれば、QAEによりペイオフの平均値を計算できることがわかります。  
+# [Rebentrost et al. (2018)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.98.022321)は、ペイオフを補助レジスタにバイナリ表現で格納し、制御回転をかける手法を提案しました。
+# しかし、量子ビット数と回路depthが増大する問題を抱えていました。
```
jij-yuichiro-nakano: I thought "回路深さ" would be more appropriate than "回路depth".

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+# $$
+# \tilde{f} (i) 
+# = 2 \frac{f(i) - f_\mathrm{min}}{f_\mathrm{max} - f_\mathrm{min}} - 1 \tag{8}
+# $$
+#
+# のように $f(i)$ を規格化したもので、その値の範囲は $\tilde{f}(i) \in [-1, 1]$ となります。
+# ここで、$f_\mathrm{max} = \max_i f(i), f_\mathrm{min} = \min_i f(i)$ です。
+# (7)式から、ペイオフ量子ビットが $\vert 1 \rangle$ となる確率は
+#
+# $$
```
jij-yuichiro-nakano, line 95: Regarding the expansion, I believe it should be `sin^2 (c f_tilde + π/4) = 1/2 [1 - cos(2c f_tilde + π/2)] = 1/2 [1 - sin(2c f_tilde)]`.

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+# = \sum_i p_i \sin^2 \left\{ c \tilde{f}(i) + \frac{\pi}{4} \right\} 
+# = \sum_i p_i \frac{1}{2} \left[ 1 + \sin \left\{ 2 c \tilde{f}(i) + \frac{\pi}{2} \right\} \right] 
+# \underbrace{\approx}_{cf \ll 1} \sum_i p_i \left( \frac{1}{2} + c\tilde{f} (i) + \mathcal{O}(c^3 \tilde{f}^3)\right) 
+# = \frac{1}{2} + c \mathbb{E} [\tilde{f}(i)] + \mathcal{O} (c^3 \tilde{f}^3) \tag{9}
+# $$
+#
+# のようになります。
+# $P_1$ を振幅推定で推定すれば、既知の $c, f_\mathrm{min}, f_\mathrm{max}$ により $\mathbb{E}[f(S)]$ を算出することが可能になります。
+# しかし、ここでは $c$ の選び方が重要になります。
+# $c$ を小さくすると、(9)式の精度は良くなりますが、$P_1 \sim \frac{1}{2}$ に値が集中し、振幅推定精度が落ちることになります。
```
jij-yuichiro-nakano: The loss of accuracy for small `c` is not caused simply by `P1` being concentrated near `1/2`. Amplitude estimation produces an error `delta_P1` in the estimated probability, and recovering the normalized payoff requires division by `c`:

`E[f_tilde] ≈ (P1 - 1/2) / c`.

Therefore, the probability-estimation error is amplified to `delta_P1 / c` when converted back to the payoff expectation. Conversely, increasing `c` reduces this amplification but increases the bias from the linear approximation, which becomes `O(c^2)` after dividing the original `O(c^3)` remainder by `c`.

Could we revise this paragraph to describe the trade-off in terms of error amplification and approximation bias, rather than saying that amplitude estimation itself becomes less accurate near `P1 = 1/2`?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+offset     = 4 * c * (S_min - K) / F_max - 2 * c + np.pi / 2
+
+alpha_min = offset / 2
+alpha_max = (offset + (2**n_qubits - 1) * theta_S) / 2
+print(f"S_min での振幅角: {alpha_min:.4f}  "
+      f"(期待値 {2*c*(S_min-K)/F_max - c + np.pi/4:.4f})")
+print(f"S_max での振幅角: {alpha_max:.4f}  "
+      f"(期待値 {np.pi/4 + c:.4f})")
+
+mc_E_f  = float(np.sum(probs * np.maximum(S_grid - K, 0)))
```
jij-yuichiro-nakano: To me, `mc_E_f` looks like an exact expectation value calculated directly from the distribution, rather than an MC-based estimate. Did you intend to write the MC estimate here, or the exact expectation value?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+mc_fair = np.exp(-r * T) * mc_E_f
+print(f"\n古典 MC E[f(S)]:     {mc_E_f:.6f}")
+print(f"古典 MC フェアバリュー: {mc_fair:.6f}")
+
+E_ftilde_theory = 2 * mc_E_f / F_max - 1
+P1_theory       = c * E_ftilde_theory + 0.5
```
jij-yuichiro-nakano: I believe the theoretical value of P1 is the equation preceding Eq. (12) in the original paper. However, `P1_theory` is in the form using the first-order approximation (Eq. (13)). Which of these is correct?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+# この分布読み込みと、(21)式にあるように制御 $R_y$ ゲートである $\mathrm{CR}_y$ を施すことで、コールオプションのペイオフをペイオフ量子ビット振幅に変換する回路を構築します。
+# `num_shots` の回数だけ測定を行い、ペイオフ量子ビットが $\vert 1 \rangle$ になる確率 $P_1$ を推定します。
+# そして(13)式から近似式 $P_1 \approx \frac{1}{2} + c\mathbb{E}[\tilde{f}]$ から、ペイオフの期待値を算出します。
+# ここで算出された $P_1$ の値が理論値と近い値となることを示すことで、ペイオフ回路が正しく機能していることを確認しています。
+
+# %%
+# ============================================================
+# 2. ペイオフ回路(k=0 の動作確認)
+# ============================================================
+@qmc.qkernel
```
jij-yuichiro-nakano, line 268: The implemented payoff circuit does not currently encode the European call payoff `max(S - K, 0)`. Although the explanation introduces a comparator qubit, `payoff_call` applies the linear controlled rotations to every basis state, including states with `S < K`. Those states should receive the constant angle `pi / 4 - c`, while the additional payoff rotation should be applied only when `S >= K`.

Could we implement the comparator-controlled piecewise rotation described in the paper, or use an equivalent uniformly controlled rotation? The target encoded angle should be

`alpha(S) = pi / 4 - c + 2 * c * max(S - K, 0) / (S_max - K)`.

Please also correct the preceding text: the comparator should be set to `|1>` when `i >= K`, not when `i <= K`. As written, the circuit underestimates the option value because it includes the negative linear region below the strike.

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+def payoff_call(slopes_p: qmc.Vector[qmc.Float]) -> qmc.Bit:
+    q_S = qmc.qubit_array(n_qubits, "q_S")
+    q_p = qmc.qubit(name="q_p")
+
+    q_S = load_distribution(q_S)
+
+    cry = qmc.control(qmc.ry)
+    for j in range(n_qubits):
```
jij-yuichiro-nakano: Inside `qkernel`, it is recommended to use `qmc.range()` for classical loops. Is there a reason why this cannot be used here?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,597 @@
...
+axes[1].set_title('P₁^(k) = sin²((2k+1)θ)')
+axes[1].legend()
+
+plt.tight_layout()
+plt.show()
+
+# %% [markdown]
+# 今回の実装が、理論値に近い結果を得られていることがわかります。
+
+# %%
```
jij-yuichiro-nakano: Please add a summary section, similar to `multidimentional_qft.py`, providing a brief summary of the tutorial.

jij-yuichiro-nakano (changes requested):
Thanks for PR! I left some comments and please check them.

jij-yuichiro-nakano (commented):
@github-nakasho  Thank you for the PR! I've left comments, please have a look at them.

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+# これにより、少ない測定回数でも $\theta_a$ を高精度に推定することが可能になります。  
+# 今、アセットの価格を $S_i$とし、それが起こる確率を $p_i$ としましょう。
+# 演算子 $\mathcal{A}$ が
+#
+# $$
+# \mathcal{A} \vert 0 \rangle_{n+1} 
+# = \sum_{i=0}^{2^n - 1} \sqrt{1-f(S_i)} \sqrt{p_i} \vert S_i \rangle \vert 0 \rangle + \sum_{i=0}^{2^n - 1} \sqrt{f(S_i)} \sqrt{p_i} \vert S_i \rangle \vert 1 \rangle \tag{1}
+# $$
+#
+# のような状態を作れるとすると、$\vert 1 \rangle$ の振幅は $a = \mathbb{E}[f(S)]$ となり、QAE によりペイオフの期待値を計算できることがわかります。
```
jij-yuichiro-nakano: [must] Here, \(a = \mathbb{E}[f(S)]\) is the probability of measuring the payoff qubit in \(\lvert 1\rangle\), not its amplitude. In Eq. (1), the norm of the component associated with \(\lvert 1\rangle\) is \(\sqrt{a}\). Could you revise this to say that the probability of measuring the payoff qubit in \(\lvert 1\rangle\) is \(a = \mathbb{E}[f(S)]\)?

docs/ja/algorithm/option_pricing.ipynb
```diff
@@ -0,0 +1,727 @@
...
+    "                q_S, q_p = apply_payoff_ucr(q_S, q_p)\n",
+    "\n",
+    "            _ = qmc.measure(anc)\n",
+    "            _ = qmc.measure(q_S)\n",
+    "            return qmc.measure(q_p)\n",
+    "\n",
+    "    return Ak\n",
+    "\n",
+    "\n",
+    "k_list      = [0, 1, 2, 4, 8, 16]\n",
```
jij-yuichiro-nakano, line 553: [must] The committed notebook output does not appear to match the latest code. Although `k_list` includes `16`, the saved output stops at `k=8`, and this code cell has no execution count. Please rerun the notebook from the beginning and commit the regenerated outputs so that the source and displayed results are consistent.

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+def load_distribution(q: qmc.Vector[qmc.Qubit]) -> qmc.Vector[qmc.Qubit]:
+    """P_X: 対数正規分布を Möttönen でロード"""
+    q = amplitude_encoding(q, amps.tolist())
+    return q
+
+# qmc.inverse で随伴を自動生成
+load_distribution_inv = qmc.inverse(load_distribution)
+
+
+# %% [markdown]
```
jij-yuichiro-nakano, line 261: [must] These references still point to Eqs. (21) and (13), but the revised tutorial only numbers Eqs. (1)-(4). There is also a later code comment referring to "Eq. 5." Please update all equation references to match the current numbering.

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+    P1_k  = ones / shots_per_k
+    P1_th = np.sin((2*k + 1) * theta_a_theory)**2
+    print(f"k={k:2d}: P1 測定={P1_k:.4f}  理論値(厳密)={P1_th:.4f}")
+
+
+# %% [markdown]
+# 複数の $k$ での測定データ (二項分布) をベースに、$P_1^{(k)} = \sin^2 ((2k+1) \theta_\alpha)$ というモデルに対する最尤推定を実施しましょう。
+# これにより振幅角 $\theta_\alpha$ を推定し、さらに $P_1 = \sin^2 \theta_\alpha$ を計算します。
+# そしてこの $P_1$ からペイオフの期待値 $\mathbb{E}[f(S)]$ を計算し、無リスク金利で割り引くことで、オプションのフェアバリューを算出します。
+# 最後に、古典モンテカルロ法から求めた参照値との比較を行っています。
```
jij-yuichiro-nakano: The reference value is calculated using the deterministic weighted sum `np.sum(probs * payoff)`, rather than classical Monte Carlo sampling. Could you replace the remaining references to "classical MC" with wording such as "the reference value calculated from the discretized distribution"?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+axes[1].set_xlabel('k  (Grover iterations)')
+axes[1].set_ylabel('P₁ estimation')
+axes[1].set_title('P₁^(k) = sin²((2k+1)θ)')
+axes[1].legend()
+
+plt.tight_layout()
+plt.show()
+
+# %% [markdown]
+# 今回の実装が、理論値に非常に近い結果を得られていることがわかります。
```
jij-yuichiro-nakano: The statement that the result is "very close" to the theoretical value seems stronger than the displayed result supports. The committed notebook output has a fair-value relative error of approximately 6.7%, which is consistent with the approximation bias introduced by the finite value of \(c\). Could you quantify the error and mention this bias instead of describing the result simply as "very close"?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+
+# %% [markdown]
+# ## まとめ
```
jij-yuichiro-nakano, line 491: [must] To keep the tutorial structure consistent with `multidimensional_qft.py`, could you add a separate `## 結果` section before `## まとめ` and place the numerical results, plots, and their interpretation in that section?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+# 先ほどの[Rebentrost et al. (2018)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.98.022321)のペイオフ計算を改良し、必要量子ビット数・ゲート数の大幅な削減に成功しました。  
+# そこで[Stamatopoulos et al. (2020)](https://quantum-journal.org/papers/q-2020-07-06-291/)では、[Rebentrost et al. (2018)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.98.022321)のペイオフ計算手法をオプションプライシングに拡張し、NISQフレンドリーな実装を示しました。
+#
+# ## アルゴリズム: NISQでのオプションプライシング計算
+#
+# ### ペイオフの計算
+#
+# QAE ではグローバー演算子 $\mathcal{Q}$ を $k$ 回繰り返すことで、測定確率が
+# $\sin^2((2k+1)\theta_a)$ という既知の関数形で変化します。
+# これにより、少ない測定回数でも $\theta_a$ を高精度に推定することが可能になります。
```
jij-yuichiro-nakano: There are trailing spaces on lines 78, 107, 113, and 229 in the updated diff. Could you remove them?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+# ### 先行研究
+#
+# モンテカルロプライシングの量子アルゴリズムの理論的枠組みを示した先行研究として、[Rebentrost et al. (2018)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.98.022321)があります。
+# この論文は、プライシング計算において、古典に対し2次高速化が得られることを理論的に示しました。
+# そして、特にヨーロピアンオプションとアジアンオプションに対して数値シミュレーションを行い、その有効性を示しました。
+# しかし、その成果は理論的な枠組みの提示と数値シミュレーションに留まり、実機での実行は示されていません。  
+# もう一つの重要な先行研究として、[Woerner & Egger (2019)](https://www.nature.com/articles/s41534-019-0130-6)があります。
+# これはプライシングに対してではなく、VaR や Conditional Value at Risk (CVaR) に QAE を適用した研究です。
+# 先ほどの[Rebentrost et al. (2018)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.98.022321)のペイオフ計算を改良し、必要量子ビット数・ゲート数の大幅な削減に成功しました。  
+# そこで[Stamatopoulos et al. (2020)](https://quantum-journal.org/papers/q-2020-07-06-291/)では、[Rebentrost et al. (2018)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.98.022321)のペイオフ計算手法をオプションプライシングに拡張し、NISQフレンドリーな実装を示しました。
```
jij-yuichiro-nakano: Describing the implementation as "NISQ-friendly" appears to overstate what the cited work establishes. The complete option-pricing algorithm still requires amplitude-estimation circuits with repeated Grover operators, whose coherent circuit depth increases with \(k\). Avoiding phase estimation alone does not make the overall algorithm suitable for NISQ hardware. Could you either remove "NISQ-friendly" or qualify the statement by specifying the particular circuit component whose resource requirements are reduced?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+#
+# * アセット (原資産): 株式など、オプションの対象となる金融商品
+# * オプション: 将来の決められた時点 (満期) に、決められた価格 (権利行使価格) でアセットを売買する権利
+# * ペイオフ: 満期時点でのオプションの価値
+#
+# ### 問題: オプションプライシング
+#
+# 金融リスクの計算は、大きな需要があります。
+# モンテカルロ (MC) 計算はその中心的な計算手法であり、Value at Risk (VaR) の推定や店頭デリバティブの価格決定など、幅広く用いられています。
+# しかし収束が悪く、その誤差は $\varepsilon = \mathcal{O} (M^{-1/2})$ で減衰するという欠点があります。
```
jij-yuichiro-nakano: It is unclear what "しかし収束が悪く" is being compared against. Since the following expression already states the convergence rate precisely, could you remove this phrase and simply explain that the error decreases as \(\varepsilon=\mathcal{O}(M^{-1/2})\)?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+# ここで $M$ はサンプル数です。  
+# 対象となる確率分布が量子状態として準備される場合、量子振幅推定 (QAE) により、期待値などの統計量を $\mathcal{O} (M^{-1})$ で推定できることが知られています。
```
jij-yuichiro-nakano: Here, \(\mathcal{O}(M^{-1})\) describes the scaling of the estimation error rather than the estimated quantity itself. Could you make this explicit by changing the wording to "\(\mathcal{O}(M^{-1})\) の誤差で推定できる"?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+# これは先ほどの古典MCに対し、二次の高速化をもたらすことがわかります。  
+# [Stamatopoulos et al. (2020)](https://quantum-journal.org/papers/q-2020-07-06-291/)では、先行研究のプライシング手法を拡張した量子回路実装を示しました。
```
jij-yuichiro-nakano: The same work by Stamatopoulos et al. is introduced again in the following prior-work section, making this explanation somewhat repetitive. Could you discuss it in only one place? I think keeping the more detailed introduction in the prior-work section would make the structure clearer.

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+# 今、アセットの価格を $S_i$とし、それが起こる確率を $p_i$ としましょう。
+# 演算子 $\mathcal{A}$ が
+#
+# $$
+# \mathcal{A} \vert 0 \rangle_{n+1} 
+# = \sum_{i=0}^{2^n - 1} \sqrt{1-f(S_i)} \sqrt{p_i} \vert S_i \rangle \vert 0 \rangle + \sum_{i=0}^{2^n - 1} \sqrt{f(S_i)} \sqrt{p_i} \vert S_i \rangle \vert 1 \rangle \tag{1}
+# $$
+#
+# のような状態を作れるとすると、$\vert 1 \rangle$ の振幅は $a = \mathbb{E}[f(S)]$ となり、QAE によりペイオフの期待値を計算できることがわかります。
+# [Rebentrost et al. (2018)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.98.022321) は補助レジスタにペイオフをバイナリ表現で格納する手法を提案しましたが、量子ビット数と回路 depth が増大する問題がありました。
```
jij-yuichiro-nakano: For consistency with the Japanese text, could you replace "回路 depth" with "回路の深さ"?

docs/ja/algorithm/option_pricing.py
```diff
@@ -0,0 +1,525 @@
...
+
+
+# %% [markdown]
+# ### 量子振幅推定の実装
+#
+# 振幅推定のためのGrover演算子 $\mathcal{Q} = \mathcal{A} \mathcal{S}_0 \mathcal{A}^\dagger \mathcal{S}_\chi$ を $k$ 回適用する量子回路を構築しましょう。
+# $k$ 回適用した量子状態において、ペイオフ量子ビットが $\vert 1 \rangle$ になる確率 $P_1^{(k)} = \sin^2 ((2k+1) \theta_\alpha)$ を求めます。
+
+# %%
+def build_Ak_kernel(k: int):
```
jij-yuichiro-nakano, line 313: [question] Qamomile recently added Grover-related building blocks such as `grover_search` and `mcx`. Could any of these be reused to simplify `build_Ak_kernel`, particularly the manually implemented \(S_0\) reflection and ancilla ladder? If `grover_search` is not directly applicable because this amplitude-estimation circuit uses a custom state-preparation operator \(\mathcal{A}\), it would be helpful to clarify that as well.

jij-yuichiro-nakano (approved):

PR #470, opened 2026-06-20
fix: inline CallBlockOperations inside ControlledUOperation.block

ksk-jij:
## Problem

`qmc.control` embeds the controlled kernel's body as a nested `Block` in `ControlledUOperation.block`. `InlinePass` recursed into `InverseBlockOperation`'s nested blocks but not into `ControlledUOperation.block`, so a pass-through wrapper kernel whose body is a single `CallBlockOperation` (`def wrap(q): return leaf(q)`) kept that unexpanded call inside the controlled block. At emit, `blockvalue_to_gate` cannot turn a `CallBlockOperation` into a gate, so the wrapped unitary collapsed to the identity and the controlled gate was silently dropped. `control(leaf)` worked; only `control(wrap)` lost the gate.

Minimal repro on the Qiskit backend (prints `True` on `main` — the controlled X was dropped):

```python
import numpy as np
from qamomile.circuit import qkernel
from qiskit.quantum_info import Operator

@qkernel
def x_body(q: qmc.Qubit) -> qmc.Qubit:
    return qmc.x(q)

@qkernel
def wrap(q: qmc.Qubit) -> qmc.Qubit:   # body is a single call to x_body
    return x_body(q)

@qkernel
def circ() -> qmc.Bit:
    ctrl = qmc.qubit("ctrl"); q = qmc.qubit("q")
    ctrl, q = qmc.control(wrap)(ctrl, q)
    return qmc.measure(ctrl)

qc = QiskitTranspiler().transpile(circ, bindings={}).compiled_quantum[0].circuit
U = Operator(qc.remove_final_measurements(inplace=False)).data
print(np.allclose(U, np.eye(4)))   # main: True (gate dropped) -> branch: False
```

## Fix

`InlinePass._serialize_operations` now inlines calls inside `ControlledUOperation.block`, mirroring the existing `InverseBlockOperation` branch. `_has_any_call_block` and `count_call_blocks` also descend into it so the output `BlockKind` and the unroll-loop termination signal stay correct. The fix is at the IR level and emit is untouched, so it applies uniformly to every backend. Confirmed equivalent to `control(leaf)` across `num_controls` 1/2/3, `power` 1/2/3, and multi-level wrappers.

## Secondary: clearer error for control/inverse of a recursive kernel

Controlling (or inverting) a self-recursive kernel stays unsupported: the recursion lives inside `ControlledUOperation.block`, where `partial_eval` never folds the base-case `if` (its `OperationTransformer` only descends into `HasNestedOps` bodies, not operation-owned blocks). Previously this surfaced as the generic "did not terminate after 64 unroll iterations" error, which wrongly blamed the bindings — the recursion does terminate and its driver is concrete. A new `count_unrollable_call_blocks` helper lets `unroll_recursion` detect that every residual call is trapped inside an operation-owned block and fail fast (one iteration, not 64) with a cause-specific message naming `qmc.control` / `qmc.inverse` and the non-recursive-rewrite workaround. Genuinely non-terminating top-level recursion still gets the original message. The limitation is documented in `LIMITATIONS.md`.

## Tests

- `tests/circuit/test_controlled.py::TestControlledPassThroughWrapperInlined`: a white-box check that no `CallBlockOperation` survives inside `ControlledUOperation.block` after inline; a Qiskit unitary check that `control(wrapper) == control(leaf)` and the gate actually fires (1- and 2-level wrappers); and cross-backend (Qiskit / QuriParts / CUDA-Q, guarded by `importorskip`) sampling + expval parity with randomized seeds and boundary angles.
- `tests/circuit/test_self_recursive_qkernel.py::test_control_of_recursive_kernel_raises_targeted_error`: pins the targeted error and asserts the generic "did not terminate" wording is not used.

All new regression tests fail on `main` and pass on this branch.

## Verification

- `/local-review`: clean — `ruff check`, `ruff format --check`, and `zuban check` all pass.
- Full `tests/transpiler/` (925 passed) and `tests/circuit/` (5511 passed) suites: no regressions.


ksk-jij:
# Branch design summary (English)

Detailed design write-up for this PR, complementing the description above.

## 1. Problem overview

Qamomile is a quantum-circuit compiler: it traces a Python function decorated with `@qm.qkernel` into an IR `Block` (a list of `Operation`s) and compiles it to a circuit for each backend (Qiskit / QuriParts / CUDA-Q). `qmc.control(kernel)` is the API that builds a controlled version of a kernel; internally it emits a single `ControlledUOperation` IR op. That op holds the "controlled unitary" not as child operations but as a nested `Block` in its `.block` field.

Behavior on `main`: when the controlled kernel is a "pass-through wrapper" — its body just calls another kernel and returns it (`def wrap(q): return x_body(q)`) — the controlled gate produced by `qmc.control(wrap)` **collapses to the identity** (the inner X is silently dropped). Controlling the leaf kernel directly, `qmc.control(x_body)`, correctly yields a controlled X. The repro below prints `True` on `main` (identity = bug) and `False` on this branch (correct controlled X).

```python
import numpy as np
from qamomile.circuit import qkernel
from qiskit.quantum_info import Operator

@qkernel
def x_body(q: qmc.Qubit) -> qmc.Qubit:
    return qmc.x(q)

@qkernel
def wrap(q: qmc.Qubit) -> qmc.Qubit:   # body is a single call to x_body
    return x_body(q)

@qkernel
def circ() -> qmc.Bit:
    ctrl = qmc.qubit("ctrl"); q = qmc.qubit("q")
    ctrl, q = qmc.control(wrap)(ctrl, q)
    return qmc.measure(ctrl)

qc = QiskitTranspiler().transpile(circ, bindings={}).compiled_quantum[0].circuit
U = Operator(qc.remove_final_measurements(inplace=False)).data
print(np.allclose(U, np.eye(4)))   # main: True (controlled gate dropped) / branch: False
```

The root cause is in the `inline` pass (`InlinePass`). `inline` expands `CallBlockOperation` (the IR op for "a call to another kernel") into the callee's body, advancing a block from `HIERARCHICAL` (may contain calls) to `AFFINE` (no calls remain). This expansion must recurse into ops that carry a nested block, but `InlinePass` recursed into `InverseBlockOperation`'s nested blocks while never descending into `ControlledUOperation.block`. As a result, the `CallBlockOperation` to `x_body` in the wrapper's body reaches emit (backend circuit generation) still unexpanded inside `ControlledUOperation.block`. At emit, `blockvalue_to_gate` (`qamomile/circuit/transpiler/passes/emit_support/controlled_emission.py:1213`), which converts a nested block to a backend gate, walks the block's `Operation`s in order but cannot turn a `CallBlockOperation` into a gate, so it skips past it and produces an empty (identity) unitary. That is what "the controlled gate disappears" really is. `qmc.control(x_body)` was fine because `x_body`'s block has no `CallBlockOperation` and directly contains the X `GateOperation`.

In addition, `_has_any_call_block`, which decides whether `inline`'s output block is `AFFINE` or `HIERARCHICAL`, also did not look inside `ControlledUOperation.block`, so the top block was mislabeled `AFFINE` even while a call remained nested inside.

## 2. Frontend (user-facing) changes

The user-visible change is a single correctness fix: passing a pass-through wrapper kernel to `qmc.control` now correctly produces the same controlled gate as controlling the leaf directly. There is no change to the API, signatures, or error messages; cases that used to silently collapse to the identity now compile correctly.

```python
@qkernel
def x_body(q: qmc.Qubit) -> qmc.Qubit:
    return qmc.x(q)

@qkernel
def wrap(q: qmc.Qubit) -> qmc.Qubit:   # wrapper that just calls and returns x_body
    return x_body(q)

# main:   qmc.control(wrap) becomes the identity (X is dropped)
# branch: qmc.control(wrap) is the same controlled X as qmc.control(x_body)
```

This holds for `num_controls` (number of control qubits) >= 2 and for a specified `power` (number of times the unitary is applied) as well: in every case "via wrapper == controlling the leaf directly" matches. Multi-level pass-through (`wrap2(q): return wrap(q)`) expands the same way.

One related improvement is the clarity of a user-visible error. Passing a self-recursive kernel (one whose body calls itself and stops via a constant `if` base case) to `qmc.control` / `qmc.inverse` is still unsupported, but it previously produced a generic "did not terminate after 64 unroll iterations" error. In that case the recursion actually terminates and the bindings are correct, so that wording — which reads as "your bindings are wrong / bind a constant" — misattributes the cause. This branch replaces it with a cause-specific error that immediately states a self-recursive kernel was passed to `qmc.control` / `qmc.inverse` and shows the rewrite-non-recursively workaround. The behavior (failing as unsupported) is unchanged, but it now fails in one iteration instead of spinning 64 times.

```python
@qkernel
def rec(q: qmc.Qubit, k: qmc.UInt) -> qmc.Qubit:
    if k == 0:
        return qmc.x(q)
    return rec(q, k - 1)

# rec(q, 2) called directly still expands as before.
# qmc.control(rec)(ctrl, q, k=2) is unsupported:
#   main:   "did not terminate after 64 unroll iterations" (misattributes the cause)
#   branch: "qmc.control / qmc.inverse was given a recursive @qkernel ..." (cause-specific)
```

## 3. Backend (IR / overall) changes

The central fix is in `InlinePass` in `qamomile/circuit/transpiler/passes/inline.py`. As above, `InlinePass` expands `CallBlockOperation`s to take a block from `HIERARCHICAL` to `AFFINE`; it walks each `Operation` and must expand inside ops that carry a nested block. Like `InverseBlockOperation`, `ControlledUOperation` holds its body in a `block` field rather than as "child operation lists" (the `HasNestedOps` nested-traversal protocol implemented by For / If / While etc.), so the generic control-flow recursion never reaches it. This fix adds op-specific recursion in three places.

1. A `ControlledUOperation` branch was added to the operation-expansion body `_serialize_operations` (`qamomile/circuit/transpiler/passes/inline.py:275`). Mirroring the existing `InverseBlockOperation` branch (`inline.py:258`), it passes `op.block` to `_inline_nested_block` (`inline.py:451`) to expand the inner `CallBlockOperation`, swaps in the expanded block via `dataclasses.replace`, then applies the usual value substitution. `_inline_nested_block` expands only the calls in the body while keeping the nested block's `input_values` (formal parameters), and sets `kind` to `AFFINE` / `HIERARCHICAL` depending on whether any call remains, so `ControlledUOperation.block` keeps its abstraction as "a single self-contained unitary" while only the calls disappear.

2. `_has_any_call_block` (`inline.py:39`) now also inspects `op.block` recursively (`inline.py:48`). This function decides whether the top block after `inline` is `AFFINE` or stays `HIERARCHICAL`. Before the fix it mislabeled the top as `AFFINE` even with a call remaining inside `ControlledUOperation.block`.

3. `count_call_blocks` (`inline.py:63`) now also counts calls inside `op.block` (`inline.py:77`). This function is the termination signal of the fixed-point loop `unroll_recursion` (`qamomile/circuit/transpiler/transpiler.py`, capped at `MAX_UNROLL_DEPTH = 64`) that unrolls a self-recursive kernel one layer at a time. Counting the residual calls inside `ControlledUOperation.block` keeps the loop from overlooking a "calls still remain to expand" state.

An ordinary pass-through wrapper is fully expanded in one shot by the pipeline's first `inline` (`count_call_blocks` reaches 0) and `unroll_recursion` returns immediately, so this change adds no extra iterations. The fix is a minimal IR-level removal of calls; emit (`blockvalue_to_gate` etc.) is untouched, so it applies uniformly to every backend.

The other line of change is the error replacement when a self-recursive kernel is passed to `qmc.control` / `qmc.inverse`. `unroll_recursion` (`qamomile/circuit/transpiler/transpiler.py`) is the fixed-point loop that runs `inline` and `partial_eval` to unroll self-recursion one layer at a time, but when the recursion lives inside the nested block of `ControlledUOperation.block` / `InverseBlockOperation`, `partial_eval` (`ConstantFoldingPass` / `CompileTimeIfLoweringPass`) does not descend into those operation-owned blocks (its recursion base `OperationTransformer` only enters `HasNestedOps`), so the base-case `if` is never folded and the loop never converges. To distinguish this, a new helper `count_unrollable_call_blocks` (`inline.py:87`) was added. Unlike `count_call_blocks`, it does not descend into `ControlledUOperation.block` / `InverseBlockOperation` nested blocks; it counts only the calls the loop can still resolve (top level and inside `HasNestedOps` bodies — For / ForItems / While / If). When `count_call_blocks > 0` but `count_unrollable_call_blocks == 0`, every remaining call is trapped inside an operation-owned block — the signature of controlling/inverting a self-recursive kernel — so `unroll_recursion` (around `transpiler.py:324`) raises a cause-specific `FrontendTransformError` immediately instead of spinning to `MAX_UNROLL_DEPTH = 64`. Genuine non-terminating top-level recursion still gets the original "did not terminate after 64" generic message, so the two are not confused.

## 4. Alternatives considered and why this approach was chosen

There were two options for where to expand the `CallBlockOperation` inside `ControlledUOperation.block`. Option A is expansion inside `InlinePass` (the one taken); Option B is running inline-equivalent processing on the nested block at emit time (`_prepare_nested_block_for_emit`, `controlled_emission.py:1568`, or `blockvalue_to_gate`).

A was chosen over B because `inline` is precisely the stage that removes all `CallBlockOperation`s and finalizes `HIERARCHICAL` -> `AFFINE`. A call surviving `inline` into an `AFFINE` / `ANALYZED` block is itself a pipeline-contract violation; the later `analyze`, `canonicalize`, and serialization all assume an `AFFINE` / `ANALYZED` block has no residual calls. Since `InverseBlockOperation` already expands its nested blocks in `InlinePass`, expanding `ControlledUOperation` in the same place is consistent. B would carry the IR to emit in an inconsistent state (a call hidden inside a supposedly-affine block) and would also require reimplementing inline logic on the emit side, increasing duplication.

The "expand but do not over-lower to per-qubit / per-gate" principle is also preserved. `_inline_nested_block` removes only the calls and leaves `ControlledUOperation.block` at the "single unitary" abstraction, so a backend with a native controlled gate can emit it as one instruction while a backend without one falls back to library decomposition as before.

## 5. Known limitations

Passing a self-recursive kernel to `qmc.control` / `qmc.inverse` is unsupported on this branch too. When: the controlled/inverted kernel calls itself and stops via a constant-driven `if` base case. Why: `partial_eval` (`ConstantFoldingPass` / `CompileTimeIfLoweringPass`), which folds the base-case `if`, does not descend into the nested blocks of `ControlledUOperation.block` / `InverseBlockOperation` (its recursion base `OperationTransformer` (`qamomile/circuit/transpiler/passes/control_flow_visitor.py`) only enters `HasNestedOps`). So even though `inline` unrolls the self-call inside `block` one layer at a time, the base case is never folded and no fixed point is reached. For `qmc.inverse` the failure usually happens even earlier, at trace-time inverse-block construction, which rejects `if` / `while` / `for ... in items(...)` (a separate entry, "`qmc.inverse` requires trace-time-resolvable control flow"). On this branch the shape is detected via `count_unrollable_call_blocks` and, instead of spinning to `MAX_UNROLL_DEPTH = 64` and emitting the generic "did not terminate after 64" error that blames the bindings, a cause-specific error that names `qmc.control` / `qmc.inverse` and points to the rewrite-non-recursively workaround is raised in one iteration (see Sections 2 / 3). The behavior is "unsupported, explicit failure." This limitation is recorded in the `LIMITATIONS.md` entry "`qmc.control` / `qmc.inverse` of a self-recursive kernel is unsupported." Future fix: scoped partial evaluation into operation-owned blocks (resolve the controlled-U / inverse classification param operands to constants, map them onto the nested block's formals, then run `ConstantFoldingPass` + `CompileTimeIfLoweringPass` in that scope) would let the existing inline-unroll machinery converge and make this explicit error itself unnecessary.

The fact that the counting helpers (`_has_any_call_block` / `count_call_blocks`) do not recurse into `CompositeGateOperation.implementation` was initially considered a gap, but on inspection it is correct design, not a limitation. These need to recurse only into nested blocks that survive inline as ops while only their inner calls are expanded (`ControlledUOperation.block` / `InverseBlockOperation`), and they do. An inline-eligible custom composite is **expanded body and all and disappears** via `_inline_composite` inside `_serialize_operations` (no `CompositeGateOperation` remains in the IR), so there is no nested block to recurse into. Conversely `QFT` / `QPE` / `IQFT` stay as boxes but their implementation blocks contain no `CallBlockOperation` (verified empirically). Recursing into composites unconditionally would in fact count calls inside the intentionally-not-inlined `QFT` / `QPE` / `IQFT` and cause `unroll_recursion` to spin even though they are never expanded. So the current non-recursion is consistent and no extra handling is needed.

PR #474, opened 2026-06-21
fix: map returned whole-Vector control output of qmc.control

ksk-jij:
## Problem

A concrete-mode controlled-U with a whole `Vector[Qubit]` control prefix (e.g. `cg(qmc.qubit_array(2), target)`) silently lost the returned control vector's qubit addresses. `ConcreteControlledU` expands a whole-Vector control into per-element scalar operands/results, but the user-facing output handle was re-aggregated into a throwaway next-version `ArrayValue` whose per-element `QubitAddress(uuid, i)` keys were never written to the emit `qubit_map`. A later `measure` / element access of that vector then resolved to nothing — Qiskit emitted no measurement instruction and returned an all-zeros classical register. Passing the controls as individual scalar `Qubit` handles always worked because each output `Qubit` wraps the IR result scalar directly.

## Fix

Frontend (`qamomile/circuit/frontend/operation/control.py`): `_build_control_results` re-parents the per-element scalar control results of a whole-Vector / VectorView control onto the same freshly-versioned `ArrayValue` the returned output handle wraps, recorded on `_ControlEntry.result_array` and reused by `_wrap_entry_output`. The IR result scalars and the user-facing handle now share one array UUID.

Emit (`_map_controlled_u_results`): register each control result's root-array element `QubitAddress` (the key `emit_measure_vector` / `resolve_slice_chain` looks up), mirroring `_allocate_qubit_list`'s result-chain copy. The whole-array base key is intentionally not written: no resolver path reads it for these results, and writing it would add a base key to the input array for array-element scalar controls.

Scalar `Qubit`, `VectorView`, and symbolic control paths are unchanged.

## Tests

Adds `TestControlledWholeVectorControlOutput`: cross-SDK (Qiskit / QuriParts / CUDA-Q) coverage over `num_controls` in {2, 3} and seeds, exercising the sampling and expectation-value primitives independently against the always-working scalar-control form. The sampling test prepares a random non-zero basis state on the whole-Vector control prefix and asserts the returned control vector measures it back; the expval test pins the whole-Vector form against the individual-scalar-control form.

## Verification

`/local-review` clean (ruff check, ruff format, zuban all pass). Full `tests/circuit` suite green (5915 passed); controlled / qpe / qft suites green. QuriParts skips the C³X (`num_controls=3`) legs via the existing `EmitError` pattern — an orthogonal pre-existing backend gap; the Toffoli (`num_controls=2`) legs pass there. CUDA-Q legs are guarded and run in CI.


ksk-jij:
# Branch summary: whole-`Vector` control-output qubit-mapping fix

This branch has merged the latest `origin/main`; its **net difference against `main` is a single fix to the whole-`Vector` control output of `qmc.control`** (`qamomile/circuit/frontend/operation/control.py` / `qamomile/circuit/transpiler/passes/emit_support/controlled_emission.py` plus the corresponding tests). The CUDA-Q isolation commits that previously sat on this branch have already been merged into `main`, so they are no longer branch-specific. The merge completed with no conflicts.

## 1. Problem overview

`qmc.control` wraps a quantum kernel (gate) into its controlled version. There are two ways to specify the control qubits: passing several individual scalar `Qubit` handles, or passing a whole length-`n` `Vector[Qubit]` (a qubit-array handle) allocated via `qmc.qubit_array(n)` as the control prefix.

`main` behavior: when a whole `Vector` is passed as the control, measuring the returned control vector (`c_out`) with `qm.measure(c_out)` or accessing an element with `c_out[i]` produces no measurement instruction at all, so the result is always all-zero bits. In Qiskit this surfaces as a "No counts" error or "classical register will remain all zeros". Passing the controls as individual scalar `Qubit` handles works fine. So the same logical circuit breaks purely depending on how the controls are passed — a bug whose cause is hard to see from the user's side.

This branch's behavior: measuring or element-accessing the return value of a whole-`Vector` control now resolves correctly to the same physical qubits as the input.

Root cause: in the IR, a controlled-U with a concrete number of controls is represented by the `ConcreteControlledU` op, whose control region uses a "one scalar value per control qubit" layout (`ConcreteControlledU` in `qamomile/circuit/ir/operation/gate.py`). The frontend therefore expands a whole-`Vector` control into per-element scalar operands / results. The handle returned to the user, however, was re-aggregated separately as a new `ArrayValue` produced by `next_version()` (the operation that creates the next SSA version of the same logical value). The UUID of this new array is referenced nowhere in the IR op, so during emit (backend code generation) and resource allocation (`qubit_map`: the map from each qubit address to a physical qubit index), the per-element addresses `QubitAddress(array_uuid, i)` of that array were never written into `qubit_map`. `QubitAddress` is the qubit identity key meaning "which element of which array". Because they were unwritten, a downstream `measure` could not resolve those array elements and dropped without emitting any measurement instruction.

The reason passing individual scalar `Qubit` controls worked is that there the returned `Qubit` handle directly wraps the IR result scalar itself (the scalar value in `op.results`), and emit had mapped that scalar's UUID to a physical qubit.

## 2. Frontend (user-facing code level) changes

The way users write the qkernel decorator and the API signatures themselves are unchanged. The only change is the behavior: the return value of a whole-`Vector` control can now be used correctly.

The following kernel returned empty / all-zero measurements for `c_out` on `main`, but on this branch it measures the prepared state (`|11>`) as expected.

```python

@qmc.qkernel
def _x_gate(q: qmc.Qubit) -> qmc.Qubit:
    return qmc.x(q)

@qmc.qkernel
def kernel() -> qmc.Vector[qmc.Bit]:
    ctrl = qmc.qubit_array(2, "ctrl")   # use a length-2 Vector[Qubit] as the whole control
    t = qmc.qubit(name="t")
    ctrl[0] = qmc.x(ctrl[0])
    ctrl[1] = qmc.x(ctrl[1])            # prepare the control register to |11>
    cg = qmc.control(_x_gate, num_controls=2)
    c_out, t_out = cg(ctrl, t)          # pass the whole Vector as the control prefix
    return qmc.measure(c_out)           # measure the returned control vector
# main:   no measurement emitted, all-zero bits
# branch: every shot returns (1, 1)
```

## 3. Backend (IR / pipeline) changes

The fix has two parts: the frontend makes the returned control vector share its array identity with the IR result scalars, and emit registers that array's element addresses in `qubit_map`.

Frontend (`qamomile/circuit/frontend/operation/control.py`):

- Added a `result_array` field to `_ControlEntry` (the per-control bookkeeping dataclass) (`qamomile/circuit/frontend/operation/control.py:348`). For a whole-`Vector` / `VectorView` (a slice handle pointing at part of an array) control, it records the array value that the returned handle wraps.
- Added `_build_control_results` (`qamomile/circuit/frontend/operation/control.py:1028`). It re-parents the per-element result scalars of a whole-`Vector` / `VectorView` control onto the same new-version `ArrayValue` (`result_array`) that the user-facing handle wraps (`parent_array` is the pointer indicating "which array this value is an element of"). Concretely the result scalars are built via `dataclasses.replace(scalar.next_version(), parent_array=result_array)`. A scalar `Qubit` control keeps the previous behavior (`next_version()` only, no array bookkeeping).
- In `_wrap_entry_output` (`qamomile/circuit/frontend/operation/control.py:1262`), added a top-priority branch that reuses `entry.result_array` as the returned handle's array value when it is set (`qamomile/circuit/frontend/operation/control.py:1328`). This makes the "array UUID of the handle returned to the user" equal to the "array UUID that the IR result scalars' `parent_array` points at".

Emit (`qamomile/circuit/transpiler/passes/emit_support/controlled_emission.py`):

- In `_map_controlled_u_results` (`qamomile/circuit/transpiler/passes/emit_support/controlled_emission.py:2061`), for each control result scalar, in addition to mapping that scalar's UUID to a physical qubit, it now registers the result scalar's root-array element address (the `(array_uuid, index)` obtained by walking `parent_array` / the slice chain via `resolve_root_qubit_address`) in `qubit_map` (`qamomile/circuit/transpiler/passes/emit_support/controlled_emission.py:2122`). The vector-measurement emit computes the root array via `resolve_slice_chain` and looks up its element key `(array_uuid, start + step * i)`, so writing this element address lets the measurement / element access of the returned control vector resolve. This mirrors the result-chain copy in the resource allocator's `_allocate_qubit_list`. A standalone scalar `Qubit` control from `qmc.qubit()` has no `parent_array`, so `resolve_root_qubit_address` returns `None` and it is skipped. An array-element scalar `Qubit` control such as `ctrl[0]` does have a `parent_array`, but the resolved root address is the input array's already-registered address, so the `not in qubit_map` guard makes it a harmless no-op (neither differs in behavior from the whole-`Vector` control this fix targets). The whole-array base key (no element index) is deliberately not written here: no resolver path reads the base key for these results, and writing it would add a spurious base key to the input array for array-element scalar controls.

`resolve_root_qubit_address` is a shared helper (`qamomile/circuit/ir/value.py`) that composes the element / slice chain and returns the root array's `(uuid, index)`, resolving to the address registered when the root array was allocated. So for a whole-`Vector` (root array) the new array's element addresses are newly registered, while for a `VectorView` the chain resolves to the existing root address (the latter was already resolvable via the slice chain, hence unchanged).

## 4. Alternatives considered and why this approach was chosen

There were three design options for getting the returned control vector's element addresses into `qubit_map`.

- Option A (chosen): in the frontend, point the result scalars' `parent_array` at "the array the returned handle wraps", and let emit / resource allocation write that array's element addresses into `qubit_map` via the existing root-chain registration mechanism. This requires no change to `ConcreteControlledU`'s "controls are per-element scalars" operand / result contract, and the registration reuses the existing pattern from target-array results and `_allocate_qubit_list`. The returned array stays an ordinary root `ArrayValue`, so the handle semantics are unchanged.
- Option B (rejected): build the returned array as a "slice view of the input root array" and offload everything to slice-chain resolution (zero emit changes). But this would put a whole-`Vector` handle (a whole array) into a state where it internally holds a slice, risking violations of invariants such as the `Vector` vs `VectorView` distinction, element write-back, and borrow (linear-type borrow) tracking. The change is minimal, but its side effects are hard to fully predict, so it was dropped.
- Option C (rejected): change `ConcreteControlledU`'s result layout itself so the control region is a single `ArrayValue`. This would broadly break the many places that assume "control results are per-element scalars" — the `signature` property, resource allocation, resource estimation (estimator), `_map_controlled_u_results`, etc. — so it was judged excessive and risky for the target bug.

## 5. Known limitations

- When: running a whole-`Vector` control with `num_controls=3` or more (multi-control such as C³X) on some backends like QuriParts. Why: those backends' controlled-U fallback decomposition does not support multi-control gates and raises `EmitError`. This is a pre-existing backend gap orthogonal to this fix. The regression test (`TestControlledWholeVectorControlOutput`) is parametrized over every supported SDK (Qiskit / QuriParts / CUDA-Q) via `_BUILTIN_BACKENDS` and exercises both the sampling and the expectation-value paths (not just transpile but through `sample()` / `run()`). The minimal reproduction of the target bug, `num_controls=2`, is required to transpile + execute on every SDK (an `EmitError` there is re-raised, not skipped, so it fails), guaranteeing the fix runs end-to-end on every backend. Only this orthogonal gap — the `num_controls>=3` multi-control leg — is skipped (Qiskit supports it). Future fix: implementing multi-control decomposition on each backend emitter is separate work.

PR #478, opened 2026-06-22
Clarify CUDA-Q static-only expectation-value behavior

ksk-jij:
## Summary

- Expectation-value estimation on the CUDA-Q backend goes through `cudaq.observe`, which only accepts STATIC kernels. RUNNABLE artifacts (mid-circuit measurement or measurement-dependent control flow such as `if bit:` / `while bit:`) are executed with `cudaq.run` and cannot be consumed by `cudaq.observe`.
- `CudaqExecutor.estimate` previously raised a terse `TypeError` that told users to "use sample() or run() instead", neither of which returns an expectation value. This rewrites the message to name the cause (RUNNABLE mode is emitted for `cudaq.run`, which `cudaq.observe` cannot consume) and the remedy (express the state preparation as a static kernel; expectation values are unavailable for measurement-conditioned circuits on CUDA-Q).
- Documents the static-only contract on the `CudaqExecutor` class docstring and the `estimate` method docstring, recording that this is a CUDA-Q API limitation rather than a Qamomile gap, so the `TypeError` path is intentional.
- Investigation outcome: a single Qamomile kernel cannot be both RUNNABLE and carry a terminal `qmc.expval` (the affine model rejects reusing a measured qubit), so the `ExecutableProgram.run` expval path never feeds a RUNNABLE artifact into `estimate`; the guard is reachable only when `estimate` is called directly on a RUNNABLE artifact.

No behavior change to the static path or to the RUNNABLE/STATIC classification — only the error message wording, docstrings, and test coverage.

## Test plan

- New `TestExpvalStaticOnlyCudaq` in `tests/transpiler/backends/test_cudaq_frontend.py`:
  - Static estimate path: `<Z> = cos(theta)` parametrized over seeds and boundary angles (`0`, `pi/2`, `pi`, `2*pi`) plus a random angle per seed, and a Bell-state `a*ZZ + b*X0` -> `a` reference with randomized coefficients.
  - RUNNABLE estimate path raises a `TypeError` whose message names both the cause (`cudaq.run` vs `cudaq.observe`) and the remedy (static kernel).
  - A kernel cannot be both RUNNABLE and carry a terminal expectation value (affine rejection, matched on the consumed-qubit message).
- `uv run pytest -m cudaq tests/transpiler/backends/test_cudaq.py tests/transpiler/backends/test_cudaq_frontend.py` -> 388 passed.
- `uv run pytest tests/` -> 8843 passed, 4 skipped.
- `uv run ruff check qamomile/ tests/`, `uv run ruff format --check qamomile/ tests/`, and `uv run zuban check qamomile/` all clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DtWeu7epEQtqVMDGEPaVBn)_

### DIFF
--- a/qamomile/cudaq/transpiler.py
+++ b/qamomile/cudaq/transpiler.py
@@ -2252,13 +2252,25 @@ class CudaqEmitPass(StandardEmitPass[CudaqKernelArtifact]):
 class CudaqExecutor(QuantumExecutor[CudaqKernelArtifact]):
     """CUDA-Q quantum executor.
 
-    Supports sampling via ``cudaq.sample`` and expectation value estimation
-    via ``cudaq.observe``.  Dispatches to the appropriate CUDA-Q runtime
-    API based on the artifact's ``execution_mode``.
+    Supports sampling via ``cudaq.sample`` / ``cudaq.run`` and expectation
+    value estimation via ``cudaq.observe``.  Dispatches to the appropriate
+    CUDA-Q runtime API based on the artifact's ``execution_mode``:
+
+    - ``STATIC`` artifacts (no mid-circuit measurement or runtime control
+      flow) support both sampling (``cudaq.sample``) and expectation-value
+      estimation (``cudaq.observe``).
+    - ``RUNNABLE`` artifacts (mid-circuit measurement or measurement-dependent
+      control flow such as ``if bit:`` / ``while bit:``) support sampling only,
+      via ``cudaq.run``.
+
+    Expectation-value estimation is therefore **static-only**: calling
+    :meth:`estimate` on a ``RUNNABLE`` artifact raises ``TypeError`` because
+    ``cudaq.observe`` cannot consume a kernel that requires ``cudaq.run``.
+    See :meth:`estimate` for the full rationale.
 
     Args:
-        target: CUDA-Q target name (e.g., ``"qpp-cpu"``). If None, uses
-            the default CUDA-Q target.
+        target (str | None): CUDA-Q target name (e.g., ``"qpp-cpu"``). If
+            None, uses the default CUDA-Q target.
     """
 
     def __init__(self, target: str | None = None):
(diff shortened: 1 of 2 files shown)


ksk-jij:
## Problem overview

Expectation-value estimation on the CUDA-Q backend (`CudaqExecutor.estimate`) calls CUDA-Q's `cudaq.observe`, which only accepts STATIC kernels (deterministic state-preparation kernels with no mid-circuit measurement or measurement-dependent control flow). A kernel containing mid-circuit measurement or `if bit:` / `while bit:` measurement-dependent control flow is emitted by Qamomile as a RUNNABLE-mode artifact and executed via `cudaq.run`, so it cannot be consumed by `cudaq.observe`. On `main`, calling `estimate` on a RUNNABLE artifact already raises a `TypeError`, but its message ("`cudaq.observe()` is not supported for runtime control flow circuits. Use sample() or run() instead.") states neither why it is impossible nor the fact that the suggested `sample()` / `run()` do not return expectation values, so it can mislead users. This branch rewrites that `TypeError` message to make both the cause (RUNNABLE mode is emitted for `cudaq.run` and cannot be consumed by `cudaq.observe`) and the remedy (express the state preparation as a static kernel without measurement-dependent control flow; expectation values cannot be computed for measurement-conditioned circuits on CUDA-Q) explicit, documents the static-only contract in the docstrings, and locks the behavior in with tests. This addresses the backlog acceptance criteria (document that it is static-only; test that static kernels can estimate expectation values and that RUNNABLE kernels fail with a clear `TypeError`; record whether the limitation is a CUDA-Q API limit or a Qamomile gap). The investigation conclusion is that this is a CUDA-Q API limitation, not a Qamomile gap: `cudaq.observe` evaluates the analytic expectation value of a deterministic state-preparation kernel and has no execution path for a kernel that requires the per-shot `cudaq.run` runtime. There is no safe alternative that preserves this method's "returns the exact expectation value" contract, so the `TypeError` path is kept intentionally.

## Frontend (user-facing code level) changes

There is no change to how users write `@qm.qkernel` code or to any API signature. What changes is the content of the `TypeError` raised when expectation-value estimation is requested on a RUNNABLE-kernel artifact (a kernel containing mid-circuit measurement or measurement-dependent branching): it now includes both the cause and the remedy. For example, the following code fails with a `TypeError` before and after this branch, but the message becomes specific.

```python
import qamomile.observable as qm_o
from qamomile.cudaq import CudaqTranspiler


@qmc.qkernel
def runnable() -> qmc.Bit:
    q = qmc.qubit("q")
    q = qmc.h(q)
    b = qmc.measure(q)  # mid-circuit measurement + measurement-dependent branch -> RUNNABLE mode
    if b:
        q2 = qmc.qubit("q2")
        q2 = qmc.x(q2)
    return b


transpiler = CudaqTranspiler()
exe = transpiler.transpile(runnable)
H = qm_o.Hamiltonian(num_qubits=1)
H += qm_o.Z(0)

# Expectation-value estimation is static-only. For a RUNNABLE artifact a TypeError is
# raised whose message names the cause (emitted for cudaq.run, cannot be consumed by
# cudaq.observe) and the remedy (write a static kernel).
transpiler.executor().estimate(exe.get_first_circuit(), H)
```

To obtain an expectation value, the new error message guides the user to express the state preparation as a static kernel (one that stays in STATIC mode) without measurement-dependent control flow.

## Backend (IR and overall) changes

The change is confined to a single CUDA-Q backend file (`qamomile/cudaq/transpiler.py`) and does not propagate to the IR, the transpiler pipeline, or other backends.

- The `CudaqExecutor` class docstring (`qamomile/cudaq/transpiler.py:2252`; the static-only note around `qamomile/cudaq/transpiler.py:2266`) now records the per-execution-mode table — STATIC artifacts support both sampling (`cudaq.sample`) and expectation-value estimation (`cudaq.observe`), while RUNNABLE artifacts (mid-circuit measurement or measurement-dependent control flow) support sampling only via `cudaq.run` — and states that expectation-value estimation is static-only. Here an "artifact" is the executable kernel representation emitted by the CUDA-Q backend (`CudaqKernelArtifact` / `BoundCudaqKernelArtifact`).
- The `CudaqExecutor.estimate` docstring (`qamomile/cudaq/transpiler.py:2411`) was expanded to Google style (typed `Args` / `Returns` / `Raises` / `Example`), describing why it is static-only, that this is a CUDA-Q API limitation, and why the `ExecutableProgram.run` expval path never reaches this guard (the affine constraint below).
- The `TypeError` raised after the RUNNABLE check in `estimate` (around `qamomile/cudaq/transpiler.py:2474`) was rewritten to include the symptom, the cause (RUNNABLE mode is emitted for `cudaq.run`, and the expectation-value primitive `cudaq.observe` only accepts static kernels), and the remedy (write a static kernel / expectation values cannot be computed for measurement-conditioned circuits). The decision logic itself (raise when `execution_mode` is RUNNABLE) is unchanged.
- Tests were added as `TestExpvalStaticOnlyCudaq` (`tests/transpiler/backends/test_cudaq_frontend.py:1278`). `test_static_kernel_estimate_matches_analytic` (line 1289) verifies `<Z> = cos(theta)` for an `Rx(theta)` state, parametrized over seeds and boundary angles (0, pi/2, pi, 2pi); `test_static_kernel_estimate_random_observable` (line 1316) verifies that for a Bell state the expectation of `a*ZZ + b*X0` equals `a` with randomized coefficients; `test_runnable_kernel_estimate_raises_clear_typeerror` (line 1343) verifies that `estimate` on a RUNNABLE artifact raises a `TypeError` containing the cause and remedy; `test_runnable_kernel_cannot_carry_terminal_expval` (line 1374) verifies that a single kernel cannot be both RUNNABLE and carry a terminal expval (the affine constraint rejects reusing a measured qubit). Here the "affine constraint" is Qamomile's "each quantum value may be used at most once" rule enforced at the `affine_validate` stage.

## Alternatives considered and rationale

The option of also providing expectation-value estimation for RUNNABLE kernels (for example, estimating the expectation of a diagonal observable from `cudaq.run` sample results) was rejected. `estimate` has a "returns the exact expectation value" contract, and a sample-derived estimate applies only to diagonal observables and carries shot noise, so it cannot satisfy that contract. The investigation also confirmed that `cudaq.observe` has no execution path for a kernel that requires `cudaq.run`, i.e. this is a CUDA-Q API limitation; therefore the chosen approach keeps the current policy of failing early and clearly with a `TypeError` while improving the message and documentation. The option of leaving the error message unchanged and explaining the limitation only in a docs page was also considered, but since the error message is the first cue for users who call `estimate` directly, improving the error message itself was made the primary change.

## Known limitations

On the CUDA-Q backend, expectation-value estimation for circuits with measurement-dependent control flow (RUNNABLE) remains unsupported (When: calling `estimate` directly on a RUNNABLE artifact / Why: the CUDA-Q API limitation that `cudaq.observe` cannot consume a kernel that goes through `cudaq.run` / Future fix: revisitable if CUDA-Q provides an observation API for RUNNABLE kernels, or if sample-based estimation is added explicitly as a separate method). In current Qamomile, the affine constraint makes it structurally impossible for a single kernel to be both RUNNABLE and carry a terminal expval, so the `ExecutableProgram.run` expval path never reaches this guard; the guard is reached only when `estimate` is called directly on a RUNNABLE artifact.

---

PR #533, opened 2026-06-24
docs(algorithm): add QFT tutorial documentation

jij-yuichiro-nakano:
## Summary

Adds bilingual documentation for the Quantum Fourier Transform (QFT) as an algorithm tutorial.

## What Was Added

- Added QFT tutorials under `docs/en/algorithm/qft.py` and `docs/ja/algorithm/qft.py`, with synced notebooks.
- Introduced the DFT background, QFT algorithm, controlled phase rotations, and the QFT circuit for `n = 4`.
- Added a four-qubit frequency-estimation example using `qamomile.circuit.stdlib.qft`.
- Added execution-result visualization and resource-scaling discussion using Qamomile's resource estimation feature.

## Test Plan

- `./docs/build.sh page-build docs/en/algorithm/qft.py docs/ja/algorithm/qft.py`
- `git diff --check`


docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,331 @@
...
+# フーリエ変換は、データを周波数の成分として書き直します。どの周波数がどれだけ含まれているかを調べる方法です。有限長のベクトルに対してよく使うのが**離散フーリエ変換**（Discrete Fourier Transform、DFT）です。
+#
+# ベクトル$x = (x_0, x_1, \ldots, x_{N-1})$に対して、このノートブックでは次の正規化を使います。
+#
+# $$
+# y_k = \frac{1}{\sqrt{N}}\sum_{j=0}^{N-1} x_j e^{2\pi i jk / N},
+# \qquad k = 0, 1, \ldots, N-1.
+# $$
+#
+# 出力の添字$k$は周波数を表します。位相因子$e^{2\pi i jk / N}$により、入力の各位置が異なる位相で足し合わされます。
```
Jacomichi: > 出力の添字$k$は周波数

Typically, $2\pi k/N$ is considered the definition of frequency.

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,331 @@
...
+
+# %% [markdown]
+# ## アルゴリズム
+#
+# QFTは、同じ変換を量子状態の振幅に適用します。$N = 2^n$のとき、基底状態$\lvert x\rangle$への作用は次の通りです。
```
Jacomichi: > 基底状態$\lvert x\rangle$

I don't think there's any reason it has to be the ground state.

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,331 @@
...
+#
+# $$
+# \mathrm{QFT}_N \lvert x \rangle =
+# \frac{1}{\sqrt{N}}\sum_{k=0}^{N-1} e^{2\pi i xk/N}\lvert k \rangle.
+# $$
+#
+# 古典のDFTでは、出力ベクトル全体を得られます。一方、QFTは量子状態の振幅を変換します。QFTの直後に測定しても、得られるのはサンプルだけです。ただし、位相推定のように、後続の量子演算では変換後の位相情報を利用できます。
```
Jacomichi: > QFTは量子状態の振幅を変換します。

QFTは量子状態の振幅を変換し，量子状態を返します。

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,331 @@
...
+#
+# ### ステップ1：対象量子ビットを1つ選ぶ
+#
+# レジスタを対象量子ビットごとに処理します。最後の量子ビットにアダマールゲートを適用すると、QFTの出力に現れる最初の因子が得られます。
+#
+# $$
+# \lvert x_n\rangle
+# \xrightarrow{H}
+# \frac{1}{\sqrt{2}}
+# \left(\lvert 0\rangle + e^{2\pi i[0.x_n]}\lvert 1\rangle\right).
```
Jacomichi, line 92: Depending on the reader's level, this step might be too much of a leap for beginners. It would be more reader-friendly to break it down into cases for $x_n$

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,331 @@
...
+# ## Qamomileでの実装：`qft`関数
+#
+# `qamomile.circuit.stdlib.qft.qft`関数は、上で説明した回路を適用します。アダマールゲート、制御付き位相回転、最後のスワップをまとめて扱えます。
+#
+# ### 問題例
+#
+# $N=16$個のサンプルを使うので、4量子ビットで表せます。この例では$n=4$、$N=16$、$\omega = e^{2\pi i/16}$であり、次のように書けます。
+#
+# $$
+# F_{16}\lvert x \rangle =
```
Jacomichi: > F_{16}

Does this mean $\mathrm{QFT}_{16}$ ?

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,331 @@
...
+# 出力の添字$k$は周波数を表します。位相因子$e^{2\pi i jk / N}$により、入力の各位置が異なる位相で足し合わされます。
+
+# %% [markdown]
+# ## アルゴリズム
+#
+# QFTは、同じ変換を量子状態の振幅に適用します。$N = 2^n$のとき、基底状態$\lvert x\rangle$への作用は次の通りです。
+#
+# $$
+# \mathrm{QFT}_N \lvert x \rangle =
+# \frac{1}{\sqrt{N}}\sum_{k=0}^{N-1} e^{2\pi i xk/N}\lvert k \rangle.
```
Jacomichi, line 63: This expression seems to be for the case where $x \in \mathbb{Z}$. For a general state $|x\rangle$, shouldn't we write it as $|x\rangle = \sum_j x_j |j\rangle$ and apply the QFT to these $|j\rangle$ states?

docs/ja/algorithm/qft.py
Jacomichi: I think it would be better to reorganize the order of the explanation here. Rather than explicitly showing the QFT on $|x\rangle$ first, it might serve as a better concrete example to introduce $|\psi_f\rangle$ first and then consider the action of the QFT on it."

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,331 @@
...
+# %%
+@qmc.qkernel
+def qft_frequency_estimator() -> qmc.Vector[qmc.Bit]:
+    q = qmc.qubit_array(num_qubits, name="q")
+
+    # すべてのサンプルインデックスの一様重ね合わせを作ります。
+    q = qmc.h(q)
+
+    # 位相パターン exp(-2πi f j / N) を振幅に埋め込みます。
+    q[0] = qmc.p(q[0], -2 * math.pi * frequency / dimension)
```
Jacomichi: Is there any reason why `frequency` and `dimension` must be global variables?

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,331 @@
...
+# - $\frac{n(n - 1)}{2}$個の制御付き位相回転
+# - $\left\lfloor n / 2 \right\rfloor$個のスワップ
+#
+# したがって総ゲート数は$n + \frac{n(n - 1)}{2} + \left\lfloor n / 2 \right\rfloor$であり、$O(n^2)$で増えます。
+#
+# それでは、Qamomileのリソース推定機能を使って確認してみましょう。
+# Qamomileでは、量子ビット数がバインドされていないシンボリックな量子カーネルに対して`estimate_resources()`を使うことで、量子ビット数に関する式としてスケーリングを確認することができます。
+
+# %%
+@qmc.qkernel
```
Jacomichi: Is it not possible to perform this estimation using Qamomile's `qft` resource estimation? If it is possible, I think we should add a note at the end to make it clear that users don't need to implement this themselves. If it's not possible, we should probably make it supported.

Jacomichi (changes requested):
Thanks for PR!
I left some comments

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thanks for catching this. Fixed in `1a7d61d9`: I replaced this with “the computational basis state corresponding to an integer `x`” so it no longer reads as a ground-state statement.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thanks for the helpful suggestion. Fixed in `1a7d61d9`: I reorganized the example to introduce `|ψ_f⟩` first, then describe how applying QFT to that state concentrates the output on `|f⟩`.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thanks for pointing this out. Fixed in `1a7d61d9`: the text now clarifies that the displayed formula is for the computational basis state corresponding to an integer `x`, and adds the general-state explanation by linearity over the computational basis states.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thanks. Fixed in `1a7d61d9`: the text now says that QFT transforms the amplitudes and returns the transformed quantum state.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thanks for the suggestion. Fixed in `1a7d61d9`: I added the explicit `x_n=0` and `x_n=1` cases before combining them into the compact phase-factor expression.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thanks. Fixed in `1a7d61d9`: the resource-estimation section now uses the standard-library `qft()` directly inside a small resource-estimation kernel, so users do not need to reimplement the QFT gate structure themselves. I also added the concrete four-qubit gate count and a scaling plot over 3 to 9 qubits.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thanks for pointing this out. Fixed in `1a7d61d9`: the DFT explanation now treats `k` as the frequency-component index and states that the corresponding angular frequency is `2πk/N`.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thanks for checking. There is no technical need for `frequency` and `dimension` to be global variables. I kept them as notebook-level constants here because it makes the intent of `qft_frequency_estimator()` easier to read: the qkernel body directly shows the phase pattern being encoded using the same example parameters introduced just above.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thanks. Fixed in `1a7d61d9`: I replaced the notation with `QFT_{n=4}` to make the four-qubit QFT example explicit.

jij-yuichiro-nakano:
Thank you for the comments! I have addressed the points you raised. Please take another look.

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,378 @@
...
+# QFTは、DFTと同じ変換を量子状態の振幅に適用します。$N = 2^n$のとき、整数$x$に対応する計算基底状態$\lvert x\rangle$への作用は次の通りです。
+#
+# $$
+# \mathrm{QFT}_N \lvert x \rangle =
+# \frac{1}{\sqrt{N}}\sum_{k=0}^{N-1} e^{2\pi i xk/N}\lvert k \rangle.
+# $$
+#
+# 一般の量子状態$\lvert\psi\rangle = \sum_{j=0}^{N-1} a_j\lvert j\rangle$に対しては、QFTは線形性により各計算基底状態$\lvert j\rangle$への作用を重ね合わせた状態を返します。
+#
+# 古典のDFTでは、出力ベクトル全体を得られます。一方、QFTは量子状態の振幅を変換し、変換後の量子状態を返します。QFTの直後に測定しても、得られるのはサンプルだけです。ただし、位相推定のように、後続の量子演算では変換後の位相情報を利用できます。
```
Jacomichi: > "QFTの直後に測定しても、得られるのはサンプルだけです。"

I think it would be better to state explicitly what we will obtain, rather than just using the word "sample."

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,378 @@
...
+from qamomile.circuit.stdlib.qft import qft
+from qamomile.qiskit import QiskitTranspiler
+from qiskit_aer import AerSimulator
+
+transpiler = QiskitTranspiler()
+
+# %% [markdown]
+# ## 背景: フーリエ変換
+#
+# フーリエ変換は、データを周波数の成分として書き直します。どの周波数がどれだけ含まれているかを調べる方法です。有限長のベクトルに対してよく使うのが**離散フーリエ変換**（Discrete Fourier Transform、DFT）です。
```
Jacomichi: > フーリエ変換は、データを周波数の成分として書き直します。どの周波数がどれだけ含まれているかを調べる方法です。

Both sentences have the same subject, so I think we should combine them into one sentence.

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,378 @@
...
+# $$
+# \mathrm{QFT}_N \lvert x \rangle =
+# \frac{1}{\sqrt{N}}\sum_{k=0}^{N-1} e^{2\pi i xk/N}\lvert k \rangle.
+# $$
+#
+# 一般の量子状態$\lvert\psi\rangle = \sum_{j=0}^{N-1} a_j\lvert j\rangle$に対しては、QFTは線形性により各計算基底状態$\lvert j\rangle$への作用を重ね合わせた状態を返します。
+#
+# 古典のDFTでは、出力ベクトル全体を得られます。一方、QFTは量子状態の振幅を変換し、変換後の量子状態を返します。QFTの直後に測定しても、得られるのはサンプルだけです。ただし、位相推定のように、後続の量子演算では変換後の位相情報を利用できます。
+#
+# 標準的なQFT回路は、アダマールゲート、制御付き位相回転、最後のスワップで構成されます。$n$量子ビットのレジスタでは、密な$2^n \times 2^n$行列を直接使うのではなく、$O(n^2)$個のゲートで実装できます。位相を表すために、次の2進小数表記を使います。
```
Jacomichi: > $n$量子ビットのレジスタでは、密な$2^n \times 2^n$行列を直接使うのではなく、$O(n^2)$個のゲートで実装できます。

This sentence is a bit confusing. "密な$2^n \times 2^n$行列を直接使う" seems to refer to data representation, whereas "$O(n^2)$個のゲートで実装" sounds like an explanation of the implementation itself. Therefore, it is unclear whether a fair comparison is being made.

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,378 @@
...
+#
+# つまり、$e^{-2\pi i f j/N}$の値を、正規化係数$1/\sqrt{N}$付きで$\lvert \psi_f\rangle$の振幅に埋め込んでいます。この状態にQFTを適用すると、各計算基底状態$\lvert j\rangle$からの寄与が足し合わされ、周波数インデックス$f$だけが残ります。
+#
+# $$
+# \mathrm{QFT}_{16}\lvert \psi_f \rangle =
+# \frac{1}{16}\sum_{j=0}^{15}\sum_{k=0}^{15}\omega^{j(k-f)}\lvert k\rangle
+# = \lvert f\rangle.
+# $$
+#
+# したがって、$f=5$のとき、出力は周波数インデックス$k=5$に集中するはずです。
```
Jacomichi: Since this section explains what happens in quantum states, if you are going to include code for classical FFT, it would be better to add one sentence. Otherwise, readers will likely expect QFT results come here.

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,378 @@
...
+
+assert estimate_4.qubits == 4
+assert estimate_4.gates.total == 12
+assert estimate_4.gates.single_qubit == 4
+assert estimate_4.gates.two_qubit == 8
+assert estimate_4.gates.rotation_gates == 6
+assert estimate_4.gates.clifford_gates == 6
+
+# %% [markdown]
+# 長さ$N$のベクトルを古典DFTで直接計算すると、$O(N^2)$回の演算が必要です。高速フーリエ変換（FFT）を使うと、これを$O(N\log N)$まで減らせます。一方、$N = 2^n$と書くと、厳密QFTは$O(n^2)=O((\log N)^2)$個のゲートで実装できます。したがって、直接計算する古典DFTと比べると、$N$に対して指数的に少ないゲートで同じ変換を状態に適用できます。ただし、測定だけで$N$個すべてのフーリエ係数を読み出せるわけではありません。QFTの利点は、変換後の振幅を後続の量子演算でそのまま使える場合に現れます。
```
Jacomichi, line 400: I think the discussion in this paragraph is good, but it would be better to include a mention of the next cell as well.

docs/ja/algorithm/qft.py
```diff
@@ -0,0 +1,378 @@
...
+    marker="o",
+    color="#2696EB",
+    label="QFT gate count",
+)
+ax.plot(
+    qft_qubit_counts,
+    quadratic_reference,
+    linestyle="--",
+    color="#FF6B6B",
+    label=r"theory: QFT $O(n^2)$",
```
Jacomichi: Since you have already explained the gate count in more detail, why not use that? 
I am a bit concerned about the discrepancy between Qamomile's resource estimation and the theory.

Jacomichi (commented):
Thanks to update!
I left a few minor comments.

jij-yuichiro-nakano:
@Jacomichi  Thanks for the comments! I've made the changes based on your review. Please take a look!

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thank you for the suggestion. Updated in 48adfe15. I added a sentence explaining that the next cell performs a classical DFT with NumPy's np.fft.ifft to verify the expected result before implementing the quantum kernel.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thank you for pointing this out. Updated in 48adfe15. The theoretical QFT curve now uses the exact gate-count formula n + n(n - 1)/2 + floor(n/2), and the same values are used to verify Qamomile's resource estimates.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thank you for pointing this out. Updated in 48adfe15. I removed the comparison with a dense matrix and now state only that the exact QFT circuit uses O(n^2) gates.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thank you for the suggestion. Updated in 48adfe15. I combined the two sentences into a single, more concise description of the Fourier transform.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thank you for the suggestion. Updated in 48adfe15. The paragraph now states explicitly that measuring immediately after QFT yields a computational-basis outcome sampled from the transformed probability distribution, while the transformed phase information can be used directly when QFT is used as a subroutine.

docs/ja/algorithm/qft.py
jij-yuichiro-nakano: Thank you for the suggestion. Updated in 48adfe15. I added a paragraph introducing the next cell and explaining that the plot compares the qft gate count from .estimate_resources() with the theoretical QFT and DFT scaling.

Jacomichi (approved):
LGTM

PR #536, opened 2026-06-25
docs(algorithm): add QPE tutorial documentation

jij-yuichiro-nakano:
## Summary

Adds bilingual documentation for the Quantum Phase Estimation (QPE) as an algorithm tutorial.

## What Was Added

- Added QPE tutorials under `docs/en/algorithm/qpe.py` and `docs/ja/algorithm/qpe.py`, with synced notebooks.
- Introduced the phase-kickback background, QFT-based decoding, and the QPE algorithm steps with state-transition formulas.
- Added a 4x4 diagonal-unitary example using `qamomile.circuit.qpe`, including `QFixed` phase decoding and simulator execution.
- Added precision comparison from 3 to 9 counting qubits and resource-scaling discussion using Qamomile's symbolic resource estimation feature.
- Added QPE entries to the English and Japanese algorithm index pages.

## Test Plan

- `./docs/build.sh page-build docs/en/algorithm/qpe.py docs/ja/algorithm/qpe.py`
- `git diff --check`
- `uv run pytest tests/docs/test_tutorials.py -m docs -k 'algorithm/qpe' -v`


docs/en/algorithm/qpe.py
```diff
@@ -0,0 +1,638 @@
...
+# approximation.
+
+# %%
+# Set the counting-register sizes, sampling settings, and target eigenstate.
+COUNTING_BIT_OPTIONS = tuple(range(3, 10))
+DRAW_COUNTING_BITS = 3
+EXAMPLE_COUNTING_BITS = 3
+SHOTS = 4096
+SAMPLER_SEED = 321
+TARGET_BASIS = 1  # |01>
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Fixed in `1dbb38b9`: added `QAMOMILE_DOCS_TEST` handling to the English QPE page so docs tests use 3 to 5 counting qubits and 512 shots, while normal notebook execution keeps the full 3 to 9 range and 4096 shots.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,546 @@
...
+# この行列では、すべての計算基底状態が固有状態です。対象状態$|01\rangle$を準備し、推定したい位相$\theta_{01} / 2\pi$を$0.6$に設定します。$0.6$は2進格子上で正確には表現できないため、カウント用量子ビットを増やす効果が見えやすくなります。
+
+# %%
+# カウント用レジスタのサイズ、サンプリング設定、対象固有状態を決めます。
+COUNTING_BIT_OPTIONS = tuple(range(3, 10))
+DRAW_COUNTING_BITS = 3
+EXAMPLE_COUNTING_BITS = 3
+SHOTS = 4096
+SAMPLER_SEED = 321
+TARGET_BASIS = 1  # |01>
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Fixed in `1dbb38b9`: added the same `QAMOMILE_DOCS_TEST` handling to the Japanese QPE page so CI runs a smaller workload while rendered docs keep the full outputs.

docs/en/algorithm/index.md
```diff
@@ -12,6 +12,7 @@ Concrete quantum algorithm examples built with Qamomile.
 - [QAOA for Graph Partitioning](qaoa_graph_partition) — End-to-end optimization example using OMMX, JijModeling, and `QAOAConverter`
 - [VQE for the Hydrogen Molecule](vqe_for_hydrogen) — Build a molecular Hamiltonian with OpenFermion and find the ground state energy via VQE
 - [Hamiltonian Simulation with Suzuki–Trotter](hamiltonian_simulation) — Trotter–Suzuki product formulas on the Rabi model with empirical convergence orders
+- [Introduction to the Quantum Phase Estimation (QPE)](./qpe) — Apply the built-in `qpe` helper to a 4x4 unitary, sample the decoded phase, and compare precision as the counting register grows
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Fixed in `1dbb38b9`: changed the English algorithm index link from `./qpe` to the bare `qpe` target to match the surrounding entries.

docs/ja/algorithm/index.md
```diff
@@ -12,6 +12,7 @@ Qamomileで実装した具体的な量子アルゴリズム例です。
 - [QAOAによるグラフ分割](qaoa_graph_partition) — OMMX・JijModeling・`QAOAConverter`を使ったend-to-endの最適化例
 - [水素分子のためのVQE](vqe_for_hydrogen) — OpenFermionで分子ハミルトニアンを構築し、VQEで基底状態エネルギーを求める
 - [Suzuki–Trotterによるハミルトニアンシミュレーション](hamiltonian_simulation) — RabiモデルでのTrotter–Suzuki積公式と収束次数の実験
+- [量子位相推定（QPE）入門](./qpe) — 4x4ユニタリに組み込みの`qpe`ヘルパーを適用し、復号された位相をサンプリングして、カウント用レジスタを大きくしたときの精度を比較する
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Fixed in `1dbb38b9`: changed the Japanese algorithm index link from `./qpe` to the bare `qpe` target to match the surrounding entries.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,548 @@
...
+# ---
+
+# %% [markdown]
+# ---
+# tags: [algorithm, primitive, resource-estimation]
+# ---
+#
+# # 量子位相推定（QPE）入門
+#
+# 量子位相推定（QPE）は、$U|\psi\rangle = e^{2\pi i \phi}|\psi\rangle$を満たすユニタリ$U$と固有状態$|\psi\rangle$から、固有位相$\phi$を推定するアルゴリズムです。位数発見など、ユニタリの固有値に埋め込まれた位相を使うアルゴリズムで中心的なプリミティブとして使われます{cite:p}`10.48550/arXiv.quant-ph/9511026,10.1098/rspa.1998.0164`。
```
Jacomichi: In QPE, while inputting an eigenstate makes things simpler, I don't think the input necessarily has to be an eigenstate.

Jacomichi: > 量子位相推定（QPE）

it is better to write such as (Quantum Phase Estimation; QPE)

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,548 @@
...
+# =
+# O\!\left(\frac{1}{\epsilon}\right)
+# $$
+#
+# です。つまり、カウント用量子ビット数は$1/\epsilon$に対して対数的ですが、繰り返しによる制御ユニタリーゲートの回数は$O(1/\epsilon)$で増えます{cite:p}`10.1017/CBO9780511976667`。
+#
+# 最後に、この見積もりは実装に依存します。一般には
+#
+# $$
+# C_{\mathrm{QPE}}(m)
```
Jacomichi: What does this $C$ mean? Number of gates in QPE?

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,548 @@
...
+# + O(m^2)
+# $$
+#
+# と考えられます。
+#
+# :::{note}
+# $m$個の量子ビットに対する逆QFTは、$O(m^2)$個のゲートに分解できます。詳しくはQFTのチュートリアルを参照してください。
+# :::
+#
+# 上の$O(1/\epsilon)$という見積もりは、$U^{2^k}$を$U$の繰り返しで実装する場合のものです。問題の構造を使って$\mathrm{controlled}\text{-}U^{2^k}$をより直接的に実装できる場合は、その実装コストを使ってリソースを見積もる必要があります。
```
Jacomichi: I think we should expand the explanation a bit here. This is because the example of $U$ in the first half of this tutorial is a specific case where $U^k$ can be easily calculated.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,548 @@
...
+# %% [markdown]
+# ---
+# tags: [algorithm, primitive, resource-estimation]
+# ---
+#
+# # 量子位相推定（QPE）入門
+#
+# 量子位相推定（QPE）は、$U|\psi\rangle = e^{2\pi i \phi}|\psi\rangle$を満たすユニタリ$U$と固有状態$|\psi\rangle$から、固有位相$\phi$を推定するアルゴリズムです。位数発見など、ユニタリの固有値に埋め込まれた位相を使うアルゴリズムで中心的なプリミティブとして使われます{cite:p}`10.48550/arXiv.quant-ph/9511026,10.1098/rspa.1998.0164`。
+#
+# このノートブックでは、組み込みの`qpe`ヘルパーを4x4ユニタリに適用します。既知の固有状態を準備し、ローカルのQiskitシミュレータで回路を実行して、復号された位相を目標位相$0.6$と比較します。さらに、カウント用量子ビットを増やすと精度が上がることを確認し、シンボリックなリソース推定を使って、必要な精度に対する計算量の増え方を見ます。
```
Jacomichi: > 目標位相$0.6$

I don't think concrete numbers are necessary here.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,548 @@
...
+
+# シンボリックな推定がmに依存し、この範囲で増加することを確認します。
+assert "m" in str(symbolic_resource_estimate.gates.total)
+assert all(
+    later > earlier
+    for earlier, later in zip(resource_gate_counts, resource_gate_counts[1:])
+)
+
+# %% [markdown]
+# プロットは、位相の格子を細かくするためのコストを、$2^m$に比例する参照線とともに示しています。目標とする加法誤差を$\epsilon$とすると、格子幅はおおよそ
```
Jacomichi: Wouldn't the required resources exceed $2^m$?

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,548 @@
...
+# %%
+# QPEのスケーリングを見るため、最小限の1量子ビット位相ユニタリを使います。
+@qmc.qkernel
+def phase_unitary_for_resources(q: qmc.Qubit) -> qmc.Qubit:
+    q = qmc.p(q, 1.0)
+    return q
+
+
+# m個のカウント用量子ビットを持つシンボリックなQPE風リソース量子カーネルを作ります。
+@qmc.qkernel
```
Jacomichi: Can't we use Qamomile's `qpe` for this purpose? 
If it is available, I think it would be better to use it. 
If not, I believe Qamomile should provide this feature.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,548 @@
...
+    assert phase_error <= 1 / 2**counting_bits
+
+# %% [markdown]
+# ## リソース推定
+#
+# 前のサブセクションでは、カウント用量子ビットを増やすと精度が上がることを確認しました。下のシンボリックなリソース推定用の量子カーネルでは、カウント用量子ビット数を`m`として残し、制御対象には1量子ビットの位相ユニタリを使います。まず、カウント用レジスタそのものに由来するQPEのスケーリングをセルの結果として見ます。
+
+# %%
+# QPEのスケーリングを見るため、最小限の1量子ビット位相ユニタリを使います。
+@qmc.qkernel
```
Jacomichi: For this purpose, I think it would be better to use a stub gate.

Jacomichi (changes requested):
Thanks for PR!
I left some comments

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you for the comment. Fixed in `8b0cc39b`: updated the first Japanese occurrence to `Quantum Phase Estimation; QPE`. The title now keeps the shorter `(QPE)` form, while the first explanatory sentence carries the full notation.

jij-yuichiro-nakano: Thank you for the comment. Fixed in `8b0cc39b`: added a note explaining that QPE can also take a superposition of eigenstates, and that the tutorial uses a known eigenstate only to keep the phase-kickback equations and sampling results easy to read.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you for the comment. Fixed in `8b0cc39b`: replaced the ambiguous cost notation with `G_QPE(m)` and added a definition of `G(V)` as the logical gate cost for implementing operation `V`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you for the comment. Fixed in `8b0cc39b`: expanded the resource-estimation discussion to explain that the scaling depends on how the controlled unitaries are implemented, with examples from Shor's algorithm and Hamiltonian simulation.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you for the comment. Addressed in `8b0cc39b`: I changed the purpose of this resource-estimation section to estimate the QPE circuit for the same diagonal-unitary example used in the tutorial. Since the section now intentionally checks that concrete tutorial circuit rather than an abstract placeholder operation, a stub gate is no longer needed here.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you for the comment. Fixed in `8b0cc39b`: removed the concrete target phase value from the abstract text.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you for the comment. Fixed in `8b0cc39b`: the resource-estimation section now directly applies `estimate_resources()` to the same QPE kernel factory used by `run_qpe_experiment()`, so the estimate goes through Qamomile’s `qpe` helper.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you for the comment. Fixed in `8b0cc39b`: clarified that the `O(2^m)` repetition-model estimate is implementation-dependent, and added a more detailed discussion of controlled-unitary implementation cost and initial-state preparation cost.

jij-yuichiro-nakano:
Thank you for the comments! I have addressed the points you raised. Please take another look.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+#
+# $M=2^m$個の基底状態に対するQFTは
+#
+# $$
+# \mathrm{QFT}_M|x\rangle
+# = \frac{1}{\sqrt{M}}
+#   \sum_{y=0}^{M-1} e^{2\pi i xy/M}|y\rangle.
+# $$
+#
+# これは整数$x$を、計算基底全体にわたる規則的な位相パターンへ写します。詳しい説明はQFTチュートリアルをご覧ください。逆QFTはその逆の操作です。$\phi=a/M$が$m$ビットで正確に表せる場合、制御ユニタリーゲート（$U^{2^0}, U^{2^1}, \ldots, U^{2^{m-1}}$）によってカウント用レジスタは
```
Jacomichi: Would it be possible to add a link to the QFT tutorial?

Jacomichi (commented):
Thanks to update
I think this is fine

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+# %% [markdown]
+# ---
+# tags: [algorithm, primitive, resource-estimation]
+# ---
+#
+# # 量子位相推定（QPE）入門
+#
+# 量子位相推定（Quantum Phase Estimation; QPE）は、$U|\psi\rangle = e^{2\pi i \phi}|\psi\rangle$を満たすユニタリ$U$と固有状態$|\psi\rangle$から、固有位相$\phi$を推定するアルゴリズムです。Shorのアルゴリズムなど、ユニタリの固有値に埋め込まれた位相を使うアルゴリズムで中心的なプリミティブとして使われます{cite:p}`10.48550/arXiv.quant-ph/9511026,10.1098/rspa.1998.0164`。
+#
+# このノートブックでは、組み込みの`qpe`ヘルパーを4x4ユニタリに適用します。既知の固有状態を準備し、ローカルのQiskitシミュレータで回路を実行して、復号された位相を既知の固有位相と比較します。さらに、カウント用量子ビットを増やすと精度が上がることを確認し、リソース推定を使って、必要な精度に対する計算量の増え方を見ます。
```
ksk-jij: [my opinion]
We might not need "ローカルのQiskitシミュレータで" since other tutorial book does not say this.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+
+# %%
+# 最新のQamomileをpipからインストールします！
+# # !pip install qamomile
```
ksk-jij: [must]
Now, this needs to be `# !pip install "qamomile[qiskit,visualization]"`

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+#     language: python
+#     name: python3
+# ---
+
+# %% [markdown]
+# ---
+# tags: [algorithm, primitive, resource-estimation]
+# ---
+#
+# # 量子位相推定（QPE）入門
```
ksk-jij: [my opinion]
Could we just say "量子位相推定（QPE）" without "入門"? I suppose there are several QPE, but if there is only one QPE in the document, anyway I believe people would not think this QPE notebook would be advanced version or something. In that case, I feel simply saying "量子位相推定（QPE）" is enough. Also, Iif we have something advanced, we can add the name at that point. What do you think?

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+
+import matplotlib.pyplot as plt
+import numpy as np
+from qiskit_aer import AerSimulator
+
+import qamomile.circuit as qmc
+from qamomile.qiskit import QiskitTranspiler
+
+# QPEの回路構造が見えるように、複合ゲートを分解して扱います。
+transpiler = QiskitTranspiler(use_native_composite=False)
```
ksk-jij: [must & question]
I think this leads misunderstanding on `use_native_composite` argument. I suppose this option determines if we use the native implementation (here Qiskit( or Qamomile's one rather than just decomposing the one implementation. Also, I suppose this does not contribute anything for drawing QKernel with Qamomile visualisation feature.

Does this option affect to this notebook in any way? If it does not affect anything in the notebook, I think we can safely write just `transpiler = QiskitTranspiler()` and remove the comment.

docs/ja/algorithm/qpe.py
ksk-jij, line 1: [my opinion]
I feel like that this notebook could have two implementation, one from scratch and the built-in function for the same purpose as the hamiltonian simulation. What do you think? QPE is not difficult to implement here, so I suppose that does not mess anything here either. Users can see how to do it with Qamomile and know there is a QPE feature in Qamomile. If we don't write QPE from scratch, we also can omit naive implementation from the hamiltonian simulation notebook.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+
+# %% [markdown]
+# ## 背景: 位相キックバックと量子フーリエ変換
+#
+# QPEは、制御ユニタリーゲートによる位相キックバック{cite:p}`10.1098/rspa.1998.0164`と、量子フーリエ変換{cite:p}`10.48550/arXiv.quant-ph/0201067`を組み合わせます。
+#
+# ### 位相キックバック
+#
+# 位相キックバックは、対象レジスタの固有位相を制御レジスタへ移します。次を仮定します。
```
ksk-jij: [my opinion]
Since we don't really write "レジスタ" in Qamomile elsewhere, we might want to use another word for レジスタ. I will not write this comment other part, but the same comment is applied to the same word elsewhere.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+# U|\psi\rangle = e^{2\pi i\phi}|\psi\rangle.
+# $$
+#
+# 1つの制御量子ビットを重ね合わせにし、対象レジスタを$|\psi\rangle$のままにします。
+#
+# $$
+# \frac{|0\rangle + |1\rangle}{\sqrt{2}}|\psi\rangle.
+# $$
+#
+# $U$に対応する制御ユニタリーゲートは、制御量子ビットが$|1\rangle$の枝にだけ$U$を適用します。対象レジスタは固有状態なので、その枝に固有位相が乗ります。
```
ksk-jij: [my opinion]
"制御Uゲート" might be enough?

docs/ja/algorithm/qpe.py
ksk-jij: [my opinion]
Sorry if I am wrong, but personally "位相キックバックは、対象レジスタの固有位相を制御レジスタへ移します。" feels like "位相キックバック" itself implies the operation, but I feel "位相キックバック" is a name of phenomenon, (or result? rather than the operation name).
If I'm wrong, it is alright, but otherwise "位相キックバックとは対象の量子状態の固有位相が制御量子ビットへと映る現象です。" or something like that would be nice?

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+#
+# $$
+# \frac{|0\rangle|\psi\rangle + |1\rangle U|\psi\rangle}{\sqrt{2}}
+# =
+# \frac{|0\rangle + e^{2\pi i\phi}|1\rangle}{\sqrt{2}}|\psi\rangle.
+# $$
+#
+# 対象レジスタは同じ固有状態に戻り、制御量子ビットが位相を持ちます。QPEはこの仕組みを制御ユニタリーゲートで繰り返します。制御ユニタリーゲート（$U^{2^k}$）は位相$e^{2\pi i 2^k\phi}$をキックバックするため、異なる制御量子ビットが同じ固有位相の異なる2進重みを受け取ります。QPEでは、それぞれの制御量子ビットを個別に測定せず、キックバックされた位相を測定前まで保ちます。
```
ksk-jij: [my opinion]
I believe "異なる制御量子ビットが同じ固有位相の異なる2進重みを受け取ります" explains what happens through QPE, but for now we haven't seen QPE at all and the paragraph is continuous from sentences for the general phase kickback. We saw the case having there is only one control qubit. In that case, "異なる制御量子ビット" makes the readers a bit confusing?

"QPEはこの仕組みを制御ユニタリーゲートで繰り返します。" feels a bit weird? since we just showed the case with a controlled unitary operator, which is the same as this sentence.

Also, since we didn't mention measurement in the example above, I feel we don't need to say "QPEでは、それぞれの制御量子ビットを個別に測定せず、キックバックされた位相を測定前まで保ちます。".

How about simply saying "QPEでは複数の制御量子ビットを使って，それぞれの制御量子ビットに対して位相キックバックを繰り返し起こすことで固有位相の情報を集めます" or something like that (I don't think this is the best way tho), without talking too much about QPE.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+#
+# ### 量子フーリエ変換
+#
+# $M=2^m$個の基底状態に対するQFTは
+#
+# $$
+# \mathrm{QFT}_M|x\rangle
+# = \frac{1}{\sqrt{M}}
+#   \sum_{y=0}^{M-1} e^{2\pi i xy/M}|y\rangle.
+# $$
```
ksk-jij: [my opinion]
This is a bit abrupt I feel. Like other parts we finish with "です/ます" or something like that, but this feels "だ/である". (Obviously we didn't say anything, but since we didn't say anything here, it feels like so).

Also, if possible, it would be nice to start with what QFT to have the same structure as phase kickback section.

docs/ja/algorithm/qpe.py
ksk-jij: [question]
What is "カウント用" here? I feel like this explanation is for these who know QPE?

ksk-jij: [my opinion]
Might be nice to say "逆QFT $\mathrm{QFT}_M^{-1}$はその逆の操作です" since we use this notation right below.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+# \end{pmatrix}
+# }_{\text{位相の値のベクトル}}
+# \xrightarrow{\mathrm{QFT}_M^{-1}}
+# \underbrace{|a\rangle = |a_{m-1}\cdots a_0\rangle}_{\text{計算基底のビット列}}.
+# $$
+
+# %% [markdown]
+# ## アルゴリズム
+#
+# QPEは、1つの制御量子ビットで見た位相キックバックを、レジスタ全体の手順へ拡張します。カウント用レジスタはまず、すべての2進重みを重ね合わせとして保持します。そこへ$U$の制御ユニタリーゲートを適用すると、対応する重み付きの固有位相がカウント用レジスタへキックバックされます。最後に逆QFTを使い、このフーリエ符号化された位相パターンを通常の2進整数として読み出すことで、$\phi$の推定値を得ます。
```
ksk-jij: [my opinion]
Might be nice to mention "フーリエ符号化" in QFT section if we use this word here.

ksk-jij: [my opinion]
Might be nice to say "そこへ$U$の制御ユニタリーゲートを繰り返し適用すると、繰り返しの回数に応じた重み付きの固有位相がカウント用の量子ビットたちへとキックバックされます" or something like that.

ksk-jij: [my opinion]
Might be nice to say how many kinds of quantum states are needed, which is two for counting and controlling I suppose.

ksk-jij: [my opinion]
Since we explain the steps right below, we might not need to these details here. We just need to setup what notation and words we use right below.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+# $$
+# |\Psi_1\rangle =
+# H^{\otimes m}|0\rangle^{\otimes m}|\psi\rangle
+# =
+# \frac{1}{\sqrt{M}}\sum_{r=0}^{M-1}|r\rangle|\psi\rangle.
+# $$
+#
+# ### ステップ2：制御ユニタリーゲートを適用する
+#
+# 各カウント用量子ビット$k$に対して、制御ユニタリーゲート（$U^{2^k}$）を適用します。$r=\sum_{k=0}^{m-1} r_k2^k$と書くと、対象の固有状態には位相$e^{2\pi i\phi r}$が乗ります。
```
ksk-jij: [question]
Why do we write （$U^{2^k}$) instead of just $U^{2^k}$? Personally, I feel the information $U^{2^k}$ is necessary here. (Or as same as before $U^{2^k}$の制御ユニたりゲート?)

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+# $$
+#
+# $$
+# |\Psi_3\rangle =
+# \mathrm{QFT}_M^{-1}\mathrm{QFT}_M|a\rangle|\psi\rangle
+# =
+# |a\rangle|\psi\rangle.
+# $$
+#
+# ### ステップ4：位相を測定する
```
ksk-jij: [my opinion]
Might be nice to say 復号する? since we use 位相復号 in the next section and we indeed decode the phase in this step rather than only measuring? But, in that case, the previous step title might also need to be updated since it also has 復号?

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+# ```
+
+# %% [markdown]
+# ## Qamomileでの実装：`qpe`関数
+#
+# `qmc.qpe`関数は、QPEのパターン全体を適用します。アダマールゲート、渡したユニタリ量子カーネルに対応する制御ユニタリーゲート、逆QFT、固定小数点の位相復号をまとめて扱えます。
+#
+# ### 問題例
+#
+# ここでは**対角**4x4ユニタリを使います。
```
ksk-jij: [must]
In this notebook, we've been saying ユニタリー and here is ユニたり. Personally I really like ユニタリ since the total number of letters is one less. And I believe other notebooks say ユニタリ? (This is a wild guess tho). Anyway we should use the same notation.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+print("phase fractions:", np.round(phase_fractions, 6))
+print("target phase fraction:", f"{TARGET_PHASE_FRACTION:.8f}")
+print("U =")
+print(np.round(unitary, 3))
+assert 0.0 <= TARGET_PHASE_FRACTION < 1.0
+
+# %% [markdown]
+# ### `qpe`による量子カーネルの実装
+#
+# まず、位相を推定したい4x4ユニタリを定義します。下の`diagonal_4x4`量子カーネルは、この行列を直接実装します。Qamomileでは、`qmc.p(q, theta)`は`q`の$|1\rangle$成分に$e^{i\theta}$を掛けます。そのため、`qmc.p(q[0], phi10)`は対象レジスタの最初のビットが1であるすべての基底状態に位相$e^{i\theta_{10}}$を与え、`qmc.p(q[1], phi01)`は対象レジスタの2つ目のビットが1であるすべての基底状態に位相$e^{i\theta_{01}}$を与えます。この時点で$|11\rangle$には$e^{i(\theta_{10}+\theta_{01})}$が乗っているため、制御位相シフトゲートでは補正分
```
ksk-jij: [question]
"Qamomileでは、`qmc.p(q, theta)`は`q`の$|1\rangle$成分に$e^{i\theta}$を掛けます。" Is this Qamomile unique stuff? I feel this is the definition of the P gate not only Qamomile.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+diagonal_4x4.draw(
+    q=2,
+    phi01=PHI_01,
+    phi10=PHI_10,
+    phi11=PHI_11,
+    fold_loops=False,
+)
+
+# %% [markdown]
+# `qmc.qpe`関数は`QFixed`の位相レジスタを返します。`QFixed`は、量子レジスタで固定小数点数を表すQamomileの型です。ここでは位相を$[0,1)$の小数として保持し、測定すると浮動小数点の位相推定値を直接取得できます。下の量子カーネルでは、固有状態$|01\rangle$を準備し、`diagonal_4x4`にQPEを適用して、復号済みの位相を測定します。ここでは小さなファクトリを使い、同じ問題ユニタリに対してカウント用量子ビット数を変えられるようにします。
```
ksk-jij: [my opinion]
"復号済みの位相を測定します。" feels a bit weird? Like if we say 測定 in this context, it feels quantum measurement, but we don't perform quantum measurement to get a already-decoded-phase?

ksk-jij: [question]
Is there any specific reason to use the factory function? I feel the following qkernel does the trick?

```python
# カウント用レジスタサイズを変えられるQPE量子カーネルを作ります。
@qmc.qkernel
def qpe_with_stdlib(
    counting_bits: qmc.UInt,
    phi01: qmc.Float,
    phi10: qmc.Float,
    phi11: qmc.Float,
) -> qmc.Float:
    # カウント用レジスタと2量子ビットの対象レジスタを確保します。
    counting = qmc.qubit_array(counting_bits, name="counting")
    target = qmc.qubit_array(2, name="target")
    # 対象固有状態として|01>を準備します。
    target[1] = qmc.x(target[1])

    # 標準ライブラリのQPEを適用し、復号された固定小数点位相を測定します。
    phase = qmc.qpe(
        target,
        counting,
        diagonal_4x4,
        phi01=phi01,
        phi10=phi10,
        phi11=phi11,
    )
    return qmc.measure(phase)



# 回路図で使う小さなQPEインスタンスを描画します。
qpe_with_stdlib.draw(
    counting_bits=DRAW_COUNTING_BITS,
    phi01=PHI_01,
    phi10=PHI_10,
    phi11=PHI_11,
    fold_loops=False,
)
```

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+# $$
+#
+# この行列では、すべての計算基底状態が固有状態です。対象状態$|01\rangle$を準備し、推定したい位相$\theta_{01} / 2\pi$を$0.6$に設定します。$0.6$は2進格子上で正確には表現できないため、カウント用量子ビットを増やす効果が見えやすくなります。
+
+# %%
+# カウント用レジスタのサイズ、サンプリング設定、対象固有状態を決めます。
+docs_test_mode = os.environ.get("QAMOMILE_DOCS_TEST") == "1"
+COUNTING_BIT_OPTIONS = tuple(range(3, 6)) if docs_test_mode else tuple(range(3, 10))
+DRAW_COUNTING_BITS = 3
+EXAMPLE_COUNTING_BITS = 3
```
ksk-jij: [my opinion]
Personally, these don't really need to define as constants. I feel defining them here is not really helpful for anyone?

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+    phi11=PHI_11,
+    fold_loops=False,
+)
+
+# %% [markdown]
+# ## 実行結果
+#
+# ### 回路のトランスパイルと実行
+#
+# 測定された`QFixed`値は浮動小数点の位相推定値として返るため、ビット列を手動で復号する必要はありません。最初の実行では、回路図と同じ3個のカウント用量子ビットを使います。実験では、`sample_result.results`のうち測定回数が最も多い要素をそのまま表示し、後続セルで使う推定位相だけを返します。
```
ksk-jij: [my opinion]
"最初の実行" feels weird since in this section we execute only once. I think the purpose of the section is to show transpilation and execution, but maybe we can just omit this subsection completely. We just can show the graph of the accuracy and the number of qubits for counting.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+    marker="o",
+    color="#2696EB",
+    label="QPE gate count",
+)
+ax.plot(
+    bits,
+    scaling_reference,
+    linestyle="--",
+    color="#DB4D3F",
+    label=r"theory: QPE O$(2^m)$",
```
ksk-jij: [my opinion]
How about just saying O(2^m)? Theory: QPE feels like this plot shows the exact number of gates for QPE theoretically.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+plt.show()
+
+# 直接推定したゲート数が、この範囲で増加することを確認します。
+assert all(
+    later > earlier
+    for earlier, later in zip(resource_gate_counts, resource_gate_counts[1:])
+)
+
+# %% [markdown]
+# プロットは、位相の格子を細かくするためのコストを、$2^m$に比例する参照線とともに示しています。目標とする加法誤差を$\epsilon$とすると、格子幅はおおよそ
```
ksk-jij: [question]
Sorry for my ignorance, but "位相の格子" is a standard word? I get what you mean, but not super sure if this word makes the explanation easier to understand. Like, how about directly saying "プロットは、QPEの精度を上げるために、必要となる量子ビットの数とそのときのゲートの数を$2^m$に比例する参照線とともに示しています。"

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+#
+# を満たす必要があります。そのため、必要なカウント用量子ビット数は
+#
+# $$
+# m = O\!\left(\log\frac{1}{\epsilon}\right)
+# $$
+#
+# です。
+#
+# 制御ユニタリーを繰り返しモデルで考える場合、QPEは重み$1,2,\ldots,2^{m-1}$に対応する制御ユニタリーゲートを適用します。したがって、制御ユニタリーゲートの適用回数は
```
ksk-jij: [my opinion]
We have introduced only one way, I think we don't need to mention this at least here.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+    label="QPE gate count",
+)
+ax.plot(
+    bits,
+    scaling_reference,
+    linestyle="--",
+    color="#DB4D3F",
+    label=r"theory: QPE O$(2^m)$",
+)
+ax.set_xlabel("counting qubits")
```
ksk-jij: [my opinion]
"counting qubits (m)" might be kind.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+# =
+# O\!\left(\frac{1}{\epsilon}\right)
+# $$
+#
+# です。つまり、カウント用量子ビット数は$1/\epsilon$に対して対数的ですが、繰り返しによる制御ユニタリーゲートの回数は$O(1/\epsilon)$で増えます{cite:p}`10.1017/CBO9780511976667`。
+#
+# 最後に、このスケーリングは実装方法に依存します。QPEでは
+# $\mathrm{controlled}\text{-}U^{2^0}, \ldots,
+# \mathrm{controlled}\text{-}U^{2^{m-1}}$という制御ユニタリーが必要です。
+# これらを$U$の繰り返しで実装する場合、コストは$O(2^m)=O(1/\epsilon)$となり、
```
ksk-jij: [my opinion]
I suppose we are repeating the same stuff here again, so how about "今回のようなナイーブな実装でのゲート数は先ほど見た通り$O(1/\epsilon)$となり，高精度なQPEを実行するのは効率的ではありません．"

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+#
+# $$
+# \sum_{k=0}^{m-1} 2^k = 2^m - 1
+# =
+# O\!\left(\frac{1}{\epsilon}\right)
+# $$
+#
+# です。つまり、カウント用量子ビット数は$1/\epsilon$に対して対数的ですが、繰り返しによる制御ユニタリーゲートの回数は$O(1/\epsilon)$で増えます{cite:p}`10.1017/CBO9780511976667`。
+#
+# 最後に、このスケーリングは実装方法に依存します。QPEでは
```
ksk-jij: [must]
The artificial breakdowns are inserted from here. Could you eliminate them?

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+# 高精度では通常効率的ではありません。より一般には、コストは
+#
+# $$
+# G_{\mathrm{QPE}}(m)
+# =
+# \sum_{k=0}^{m-1} G\!\left(\mathrm{controlled}\text{-}U^{2^k}\right)
+# + O(m^2)
+# $$
+#
+# と書けます。ここで$G(V)$は、操作$V$を実装するための論理ゲートコストを表します。
```
ksk-jij: [my opinion]
This feels a bit abrupt. How about saying
```
"より一般には，あるユニタリ$V$を実行するのに必要なゲート数を$G(V)$として，QPE全体で必要となるゲート数は
$$
G_{\mathrm{QPE}}(m)
=
\sum_{k=0}^{m-1} G\!\left(\mathrm{controlled}\text{-}U^{2^k}\right)
+ O(m^2)
$$
とかけます．"
```

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+# このページでは、コストの単位がゲート数であることを明示するために$G$を使います。
```
ksk-jij: [my opinion]
This sentence is too late to mention I think. Also, if we just say ゲート数, then we don't need to say this.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+#
+# :::{note}
+# $m$個の量子ビットに対する逆QFTは、$O(m^2)$個のゲートに分解できます。詳しくはQFTのチュートリアルを参照してください。
+# :::
+#
+# しかし、実際には以下の要因によってこの見積もりは大きく変わる可能性があります。
```
ksk-jij: [my opinion]
Do you mean $O(1/\epsilon)$ by "この見積もり" or the general one? I thought it means $O(1/\epsilon)$, but if so, this sentence order makes a bit difficult to understand that.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+#
+# 1. **制御ユニタリーの実装コスト**
+#
+#    多くの場合、主なコストは各制御ユニタリーをどう実装するかで決まります。
```
ksk-jij: [my opinion]
This feels like saying the same stuff as the general gate counts above. We might want to move the formula from where we now put to here.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+#    Shorのアルゴリズムでは、モジュラー乗算に対応する制御ユニタリーを問題構造から構成できるため、
+#    $U$を指数回反復する必要はありません。量子系のハミルトニアンシミュレーションでも、
+#    ハミルトニアンの形やシミュレーション手法によっては、制御された時間発展を多項式程度のリソースで実装できる場合があります。
```
ksk-jij: [my opinion]
I believe this notebook is for two kinds of people, one is person who knows QPE but would like to see how to implement QPE with Qamomile, and the other one is who would like to learn QPE (otherwise explaining QPE is redundant.) In that case, these sentences without saying QPE is used in Shor or Hamiltonian simulation are abrupt for the person who would like to learn QPE.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+#    そのため、QPEのリソースを見積もるときは、制御ユニタリーを$U$の反復として数えるのか、
+#    直接合成した回路として数えるのか、あるいは問題固有の算術回路やシミュレーション回路として与えるのかを明示する必要があります。
+#
+# 2. **初期状態の準備コスト**
+#
+#    上の見積もりには、対象レジスタの初期状態を準備するコストも含まれていません。
```
ksk-jij: [question]
Which one would you like to mention, general one or one for the simple repeat version? I thought both, if so, how about "ここまでのゲート数の見積もりでは，初期状態を準備するコストについては考慮していません。一方で，QPEの成功確率は初期状態をどれだけ固有状態に近似できるかが重要になってきます。一般に良い近似状態を用意するための量子ゲート数は非自明であり，実際の応用で全体のリソースを評価する時にはこの準備にかかる量子ゲート数は別途見積もる必要があります．" or something like that.

docs/ja/algorithm/qpe.py
```diff
@@ -0,0 +1,540 @@
...
+# %% [markdown]
+# ## まとめ
+#
+# このノートブックでは、対角4x4ユニタリでQPEを実装し、復号された位相をサンプリングして、カウント用レジスタを大きくしたときの精度とリソースの変化を確認しました。
+#
+# - `qmc.qpe`は、渡したユニタリ量子カーネルに対して、アダマールゲート、制御ユニタリーゲート、逆QFT、固定小数点の位相復号を適用します。
+# - この例では固有状態$|01\rangle$を準備し、目標位相$0.6$を浮動小数点の`QFixed`測定結果として直接推定します。
+# - カウント用量子ビットを増やすと2進位相の格子が細かくなるため、サンプリングされた推定値が目標位相に近づきます。
+# - `estimate_resources()`を直接呼び出すことで、単純な繰り返し実装では$m=O(\log(1/\epsilon))$個のカウント用量子ビットに対して、$O(1/\epsilon)$回の制御ユニタリーゲートが必要になることを確認できます。
+# - 実際の応用では、制御ユニタリーの実装方法と初期状態の準備コストを含めるかどうかを明示することが重要です。
```
ksk-jij: [must]
Would be nice to stick with the skeleton.
```md
# ## Summary
#
# In this notebook, we:
#
# - Take-home message 1.
# - Take-home message 2.
# - Take-home message 3.
```

Currently, I think this summary section lists what we literally did. But, I feel here it would be nice to list only important stuff rather than what we actually did. Like "この例では固有状態$|01\rangle$を準備し、目標位相$0.6$を浮動小数点の`QFixed`測定結果として直接推定します。" can be the one we can ommit.

ksk-jij (changes requested):
@jij-yuichiro-nakano Thank you for the PR! I've left comments, please have a look at them :)

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Agreed. I simplified the title to “Quantum Phase Estimation (QPE)”. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Agreed. I added both a from-scratch QPE implementation and an implementation using the built-in `qmc.qpe` function under the Qamomile implementation section. The notebook also checks that both implementations return the same estimate. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you for raising this distinction. I revised the explanation to describe phase kickback as the mechanism by which applying a controlled-\(U\) gate to an eigenstate causes the eigenphase to appear as a relative phase on the control qubit. This distinguishes the controlled operation from the resulting mechanism. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I updated the installation command to `pip install "qamomile[qiskit,visualization]"`. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I replaced the register-based wording throughout the Japanese page with qubit-based wording, such as “counting qubits”. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I removed the backend-specific simulator wording from the overview and rewrote it to focus on what the notebook covers. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. This option was not needed for the notebook, so I replaced it with the default `QiskitTranspiler()` and removed the misleading comment. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Agreed. I standardized the terminology to “controlled-\(U\) gate” throughout the explanation. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I added an introductory definition of the QFT before presenting its formula and adjusted the prose to match the style of the preceding section. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I simplified this transition to state only that QPE deliberately uses the same mechanism with multiple control qubits. The controlled powers, binary weights, and measurement details are now deferred to the Algorithm section. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. The Algorithm section now explicitly defines counting qubits as auxiliary qubits used to read the phase as an \(m\)-bit binary fraction. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I now define a Fourier-encoded state in the QFT section before using the term later in the notebook. Fixed in `d0c46ff8`.

jij-yuichiro-nakano: Thank you. The Algorithm introduction now identifies the two groups of qubits: \(m\) counting qubits and the target qubits holding the eigenstate. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Agreed. I introduced the notation \(\mathrm{QFT}_M^{-1}\) when defining the inverse QFT. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. Step 2 now explicitly explains that each counting qubit controls a controlled-\(U^{2^k}\) gate and shows how the weighted eigenphase is encoded. Fixed in `d0c46ff8`.

jij-yuichiro-nakano: Agreed. I reduced the introduction to the required terminology, initial states, and notation, leaving the procedural details to the four steps below. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Agreed. I updated Step 3 to describe converting phase information into a bit string with the inverse QFT, and Step 4 now describes measuring the bit string and decoding the phase. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I revised the explanation to distinguish the quantum measurement of the qubit string from its subsequent decoding into a fixed-point phase value. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I standardized the Japanese transliteration of “unitary” throughout the page and corrected the inconsistent spelling. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Agreed. I removed the factory function and made `counting_bits` a `qmc.UInt` argument of the qkernel, which is supplied as a compile-time binding. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I clarified that multiplying the \(|1\rangle\) component by \(e^{i\theta}\) is the definition of the phase gate, and then separately explained that Qamomile expresses it as `qmc.p(q, theta)`. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I removed the constants that only represented fixed counting-qubit choices and now construct those values directly where they are used. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I rewrote the Step 2 title and description to refer directly to controlled-\(U^{2^k}\) gates without the parenthetical notation. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Agreed. I removed the separate first-execution subsection and consolidated execution into the comparison across counting-qubit sizes. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Agreed. I moved the definition of \(G(V)\) before the formula and removed the later explanatory sentence. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I replaced “phase grid” with a direct explanation of the representable phase spacing and rewrote the plot description in terms of counting qubits, precision, and total gate count. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I updated the horizontal-axis label to `counting qubits (\(m\))`. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I removed the artificial line breaks and kept the prose as continuous paragraphs. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Agreed. I changed the reference-curve label to \(O(2^m)\) so it does not imply an exact theoretical QPE gate count. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I introduced the general QPE gate-count formula using \(G(V)\) as the gate count required to implement a unitary \(V\). Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I rephrased the sentence to state that this repeated implementation has an \(O(1/\epsilon)\) gate count and is inefficient for high-precision QPE. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Agreed. I now state directly that this particular implementation constructs controlled-\(U^{2^k}\) gates by repeating \(U\), instead of introducing it as one model among several at that point. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I agree that moving the formula would group the discussion by topic. However, I kept it before the numbered list because placing the formula and the inverse-QFT note under the first item would make that item disproportionately long relative to the initial-state preparation item. The transition now explicitly explains what the formula includes and introduces the two factors separately, so I believe the current placement provides a better balance. No change was made for this suggestion.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I added context explaining how QPE is used for order finding in Shor’s algorithm and for energy estimation through time-evolution eigenphases in quantum simulation. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I clarified that the displayed formula covers the gate count of the QPE procedure itself, excludes initial-state preparation, and depends on how each controlled-\(U^{2^k}\) gate is implemented. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I clarified that the probability of obtaining a particular eigenphase is determined by the squared overlap with the corresponding eigenstate, and that preparing a sufficiently good approximate state may require a separate, nontrivial gate-cost estimate. Fixed in `d0c46ff8`.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. I rewrote the Summary section to follow the documentation skeleton and reduced it to three main takeaways: how QPE reads an eigenphase, how to implement it with Qamomile, and what determines its precision and gate count. Fixed in `d0c46ff8`.

jij-yuichiro-nakano:
@ksk-jij  Thank you for the comments! I have addressed the points you raised. Please take another look.

docs/ja/algorithm/qpe.py
jij-yuichiro-nakano: Thank you. The QFT tutorial is currently in an open PR, and we plan to add the link once it's merged.

Jacomichi (approved):
LGTM

PR #538, opened 2026-06-29
Add qmc.uncompute primitive for in-place adjoint application

ksk-jij:
## Summary

Adds `qmc.uncompute(target, *args, **kwargs)`, a thin, intent-revealing wrapper over the existing `qmc.inverse` that applies the adjoint of an operation at the current trace site and returns the resulting handle(s). It makes the *compute -- use -- uncompute* pattern explicit: `qmc.uncompute(layer, q, theta)` replaces the more verbose `qmc.inverse(layer)(q, theta)` that the docs currently spell out by hand.

Unlike `inverse`, which returns a reusable callable adjoint wrapper for later application, `uncompute` applies the adjoint immediately. Every target kind supported by `inverse` is supported here (native gates, `qmc.qft`/`qmc.iqft`, and user `QKernel`s including multi-quantum-input kernels that return tuples), and the existing diagnostics propagate unchanged: `TypeError` for a direct `CompositeGate` instance, `NotImplementedError` for kernels containing un-invertible control flow.

The implementation is deliberately minimal — it delegates to `inverse` and adds no new IR nodes, transpiler passes, or backend code. The symbol is exported from `qamomile.circuit` (import and `__all__`).

A context-manager / `with`-scope form was considered but deferred: user-authored `with` blocks are not currently supported by the frontend AST transform, so a scope API would require a large frontend change. The acceptance criteria explicitly permit a helper form, which this PR delivers. Closes BACKLOG-7975.

## Test plan

New `tests/circuit/test_uncompute.py` follows the cross-backend execution policy in `CLAUDE.md`:

- Runs on Qiskit, QuriParts, and CUDA-Q (each guarded with a skip when the SDK is absent; CUDA-Q gated behind its mark).
- Exercises **both** the sampling path (`executable.sample(...)`) and the expectation-value path (`executable.run(...)` over a sum-Z observable), parametrized over seeds, register sizes (1, 2, 3, 5), and boundary/random angles (`0`, `π`, `2π`, seeded random).
- Covers native gates, native rotation gates with keyword-argument forwarding, multi-quantum-input kernels returning tuples, and the stdlib QFT.
- Asserts equivalence with the `qmc.inverse(...)` spelling via Qiskit statevector comparison, and verifies the `TypeError` / `NotImplementedError` diagnostics propagate.

Verification on this branch:

- `uv run ruff check qamomile/ tests/` — passes
- `uv run ruff format --check qamomile/ tests/` — passes
- `uv run zuban check qamomile/` — passes
- `uv run pytest tests/` — 8540 passed, 443 skipped, 1 xfailed

---
_Generated by [Claude Code](https://claude.ai/code/session_01AsKKLEszo99GHyZxgRThoB)_

### DIFF
--- a/qamomile/circuit/__init__.py
+++ b/qamomile/circuit/__init__.py
@@ -24,7 +24,7 @@ from .frontend.operation.cast import cast
 from .frontend.operation.control import control
 from .frontend.operation.control_flow import for_items, items, range
 from .frontend.operation.expval import expval
-from .frontend.operation.inverse import inverse
+from .frontend.operation.inverse import inverse, uncompute
 from .frontend.operation.measurement import measure
 from .frontend.operation.pauli_evolve import pauli_evolve
 from .frontend.operation.qubit_gates import (
@@ -97,6 +97,7 @@ __all__ = [
     "CompositeGate",
     "control",
     "inverse",
+    "uncompute",
     "cast",
     "bit",
     "float_",
--- a/qamomile/circuit/frontend/operation/inverse.py
+++ b/qamomile/circuit/frontend/operation/inverse.py
@@ -1860,3 +1860,72 @@ def inverse(target: QKernel | Callable[..., Any]) -> Any:
         )
     qkernel = _qkernel_for_callable(target, caller="inverse")
     return InverseGate(qkernel)
+
+
+def uncompute(
+    target: QKernel | Callable[..., Any],
+    *args: Any,
+    **kwargs: Any,
+) -> Any:
+    """Uncompute an operation by applying its adjoint at the current site.
+
+    Many quantum algorithms (Grover oracles, phase-kickback tricks,
+    amplitude amplification, arithmetic and chemistry subroutines) follow a
+    *compute -- use -- uncompute* pattern in which the uncompute step is the
+    adjoint (dagger) of the compute step. `uncompute` makes that final step
+    explicit and applies it immediately, so
+
+        >>> q = qmc.uncompute(layer, q, theta)  # doctest: +SKIP
+
+    reads as "undo ``layer(q, theta)``" and is equivalent to the more
+    verbose `qmc.inverse(layer)(q, theta)`. It is a thin, intent-revealing
+    wrapper over `inverse`: `target` is inverted with the same machinery
+    (native gates, QFT/IQFT, and user `QKernel`s are all supported), and the
+    resulting adjoint is invoked with the supplied arguments.
+
+    Unlike `inverse`, which returns a *callable* adjoint wrapper for later
+    application, `uncompute` applies the adjoint at the call site and returns
+    the resulting quantum handle(s). Use `inverse` when a reusable inverse
+    callable is needed (for example to pass it around or call it multiple
+    times); use `uncompute` to undo a computation in place.
+
+    Args:
+        target (QKernel | Callable[..., Any]): Native gate function,
+            `QKernel`, or supported stdlib function (e.g. `qmc.qft`) whose
+            adjoint should be applied. The same targets accepted by
+            `inverse` are accepted here.
+        *args (Any): Positional arguments forwarded to the adjoint of
+            `target` -- typically the quantum handle(s) (and any classical
+            parameters) the forward computation was applied to.
+        **kwargs (Any): Keyword arguments forwarded to the adjoint of
+            `target`.
+
+    Returns:
+        Any: The quantum output handle produced by applying the adjoint, or
+            a tuple of handles when `target` has multiple quantum inputs --
+            mirroring the return shape of the forward call.
+
+    Raises:
+        TypeError: If `target` cannot be interpreted as a gate-like callable,
+            if a `CompositeGate` instance is passed directly, or if the
+            forwarded arguments do not match the adjoint's signature.
+        NotImplementedError: If the inverted kernel uses operations that the
+            inverse machinery cannot reverse (e.g. `if`/`while`/`for items`
+            control flow, `QInit`, or a `ForOperation` whose bounds are not
+            compile-time constants).
+
+    Example:
+        >>> import qamomile.circuit as qmc
+        >>> @qmc.qkernel
+        ... def prep(q: qmc.Qubit, angle: qmc.Float) -> qmc.Qubit:
+        ...     q = qmc.h(q)
+        ...     q = qmc.rz(q, angle)
+        ...     return q
+        >>> @qmc.qkernel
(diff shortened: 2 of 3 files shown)


ksk-jij:
## Branch summary — add-uncompute-primitive

### Problem overview

This branch is a feature addition for BACKLOG-7975 ("Add an uncompute primitive"), not a bug fix. A large class of quantum algorithms (Grover oracles, phase kickback, amplitude amplification, arithmetic and chemistry subroutines) follow a "compute → use → uncompute (reverse to restore)" shape, where the final uncompute step is the adjoint (dagger) of the compute step. `main` already provides `qmc.inverse` (a function returning a *callable wrapper* representing the adjoint of a `QKernel` — one quantum subroutine traced by the qkernel decorator — or of native gates), but there was no dedicated primitive that lets users write "uncompute here" directly. On `main`, writing the uncomputation requires the two-step form `q = qmc.inverse(layer)(q, theta)` ("first obtain the inverse wrapper, then call it"); in fact the docs expressed it with the idiom `qmc.inverse(layer)(q, angle)  # uncompute`. On this branch, the new `qmc.uncompute(layer, q, theta)` produces the same result in a single step and makes the actual meaning of uncompute — "apply the adjoint right here and return the resulting handle" — explicit in the API.

### Frontend changes (the level of code users write)

From the user's perspective, a new public function `qmc.uncompute(target, *args, **kwargs)` becomes available directly from `import qamomile.circuit as qmc`. The first argument `target` is the object to uncompute (a native gate function, `qmc.qft`/`qmc.iqft`, or a user-defined `QKernel` — the same set `qmc.inverse` accepts), followed by the same arguments (quantum handles and any classical parameters) the forward computation was applied with; its adjoint is applied immediately at the current trace position and the resulting handle (or a tuple of handles when the target has multiple quantum inputs) is returned. Whereas `qmc.inverse` returns "a callable wrapper to be invoked later", `qmc.uncompute` "applies it in place right now and returns the result"; this distinction is stated in the docstring. One code example:

```python

@qmc.qkernel
def prep(q: qmc.Qubit, angle: qmc.Float) -> qmc.Qubit:
    q = qmc.h(q)
    q = qmc.rz(q, angle)
    return q

@qmc.qkernel
def circuit(angle: qmc.Float) -> qmc.Qubit:
    q = qmc.qubit("q")
    q = prep(q, angle)                 # compute
    q = qmc.uncompute(prep, q, angle)  # uncompute (q returns to |0>)
    return q
```

Error behavior is also consistent from the user's point of view. Passing a direct `CompositeGate` instance raises `TypeError` just as `qmc.inverse` does (with guidance to pass the functions `qmc.qft`/`qmc.iqft` for QFT/IQFT); a kernel containing non-invertible control flow (`if`/`while`/`for items`, etc.) raises `NotImplementedError`, and both propagate unchanged through `uncompute`.

### Backend changes (IR and the rest of the stack)

The implementation is minimal: a new function `uncompute` was added to the frontend-layer file `qamomile/circuit/frontend/operation/inverse.py` (defined at `qamomile/circuit/frontend/operation/inverse.py:1865`, with a one-line body `return inverse(target)(*args, **kwargs)` at `qamomile/circuit/frontend/operation/inverse.py:1931`). It is a thin wrapper delegating to the existing `inverse` (`qamomile/circuit/frontend/operation/inverse.py:1812`); no changes were made to IR nodes, transpiler passes, or any backend emitter. Because `inverse`'s return value is, for every kind of target, "a callable object returning handles" (native gates: the gate function or an angle-negating `_InverseRotationCallable`; QFT/IQFT: the paired function; `QKernel`: `InverseGate`), simply calling it with `(*args, **kwargs)` handles all kinds uniformly. For public-API registration, `uncompute` was added to the import (`qamomile/circuit/__init__.py:27`) and to `__all__` (`qamomile/circuit/__init__.py:100`). Tests live in the new file `tests/circuit/test_uncompute.py` (8 test functions): for the three supported backends (Qiskit / QuriParts / CUDA-Q, with skip guards when the SDK is absent) they exercise both the sampling path and the expectation-value path, parametrized over seeds, register sizes (1, 2, 3, 5), and angles (0, π, 2π, random seeds), and they additionally verify statevector equivalence between `uncompute` and `inverse`, keyword-argument forwarding, tuple returns for multi-quantum-input kernels, and propagation of the diagnostics (`TypeError`/`NotImplementedError`).

### Alternatives considered and rationale

Three options were considered: (A) a context-manager / scope syntax such as `with qmc.compute(...):` that automatically reverses the enclosed region, (B) an alias function that, like `qmc.inverse`, merely "returns a callable wrapper", and (C) a helper that applies in place. Because BACKLOG-7975's acceptance criteria explicitly permit "a scope OR a helper", (A) was examined first, but a user-written `with` statement inside a qkernel body is not currently supported by the frontend AST transform (there is no `visit_With`-equivalent user-facing hook in `ast_transform.py`); supporting it would require large additions to the tracer and AST transform, which is risky with respect to the "minimal change to complete" policy and the handling of the quantum linear-type discipline (each quantum handle must be used exactly once), so it was deferred. (B) was rejected because it would amount to little more than an alias of the existing `inverse` and adds little value. Ultimately (C) was adopted, giving `uncompute` the semantics "apply the adjoint in place and return the resulting handle". This best matches the meaning of the word uncompute ("reverse-compute here"), provides the practical benefit of replacing the docs idiom `qmc.inverse(layer)(q, angle)  # uncompute` in a single step, and is low-risk because it delegates to the well-tested `inverse` foundation. The role difference from `inverse` (deferred callable vs immediate application) is stated in the docstring to avoid confusing the two.

### Known limitations

The context-manager / scope form (the ergonomics of wrapping a compute region in a `with` block that auto-reverses on exit) is not provided in this branch. When: when a user wants the "enclose the compute region in a block and auto-uncompute on close" style. Why: as above, a user-facing `with` syntax is not supported by the frontend AST transform, and supporting it would require a large change, so it is out of scope for this minimal change. Future fix: when `with`-scope support inside a qkernel body is added in the future, a scope API built on top of `qmc.uncompute` can be considered as a separate backlog / separate PR. Also, the set of invertible targets is limited to what `qmc.inverse` supports; kernels containing non-invertible constructs such as `if`/`while`/`for items` still raise `NotImplementedError` (this is a pre-existing constraint inherited from `inverse`, neither widened nor narrowed by this change). Furthermore, this PR leaves out the "internal use in Grover-style stdlib kernels" and "dedicated tutorial" items mentioned by BACKLOG-7975, since the repository currently has no corresponding Grover stdlib kernel, and adding docs entails translation and docs-build conventions that would broaden the change; these are deferred to future work. For the project-wide list of constraints, see LIMITATIONS.md.

---

PR #543, opened 2026-07-02
Reject aliased quantum arguments in qkernel calls

ksk-jij:
## Summary

Passing the same `VectorView` twice to a plain qkernel sub-kernel silently aliased both parameter registers onto the same physical qubits, with no diagnostic. Minimal repro:

```python
from qamomile.qiskit.transpiler import QiskitTranspiler

@qmc.qkernel
def pair(a: qmc.Vector[qmc.Qubit], b: qmc.Vector[qmc.Qubit]) -> tuple[qmc.Vector[qmc.Qubit], qmc.Vector[qmc.Qubit]]:
    a = qmc.h(a)
    b = qmc.x(b)
    return a, b

@qmc.qkernel
def circuit() -> qmc.Vector[qmc.Bit]:
    q = qmc.qubit_array(4, "q")
    v = q[0:2]
    x, y = pair(v, v)  # same view twice — silently aliased
    q[0:2] = x
    return qmc.measure(q)

QiskitTranspiler().to_circuit(circuit, bindings={})
# before: emits h[0], h[1], x[0], x[1] — both formals aliased onto qubits 0,1, no error
```

With an entangling callee (`cx(a[0], b[0])`) it instead crashed deep in emit with a raw qiskit `CircuitError: 'duplicate qubit arguments'` and no Qamomile diagnostic.

## Root cause

`QKernel.__call__` defers `VectorView` consumption until after the call is recorded and keys `input_view_metas` by `(root, start, step, length)`, so two identical views collapse to a single entry and never trip the per-handle `consume()` check that plain `Qubit` / `Vector` arguments hit. The IR-level `AffineValidationPass` cannot catch it either, because `InlinePass` dissolves the `CallBlockOperation` (whose operands are the same value twice) before `affine_validate` runs in the pipeline.

## Fix

- **`QKernel.__call__`** rejects two quantum arguments backed by the same `Value` (compared by `uuid`, mirroring `composite_gate._reject_duplicate_targets`) before any handle is consumed, raising `QubitConsumedError` with affine-rule wording. A broadcast gate on a full view keeps the same backing `Value`, so a stale view version passed alongside its live successor collides here too (this previously emitted the caller's gates twice on the same wires).
- An already-consumed `VectorView` argument now raises the standard consumed error at the call boundary instead of slipping past the deferred-consumption loop.
- **`InlinePass`** rejects duplicate quantum operands on a `CallBlockOperation` as defense in depth for hand-built or deserialized IR (which never went through `QKernel.__call__`), keyed on `uuid` and raising before the call is dissolved.

## Tests

New `TestDuplicateQuantumCallArgs` in `tests/circuit/test_affine_types.py` covers the silent-alias and entangling-crash variants, stale-version aliasing, consumed-view arguments, the whole-`Vector` duplicate, and the tampered-IR inline path, and asserts disjoint views of one array still compile.

## Verification

- `uv run pytest` (default suite): 8561 passed, 409 skipped, 1 xfailed (the pre-existing `test_noncommuting_trotter_converges_with_steps` local segfault deselected, unrelated).
- `uv run ruff check qamomile/ tests/`, `uv run ruff format --check`, `uv run zuban check qamomile/`: all clean.
- `/local-review`: no remaining findings.

### DIFF
--- a/qamomile/circuit/frontend/qkernel.py
+++ b/qamomile/circuit/frontend/qkernel.py
@@ -46,6 +46,7 @@ from qamomile.circuit.ir.types import BitType, FloatType, ObservableType, UIntTy
 from qamomile.circuit.ir.value import ArrayValue, DictValue, Value
 from qamomile.circuit.transpiler.errors import (
     FrontendTransformError,
+    QubitConsumedError,
     QubitRebindError,
 )
 
@@ -117,6 +118,118 @@ def _promote_literal_to_handle(value: Any, expected_type: Any) -> Any:
         if is_bool:
             return bit(value)
     return value
+
+
+def _quantum_handle_display_name(handle: Handle) -> str:
+    """Return a human-readable name for a quantum handle in error messages.
+
+    Args:
+        handle (Handle): Handle to name. Falls back from the handle's own
+            name to its backing value's name to a shortened handle id.
+
+    Returns:
+        str: Non-empty display name for the handle.
+    """
+    return handle.name or handle.value.name or f"qubit_{handle.id[:8]}"
+
+
+def _reject_aliased_quantum_args(kernel_name: str, arguments: dict[str, Any]) -> None:
+    """Reject two call arguments backed by the same quantum register.
+
+    ``QKernel.__call__`` defers ``VectorView`` consumption until after the
+    call is recorded, so binding the same view to two parameters would
+    bypass the per-handle ``consume()`` check that plain ``Qubit`` /
+    ``Vector`` arguments hit, silently aliasing both formal registers onto
+    the same physical qubits.  This guard runs before any handle is
+    consumed and compares the backing ``Value.uuid``, mirroring
+    ``composite_gate._reject_duplicate_targets``.
+
+    Keying on ``uuid`` (a single ``Value`` instance bound twice) rather
+    than ``logical_id`` (stable across SSA versions) is deliberate and
+    complete.  The affine invariant forbids two *live* handles of one
+    wire, so two arguments that share a ``logical_id`` but not a ``uuid``
+    always include a stale (already-consumed) version, which the
+    consumption loop below rejects anyway — a stale view via
+    :func:`_reject_consumed_view_arg`, a stale whole register via
+    ``Handle.consume()``.  (A broadcast gate on a full view keeps the same
+    backing ``Value``, so a view and its post-gate successor collide on
+    ``uuid`` here directly.)
+
+    Args:
+        kernel_name (str): Name of the called kernel, used in the error
+            message.
+        arguments (dict[str, Any]): Bound call arguments keyed by parameter
+            name.  Non-``Handle`` entries and classical handles are
+            ignored.
+
+    Returns:
+        None.
+
+    Raises:
+        QubitConsumedError: If two quantum arguments are backed by the same
+            ``Value`` instance (the same live handle passed twice).
+    """
+    seen: dict[str, str] = {}
+    for name, handle in arguments.items():
+        if not isinstance(handle, Handle) or not handle._should_enforce_linear():
+            continue
+        backing_uuid = handle.value.uuid
+        first_name = seen.get(backing_uuid)
+        if first_name is not None:
+            display_name = _quantum_handle_display_name(handle)
+            raise QubitConsumedError(
+                f"Arguments '{first_name}' and '{name}' of "
+                f"'QKernel[{kernel_name}]' are backed by the same qubit "
+                f"register ('{display_name}').\n\n"
+                f"Affine type rule: Each qubit handle can be passed to a "
+                f"kernel call at most once — binding one register to two "
+                f"parameters would alias both onto the same physical "
+                f"qubits.\n\n"
+                f"Fix: pass disjoint registers, e.g.:\n"
+                f"  x, y = {kernel_name}(q[0:2], q[2:4])  # disjoint slices",
+                handle_name=display_name,
+                operation_name=f"QKernel[{kernel_name}]",
+            )
+        seen[backing_uuid] = name
+
+
+def _reject_consumed_view_arg(kernel_name: str, handle: Handle) -> None:
+    """Reject an already-consumed ``VectorView`` passed as a call argument.
+
+    View consumption is deferred until after the call is recorded, and the
+    deferred loop only consumes views that are still live — so a view that
+    was already consumed before the call would slip through without the
+    ``QubitConsumedError`` that a plain handle gets from ``consume()``.
+    This guard restores that error at the call boundary.
+
+    Args:
+        kernel_name (str): Name of the called kernel, used in the error
+            message.
+        handle (Handle): View argument to check.
+
+    Returns:
+        None.
+
+    Raises:
+        QubitConsumedError: If ``handle`` was already consumed.
+    """
+    if not handle._consumed:
+        return
+    display_name = _quantum_handle_display_name(handle)
+    raise QubitConsumedError(
+        f"Qubit view '{display_name}' was already consumed by "
+        f"'{handle._consumed_by}' and cannot be used again in "
+        f"'QKernel[{kernel_name}]'.\n\n"
+        f"Affine type rule: Each qubit handle can only be used once. "
+        f"After a gate operation, reassign the result to use the new "
+        f"handle.\n\n"
+        f"Fix:\n"
+        f"  v = qm.h(v)  # Reassign to capture the new handle\n"
+        f"  {kernel_name}(v)  # Pass the reassigned handle",
+        handle_name=display_name,
+        operation_name=f"QKernel[{kernel_name}]",
+        first_use_location=handle._consumed_by,
+    )
 
 
 def _const_int(value: Value | None) -> int | None:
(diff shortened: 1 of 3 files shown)


ksk-jij:
## Branch summary

### 1. Problem

In Qamomile a qkernel (the unit of a user-written quantum program, defined by decorating a Python function with the qkernel decorator) can be called from inside another qkernel like a function call. On such a call, passing the same quantum register to two arguments caused main to **silently bind both formal parameters onto the same physical qubits, with no error**.

Qamomile enforces an affine-type rule ("each quantum value may be used at most once", corresponding to the no-cloning theorem for qubits) and normally rejects this kind of double use at compile time. However, when a view into part of an array (`VectorView`, the handle obtained from a slice such as `q[0:2]`) was passed to two arguments of the same kernel call, the check was bypassed.

Reproduction on main:

```python
from qamomile.qiskit.transpiler import QiskitTranspiler

@qmc.qkernel
def pair(a: qmc.Vector[qmc.Qubit], b: qmc.Vector[qmc.Qubit]) -> tuple[qmc.Vector[qmc.Qubit], qmc.Vector[qmc.Qubit]]:
    a = qmc.h(a)
    b = qmc.x(b)
    return a, b

@qmc.qkernel
def circuit() -> qmc.Vector[qmc.Bit]:
    q = qmc.qubit_array(4, "q")
    v = q[0:2]
    x, y = pair(v, v)   # the same view is passed twice
    q[0:2] = x
    return qmc.measure(q)

QiskitTranspiler().to_circuit(circuit, bindings={})
```

Behavior on main: no error; it emits `h[0], h[1], x[0], x[1]`. Both `pair` formals `a` and `b` land on physical qubits 0,1, so a circuit that contradicts the user's intent (apply `h` and `x` to two separate registers) is generated silently.

It is even worse when the callee entangles qubits. For example, if the callee contains `cx(a[0], b[0])`, the failure surfaces deep in the final compilation stage (emit) as a raw Qiskit exception `CircuitError: 'duplicate qubit arguments'`. No readable Qamomile diagnostic is produced at all.

The root cause is in `QKernel.__call__` (`qamomile/circuit/frontend/qkernel.py:671` onward). This method is designed to defer "consumption" of view arguments (the affine-type bookkeeping that marks a spent quantum handle as used) until after the call has been recorded. Deferred views are registered in a dictionary `input_view_metas` (`qamomile/circuit/frontend/qkernel.py:783`) keyed by `(slice-source array, start, step, length)`; passing the identical view twice makes **both collapse onto the same key into a single entry**, so it slips past the per-argument consumption check.

There is also a structural reason the pipeline cannot catch it later. The pass that enforces the affine-type rule at the IR (intermediate representation) level is `affine_validate`, but the `inline` pass that runs before it (which expands and dissolves the `CallBlockOperation` — the IR node representing a kernel call — into the caller) removes the duplicated operands (the state where the same value appears twice) first, so by the time `affine_validate` runs the trace of the duplicate is gone.

This branch fixes that by detecting the duplicate at the call boundary and raising an affine-type-rule error.

### 2. Frontend (user-facing code level) changes

Passing the same quantum register to multiple arguments of one kernel call now raises `QubitConsumedError` (the existing exception class for affine-type-rule violations) at compile time (when `build()` / `to_circuit()` runs). The `pair(v, v)` in the reproduction above is rejected with a message like:

```
Arguments 'a' and 'b' of 'QKernel[pair]' are backed by the same qubit register ('q[slice]').

Affine type rule: Each qubit handle can be passed to a kernel call at most once — binding one register to two parameters would alias both onto the same physical qubits.

Fix: pass disjoint registers, e.g.:
  x, y = pair(q[0:2], q[2:4])  # disjoint slices
```

For an entangling callee too, instead of the raw Qiskit exception at emit time, the same `QubitConsumedError` is now raised at the tracing stage (very early in circuit construction). The problem is explained in Qamomile's own terms before reaching the backend.

Relatedly, the following two input patterns are now clearly rejected at the call boundary:

- **Passing an already-consumed view**: e.g. after consuming `v` with `v = q[0:2]; qmc.h(v)`, passing `pair(v, ...)` raises a `QubitConsumedError` at the call site saying `v` was already consumed by `H`. On main this became a confusing borrow error much later, or produced nothing at all when the callee dropped the view.
- **Mixing different versions of the same array**: the case `v = q[0:2]; v2 = qmc.h(v)` then `pair(v, v2)`. `v2` is the handle after applying `h` to `v`, but a gate applied to a full view keeps the same backing `Value`, so `v` and `v2` point at the same instance. This too is rejected by the "same register in two arguments" error above.

On the other hand, the legitimate usage of passing **non-overlapping slices** of the same array still works. For example, the following is not an error:

```python
@qmc.qkernel
def circuit() -> qmc.Vector[qmc.Bit]:
    q = qmc.qubit_array(4, "q")
    x, y = pair(q[0:2], q[2:4])   # disjoint slices: works as before
    q[0:2] = x
    q[2:4] = y
    return qmc.measure(q)
```

### 3. Backend (IR and overall) changes

The fix is two-layered — an early check at the frontend call boundary, and a defense-in-depth check in the IR pass — and both decide duplication by comparing the same `Value.uuid` (the unique identifier assigned to each value instance in the IR).

**Frontend side (`qamomile/circuit/frontend/qkernel.py`)**

- `_reject_aliased_quantum_args` (`qamomile/circuit/frontend/qkernel.py:136-194`): a new helper that scans all call arguments and, if two quantum handles share the same backing `Value.uuid`, raises `QubitConsumedError`. It is called inside `QKernel.__call__` before any handle is consumed (`qamomile/circuit/frontend/qkernel.py:723`). Classical arguments (`Float` / `UInt` / `Bit`) may be passed multiple times, so they are excluded.
- `_reject_consumed_view_arg` (`qamomile/circuit/frontend/qkernel.py:196-232`): a new helper that raises `QubitConsumedError` if an already-consumed `VectorView` appears as an argument. It is called for each view argument inside the loop that defers view consumption (`qamomile/circuit/frontend/qkernel.py:809`). Because that deferred loop only consumes views that are still live, an already-consumed view would slip through without this check; this closes that gap.
- `_quantum_handle_display_name` (`qamomile/circuit/frontend/qkernel.py:123-133`): a small helper that decides the display name of a quantum handle for error messages (falling back handle name → backing value name → shortened id). Shared by the two helpers above.

The key point is that `_reject_aliased_quantum_args` decides by `uuid`, not by `logical_id` (the logical identifier that is stable across SSA versions). By the affine-type rule, "two live handles of the same array cannot exist at the same time", so if two arguments share a `logical_id` but differ in `uuid`, at least one of them is an already-consumed older version. That case is rejected separately by the consumption check (`_reject_consumed_view_arg` for a view, the existing `Handle.consume()` for a whole register), so keying on `uuid` alone misses nothing.

**IR pass side (`qamomile/circuit/transpiler/passes/inline.py`)**

- Added a duplicate-operand check in `InlinePass._inline_call`, before the call node is dissolved (`qamomile/circuit/transpiler/passes/inline.py:392-417`). It scans the quantum operands of the call and, if the same `uuid` appears twice, raises `QubitConsumedError`. This addresses the problem that `inline` dissolves the `CallBlockOperation` (the IR node representing a kernel call) before `affine_validate` runs, and serves as defense in depth for IR built by hand or via deserialization without going through the frontend. IR produced through the frontend is already rejected by the `QKernel.__call__` check above.

This pass decides by `uuid` intentionally: different SSA versions of one wire are legitimately distinct values (distinct `uuid`) in the IR, and enforcing per-wire linearity is the frontend's job, so here only "the exact same value instance appearing twice" counts as a duplicate.

### 4. Alternatives considered and why this approach was chosen

- **Duplication key: `uuid` vs `logical_id`**. The frontend duplication check could also have been implemented with `logical_id`, the logical identifier that is stable across SSA versions. In the end both the frontend and IR-pass checks were unified on `uuid` (per value instance), for two reasons. First, the sibling existing guard `_reject_duplicate_targets` in `composite_gate` (which rejects duplicate gate targets) keys on `uuid`, so matching its criterion keeps the codebase consistent. Second, as noted above, keying on `uuid` misses nothing (the older-version case is handled by the consumption check), so we can lean to the safer side — `uuid` rejects only clear aliasing and is less prone to false positives. What was given up is the simplicity of "one guard covers all cases"; in exchange we adopted a "duplication check + consumption check" split of responsibilities that follows the existing design.

- **Which layer to reject at: frontend only / IR pass only / both**. Relying only on the IR pass (`affine_validate`) does not help this bug because `inline` dissolves the call node first, so it was not adopted. Placing it only in the frontend was also considered, but IR built by hand or via deserialization does not go through the frontend and would be unprotected. In the end we chose to emit the readable early error that directly affects the user experience in the frontend (`QKernel.__call__`), and to place a light check in `inline` as well as insurance for the path that constructs IR directly — adopting both.

### 5. Known limitations

- **The `inline` guard does not reject a mixed-in older version**: When — in hand-built / deserialized IR, if two different SSA versions of the same wire (different `uuid`, same `logical_id`) are passed to two operands of one call. Why — the `inline` check intentionally looks only at `uuid` equality, so a differing `uuid` passes through. Through the frontend this shape is never generated in the first place (a live handle can exist only once at a time), so it does not arise on the main path in practice. Future fix — if the affine check for hand-built IR is to be strengthened, there is room to add a separate `logical_id`-based check at a stage before `inline`. No follow-up is registered at this point (the path with real impact is extremely limited).

- **The case of mixing a whole array with a view into part of it is rejected by a different path**: When — passing both a whole array `q` and a partial view `q[0:2]` of it at the same time, as in `pair(q, q[0:2])`. Why — the whole array and the view have different `uuid`s, so this branch's duplication check does not fire; instead the existing borrow tracker raises `UnreturnedBorrowError` because it tried to consume an array that still has slots lent out to a view. It is still rejected, so there is no real harm, but the kind of error and its message differ from the affine diagnostic added here. Future fix — if unifying the message becomes necessary, there is room to detect the containment relation between whole and view at the call boundary and align it to the same wording; for now it is left to the existing mechanism.

PR #569, opened 2026-07-06
Preserve compile-time-constant block outputs through folding

ksk-jij:
## Summary

A qkernel that returns a compile-time-constant classical value as a block output previously returned `None` at runtime instead of the constant. This affected two shapes:

- a classical expression fully folded by `partial_eval`, e.g. `return x * 2.0` with `bindings={"x": 3.0}` (expected `6.0`, got `None`);
- a `bindings`-bound argument returned directly, e.g. `return x` with `x` bound (pass-through form).

Root cause: constant folding removes the producing operation and the output value becomes a compile-time constant, but `ProgramABI` kept only the output UUIDs, so `ProgramOrchestrator._resolve_outputs` found nothing in the execution context and returned `None`. The runtime-parameter path (`parameters=["x"]`) was already correct — its expression is not folded and materializes through a classical segment.

Fix:

- `ProgramABI` gains an `output_constants` map (`segments.py`).
- `NisqSegmentationStrategy._collect_output_constants` records outputs that are compile-time constants — scalars via `Value.is_constant()` / `get_const()`, const arrays via `ArrayValue.get_const_array()` (`separate.py`).
- `ProgramOrchestrator._resolve_outputs` falls back to that map, using `val is None` (not truthiness) so a genuine `0` / `0.0` / `False` constant is not mistaken for an unresolved output. It now raises `ExecutionError` when an output resolves to neither a context value nor a constant, surfacing pipeline materialization gaps instead of silently returning `None` (`program_orchestrator.py`).

This keeps the constant knowledge in the execution ABI and resolves it at output time, rather than keeping a redundant classical op in the IR to recompute a compile-time-known value — consistent with the project's IR-abstraction principle. The IR itself is unchanged; no new IR op or pass.

## Test plan

New `tests/circuit/test_bound_classical_output.py`, run across every supported local backend (Qiskit / QURI Parts / CUDA-Q) with `importorskip` guards, both `sample` and `run` paths:

- computed form (`return x * 2.0`) and pass-through form (`return x`), parametrized over `x` (including the `0.0` boundary and negative values) and register sizes `n ∈ {1, 2, 3}`;
- the runtime-parameter (`parameters=["x"]`) variant, asserting the compile-time-constant fallback does not shadow the context path;
- an assertion that the segmentation ABI records the folded scalar output as a constant while the measured Vector is not recorded;
- the new `ExecutionError` path for a genuinely unresolvable output.

Verified locally:

- `uv run pytest tests/` — 9377 passed, 4 skipped, 1 xfailed (no regressions).
- New tests pass on Qiskit and QURI Parts (CUDA-Q covered by CI).
- `uv run ruff check qamomile/ tests/`, `uv run ruff format --check qamomile/ tests/`, `uv run zuban check qamomile/` — all clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XKDEtZT7b552gsZbmE9DvG)_

### DIFF
--- a/qamomile/circuit/transpiler/passes/separate.py
+++ b/qamomile/circuit/transpiler/passes/separate.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import dataclasses
 from abc import ABC, abstractmethod
+from typing import Any
 
 from qamomile.circuit.ir.block import Block
 from qamomile.circuit.ir.operation import Operation
@@ -188,6 +189,7 @@ class NisqSegmentationStrategy(SegmentationStrategy):
         abi = ProgramABI(
             public_inputs={name: value for name, value in block.parameters.items()},
             output_refs=[v.uuid for v in block.output_values],
+            output_constants=self._collect_output_constants(block.output_values),
         )
 
         return ProgramPlan(
@@ -196,6 +198,46 @@ class NisqSegmentationStrategy(SegmentationStrategy):
             boundaries=boundaries,
             parameters=block.parameters,
         )
+
+    @staticmethod
+    def _collect_output_constants(
+        output_values: list[Value],
+    ) -> dict[str, Any]:
+        """Collect compile-time-known values among a block's outputs.
+
+        A block output whose value folded to a compile-time constant —
+        a ``bindings``-bound classical argument returned directly, or a
+        classical expression fully evaluated by ``partial_eval`` (e.g.
+        ``return x * 2.0`` with ``x`` bound) — has no producing operation
+        left in the IR. It therefore never lands in the runtime execution
+        context, and resolving it by UUID alone yields ``None``. Recording
+        the constant here lets ``ProgramOrchestrator._resolve_outputs``
+        return the correct value.
+
+        Non-constant outputs (measurement results, runtime-parameter
+        expressions, arrays materialized by classical stores) are left out;
+        they resolve from the execution context at run time.
+
+        Args:
+            output_values (list[Value]): The block's return values, in
+                order, as produced by ``materialize_return``.
+
+        Returns:
+            dict[str, Any]: Map from output value UUID to its constant
+                Python value, containing only the outputs that are
+                compile-time constants.
+        """
+        constants: dict[str, Any] = {}
+        for value in output_values:
+            if not isinstance(value, ValueBase):
+                continue
+            if isinstance(value, ArrayValue):
+                const_array = value.get_const_array()
+                if const_array is not None:
+                    constants[value.uuid] = tuple(const_array)
+            elif value.is_constant():
+                constants[value.uuid] = value.get_const()
+        return constants
 
 
 class SegmentationPass(Pass[Block, ProgramPlan]):
--- a/qamomile/circuit/transpiler/program_orchestrator.py
+++ b/qamomile/circuit/transpiler/program_orchestrator.py
@@ -465,7 +465,38 @@ class ProgramOrchestrator(Generic[T]):
     # ------------------------------------------------------------------
 
     def _resolve_outputs(self, context: ExecutionContext) -> Any:
-        """Read final output values from execution context."""
+        """Read final output values from the execution context.
+
+        Each output ref is resolved in order:
+
+        1. A direct context entry (the common case — a measured bit,
+           a runtime-parameter expression, or a value materialized by a
+           classical segment).
+        2. Per-element composite carrier keys ``"<ref>_<i>"`` reassembled
+           into a tuple (a measured Vector's bits).
+        3. The compile-time ``output_constants`` map from the program ABI
+           — for outputs that folded to a constant during compilation
+           (e.g. a ``bindings``-bound classical argument returned directly,
+           or ``return x * 2.0`` with ``x`` bound). These have no producing
+           operation left in the IR, so they never enter the context.
+
+        Args:
+            context (ExecutionContext): Execution context populated by the
+                quantum measurement load and any classical segments.
+
+        Returns:
+            Any: The single output value when the program returns one
+                output, or a tuple of output values otherwise.
+
+        Raises:
+            ExecutionError: If an output ref resolves to nothing through
+                any of the paths above, indicating a compiler or runtime
+                invariant was violated rather than a legitimate ``None``.
+        """
+        output_constants = (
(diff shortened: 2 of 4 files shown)


ksk-jij:
## Branch summary: preserve-block-outputs-through-folding

### Problem overview

This fixes a bug where a classical value that gets folded to a compile-time constant returns `None` at runtime when it is used as a kernel return value (a block output) instead of the actual constant. A "block output" is a value a `QKernel` (a function decorated with the qkernel decorator) returns; the `ProgramPlan` (the execution plan of the compiled program) holds each output as a list of UUIDs in `output_refs`. Behavior on main: for example, transpiling `y = x * 2.0; return y, bits` with `bindings={"x": 3.0}` causes `partial_eval` (the pipeline stage that includes constant folding and bakes `bindings`-bound values such as `x` into the IR) to fold `x*2.0` into the constant `6.0` and remove that `BinOp` (the binary-operation IR node) from the IR. The folded output value itself carries the constant `6.0`, but the execution plan's ABI (`ProgramABI`) keeps only the output UUIDs, so the constant is lost. At run time `ProgramOrchestrator._resolve_outputs` (which assembles return values from the execution context) looks the UUID up in the execution context, but because folding removed the producing operation nothing was ever stored there, so that output resolves to `None`. The pass-through form `return x` (returning a bound argument directly) has the same root cause and also returns `None`. Behavior on branch: the same kernel returns the correct value for compile-time-constant outputs through both `run` and `sample` (each element of the `SampleResult`) — `6.0` in the example above, `3.0` for the pass-through form. Falsy constants such as `x=0.0` are correctly distinguished from `None` and return `0.0`. The `parameters=["x"]` form (keeping `x` as a runtime parameter) does not fold `x*2.0` and materializes it at run time through a classical segment, so it was already correct, and this fix does not change that path. Reproduction and root cause: reproduce with `transpile(kernel, bindings={"x": 3.0})` followed by `run`/`sample`, where the first element comes back as `None`. The cause splits into two parts: the folded constant output is not carried into the execution plan's ABI (`qamomile/circuit/transpiler/segments.py`, `qamomile/circuit/transpiler/passes/separate.py`), and `_resolve_outputs` has no fallback path beyond the execution context (`qamomile/circuit/transpiler/program_orchestrator.py`).

### Frontend (user-facing code) changes

There is no change to the syntax or API signatures a user writes with the qkernel decorator. What changes is the correctness of the results: a compile-time-constant return value that previously came back as `None` now returns the correct value.

```python

@qmc.qkernel
def k(x: qmc.Float) -> tuple[qmc.Float, qmc.Vector[qmc.Bit]]:
    y = x * 2.0
    qs = qmc.qubit_array(2, name="qs")
    qs = qmc.h(qs)
    bits = qmc.measure(qs)
    return y, bits

transpiler = QiskitTranspiler()
exe = transpiler.transpile(k, bindings={"x": 3.0})
first, bits = exe.run(transpiler.executor()).result()
# main: first == None  /  branch: first == 6.0
```

In addition, when a block output can be resolved from neither the context nor the constants, `_resolve_outputs` now raises an `ExecutionError` (the run-time error type) instead of silently returning `None`. This is a diagnostic improvement that surfaces a materialization gap in the pipeline to the user and has no effect on well-formed programs.

### Backend (IR and overall) changes

This change touches only the transpiler's segmentation stage and the execution-orchestrator layer, and is independent of any backend (Qiskit / QURI Parts / CUDA-Q). No IR node or pass is added. It adds an output-constants map `output_constants: dict[str, Any]` to `ProgramABI` (field definition at `qamomile/circuit/transpiler/segments.py:183`, class docstring at `segments.py:161-179`): a map from output UUID to compile-time constant value, the metadata needed to resolve at run time a constant output that has no producing operation. Non-constant outputs — measurement results, runtime-parameter expressions, and the like — are not placed here and are resolved from the execution context as before. It adds `NisqSegmentationStrategy._collect_output_constants` (`qamomile/circuit/transpiler/passes/separate.py:203`), called by `create_plan` while building the ABI (`separate.py:192`). It scans the output values finalized by `materialize_return` (which syncs `block.output_values` from the `ReturnOperation`) and collects scalars via `Value.is_constant()`/`get_const()` and constant arrays via `ArrayValue.get_const_array()`. Output values folded to constants already carry the constant at this point via `ReturnOperation` operand substitution, so no extra folding is needed. It adds a constant fallback and an explicit error to `ProgramOrchestrator._resolve_outputs` (`qamomile/circuit/transpiler/program_orchestrator.py:467`). The resolution order is (1) direct execution-context lookup, (2) reassembly of measured-vector elements in `"<ref>_<i>"` form, (3) `plan.abi.output_constants` (the map is read at `program_orchestrator.py:496-498`, the fallback at `:511-513`). If none of these resolve, it raises an `ExecutionError` (`program_orchestrator.py:515-521`). When `plan` is `None` (the plan-less execution path), it falls back to an empty map. Using `val is None` rather than truthiness for the constant check ensures a `0`/`0.0`/`False` constant output is not mistaken for an unresolved output.

### Alternatives considered and rationale

Alternative A (rejected): keep, in the constant-folding pass, a block-output `BinOp` in the IR just like array stores and materialize it at run time in a classical segment. However, under compile-time `bindings`, after folding its operand (e.g. `x`) is no longer present in the run-time context, so keeping the operation does not resolve at run time and would require an extra step to bake the operand into a constant. Moreover, the `return x` pass-through form (which has no `BinOp` at all) cannot be rescued this way, so an orchestrator-side change would still be needed separately. Alternative B (adopted): aggregate, once, the compile-time constants the output values already carry into the ABI and fall back to them at output-resolution time. This handles the folded and pass-through paths with a single mechanism, leaves no redundant classical operation in the IR, and follows the design principle of keeping the IR abstract and deferring concretization to a later stage (output resolution) — Qamomile's IR Abstraction Principle in CLAUDE.md. Array-store outputs still materialize through a classical segment as before, and `output_constants` is consulted only when context resolution fails, so there is no risk of mistaking a measured vector. For these reasons B was adopted.

### Known limitations

None. Both compile-time-constant block outputs (scalar constants and constant arrays) and non-constant outputs materialized at run time are resolved by a single mechanism, and unresolvable outputs are surfaced via `ExecutionError`, so there is no remaining gap after merge.

---

ksk-jij:
This PR is for old Qamomile and has already fixed at the latest main. Thus, I'll close this PR.

PR #571, opened 2026-07-07
Redesign callable invocation model

yuyamashiro:
## Summary

Redesign the composite-gate callable surface and compiler representation around a unified callable invocation model.

- Add the public `composite_gate` and `opaque`/`Oracle` frontend paths while keeping `CallableDef` and `InvokeOperation` internal compiler-facing concepts.
- Replace legacy call/composite IR operation paths with `InvokeOperation`, including serialization, canonicalization, printing, resource estimation, lowering, and visualization support.
- Refactor the large qkernel implementation into focused frontend modules.
- Update stdlib QFT/QPE, arithmetic, Grover, and Shor handling together with body-derived symbolic resource estimation and regression coverage.

## Motivation

The previous composite-gate implementation required too much user-side ceremony and split lowering/resource behavior across several mechanisms. This change keeps the frontend concepts smaller while giving the compiler one consistent call representation for qkernels, named composite gates, stdlib callables, and opaque oracles.

Documentation changes are split into the stacked docs-only PR #584.

## Validation

- `uv run pytest -q`


yuyamashiro:
## What this PR enables from a user's perspective

This redesign keeps the frontend centered on `QKernel` while giving the compiler one callable representation internally. A helper qkernel, a named composite, and an opaque oracle now compose through the same call path without exposing compiler-only definitions to users.

### 1. Define a named composite with one decorator

A composite is still a `QKernel`: it supports normal calls, drawing, inversion, control, transpilation, and resource estimation. The decorator can be applied directly; a second qkernel decorator is not required.

```python

@qmc.composite_gate(name="repeated_h")
def repeated_h(q: qmc.Qubit, rounds: qmc.UInt) -> qmc.Qubit:
    for _ in qmc.range(rounds):
        q = qmc.h(q)
    return q

@qmc.qkernel
def algorithm(rounds: qmc.UInt) -> tuple[qmc.Qubit, qmc.Qubit]:
    control = qmc.qubit("control")
    target = qmc.qubit("target")

    target = repeated_h(target, rounds)
    target = qmc.inverse(repeated_h)(target, rounds)
    control, target = qmc.control(repeated_h)(control, target, rounds)
    return control, target
```

All three calls refer to the same named callable. A backend may preserve the box or lower its body, while users continue to work with the same `QKernel` interface.

### 2. Derive resources from the algorithm body

Body-backed qkernels and composites no longer need a separately maintained resource model. Resource estimation interprets the actual program body, including symbolic loops, callable transforms, register sizes, and compile-time branches.

```python
symbolic = algorithm.estimate_resources()
print(symbolic.gates.total)  # 3*rounds

concrete = algorithm.estimate_resources(inputs={"rounds": 3})
print(concrete.gates.total)  # 9
```

The `inputs={...}` argument uses qkernel input names directly. It can specialize both symbolic sizes and compile-time control flow; partial inputs leave exact symbolic expressions such as `Piecewise` rather than forcing an approximation.

### 3. Model only genuinely opaque work explicitly

When no implementation body exists, an opaque oracle can carry an explicit fixed or context-dependent cost. The surrounding qkernel is still analyzed normally, and oracle calls and query complexity remain visible separately.

```python
oracle = qmc.opaque(
    "query_oracle",
    num_qubits=1,
    cost=qmc.ResourceEstimate(
        gates=qmc.GateResources(total=40, t=40, non_clifford=40),
        calls=qmc.CallResources(
            calls_by_name={"query_oracle": 1},
            queries_by_name={"query_oracle": 1},
        ),
    ),
)

@qmc.qkernel
def search_step() -> qmc.Qubit:
    (q,) = oracle(qmc.qubit("q"))
    return q

estimate = search_step.estimate_resources()
print(estimate.gates.t)                            # 40
print(estimate.calls.queries_by_name["query_oracle"])  # 1
```

This keeps explicit resource metadata at the boundary where it is actually needed: a bodyless operation.

### 4. Use one Shor implementation for estimation and execution

The stdlib order-finding implementation is an executable qkernel built from Qamomile arithmetic and QFT components. Its resources are derived from that same body, rather than from a handwritten Shor resource formula.

```python
import sympy as sp

order_finding = qmc.shor_order_finding(base=2, modulus=15)

symbolic = order_finding.estimate_resources()
n = sp.Symbol("n", integer=True, positive=True)
assert symbolic.qubits == 6 * n + 5
assert sp.Poly(symbolic.gates.total, n).degree() == 3
assert sp.Poly(symbolic.gates.toffoli, n).degree() == 3

concrete = order_finding.estimate_resources(inputs={"n": 4})
assert concrete.qubits == 29

# The same API also produces an executable program. Use a small instance for
# local state-vector sampling.
small = qmc.shor_order_finding(base=2, modulus=3)
transpiler = QiskitTranspiler()
executable = transpiler.transpile(small, bindings={"n": 2})
result = executable.sample(transpiler.executor(), shots=128).result()
```

The distinction is intentional: `inputs` specializes resource analysis, while `bindings` fixes compile-time values for transpilation.

### Additional behavior covered by the redesign

- Helper qkernel calls inline for ordinary code reuse; composites remain named calls until the selected backend lowering.
- Direct, inverse, and controlled forms use the same callable definition and resource source.
- Logical estimates report gate categories, depth, width/ancilla usage, call/query counts, assumptions, and optional traces.
- Measurement consumes a qubit in the affine model; `project_x`, `project_y`, and `project_z` represent projection when the post-measurement quantum state must remain available.
- Invalid overlapping `bindings` and runtime `parameters`, qubit aliasing, unresolved loop-indexed measurements, and misspelled estimation inputs now fail explicitly instead of silently changing program meaning.
- The callable and stdlib paths are exercised across Qiskit, QuriParts, and CUDA-Q; the complete docs execution suite also passes (157 passed, with only the two credential-dependent qBraid pages skipped).

PR #597, opened 2026-07-16
Add semantic SELECT multiplexers and value controls

ksk-jij:
## Summary

- Add `qmc.select` as a semantic quantum multiplexer with inferred, explicit, and compile-time symbolic index widths.
- Add value-activated coherent controls so zero-valued control bits lower through X bracketing while preserving Qamomile's LSB-first register convention.
- Preserve SELECT through CircuitProgram lowering as a reusable semantic call with a portable controlled-case fallback, including global phase, inverse, enclosing controls, nesting, and identity states.
- Extend serialization, validation, resource estimation, visualization, HUGR/Quration boundaries, documentation, and regression coverage.

## Motivation

SELECT is a core primitive for block encoding, qubitization, and LCU-style algorithms. Keeping it semantic until target legalization allows native implementations where available while retaining one backend-neutral fallback and avoiding an early expansion into backend-shaped gates.

## API behavior

The first flattened index qubit is bit zero. An omitted width is inferred from the case count; a wider explicit width leaves unassigned index states as identity; a symbolic `UInt` width is resolved and checked during transpilation. Every case has one shared target and classical-parameter signature so the selected branch remains a coherent unitary on the same quantum footprint.

## Testing

- `uv run pytest`: 9086 passed, 4 skipped, 2948 deselected, 1 xfailed
- Focused SELECT/control/serialization/CircuitProgram/HUGR/Quration suite: 269 passed
- Cross-backend execution: Qiskit, QuriParts, CUDA-Q, HUGR, and Quration paths exercised according to installed target support


ksk-jij:
# Semantic SELECT Multiplexers and Value-Activated Controls

## 1. Problem Overview

This branch adds the quantum multiplexer `qmc.select`, which applies different unitaries to the same target according to the value of an index quantum register, and `qmc.control(..., control_value=...)`, which fires only for a specified computational-basis value. For a SELECT with width `k` and `n` cases, the semantics are `Σ_{i=0}^{n-1} |i><i| ⊗ U_i + Σ_{i=n}^{2^k-1} |i><i| ⊗ I`. SELECT is a fundamental building block for LCU, or linear combinations of multiple unitaries, and block encoding, which embeds a matrix as a submatrix of a larger unitary.

On `origin/main`, users had to write the multi-qubit controls for each case manually, including the X gates placed before and after control qubits that must be zero. With that representation, users had to keep case numbering, unassigned states, phases, and nested controls consistent themselves, while the compiler could not recognize that the circuit represented SELECT. This branch preserves SELECT as a single semantic IR operation. The IR is the intermediate representation through which the compiler preserves circuit meaning, and concretization into a form accepted by each target is deferred until the backend-neutral circuit representation.

The index convention is LSB-first, consistent with the rest of Qamomile. LSB-first means that the first qubit after flattening is interpreted as integer bit 0; for two qubits, case number 2 therefore has the activation pattern `(0, 1)`. The number of cases does not need to be a power of two, and index values without a corresponding case apply the identity operation.

Users create a selector with `qmc.select([U_0, U_1, ...], num_index_qubits=...)`. When the width is omitted, `ceil(log2(number of cases))` is used; when a wider concrete width is specified, the additional states apply the identity operation. A `qmc.UInt` width can be resolved later as a compile-time binding, meaning a value supplied through `transpile(..., bindings=...)`. At invocation, the leading `Qubit`, `Vector[Qubit]`, which represents a qubit array, and `VectorView[Qubit]`, which references part of an array, are treated as index arguments until their combined width matches the required width; the remaining quantum arguments become the targets shared by all cases.

```python


@qmc.qkernel
def apply_xz_select(
    signal: qmc.Qubit,
    system: qmc.Qubit,
) -> tuple[qmc.Qubit, qmc.Qubit]:
    signal = qmc.h(signal)
    signal, system = qmc.select([qmc.x, qmc.z])(signal, system)
    signal = qmc.h(signal)
    return signal, system


@qmc.qkernel
def circuit() -> tuple[qmc.Bit, qmc.Bit]:
    signal = qmc.qubit("signal")
    system = qmc.qubit("system")
    signal, system = apply_xz_select(signal, system)
    return qmc.measure(signal), qmc.measure(system)
```

The `H–SELECT(X, Z)–H` sequence above is a minimal block encoding whose block projected onto signal state `|0⟩` is `(X + Z) / 2`. Cases may be qkernels, qkernel-backed composite gates, or built-in gates such as `qmc.x`. Every case must have an identical signature, including argument names, types, order, and default values, and must return the same quantum targets in the same order. SELECT misuse raises an early exception when fewer than two cases are supplied, a concrete width cannot represent all cases, a case contains measurement, reset, or internal qubit allocation, or a case drops or reorders targets. This contract and its errors are defined in `qamomile/circuit/frontend/operation/select.py:391` and the public function at `qamomile/circuit/frontend/operation/select.py:619`.

`qmc.control` gains a single integer `control_value`. For example, `qmc.control(qmc.x, num_controls=2, control_value=2)` fires when the first flattened control is 0 and the second is 1. `None` and the all-ones value preserve the conventional behavior in which every control must be 1, while negative values, values that do not fit the width, and booleans are rejected. The activation value is preserved through inverse operations, outer controls, composite gates, and the concrete-width Oracle path, but a symbolic control width cannot be combined with a fixed `control_value`. The public contract is defined at `qamomile/circuit/frontend/operation/control.py:2766`. English and Japanese Tutorial 04 intentionally cover only `qmc.control`, using executable examples to explain `control_value`, LSB-first flattening, power, global phase, scalar-to-Vector broadcast, symbolic widths, and failure cases. The tutorial does not include SELECT API documentation or a SELECT demonstration.

The backend goal is to preserve SELECT and value-activated controls as meaningful structures while integrating them with the existing CircuitProgram, target capabilities, legalization, and materialization pipeline. CircuitProgram is a circuit IR independent of backend SDKs; target capabilities declare which semantic operations a target accepts; legalization chooses a native or fallback implementation according to those declarations; and materialization is the final step that creates concrete objects such as Qiskit circuits from the legalized circuit. In the following discussion, lowering means transforming a higher-level semantic representation into the next, more concrete representation.

### Recording Semantic Operations and Activation Values

The first goal is to retain information that later stages can recognize as SELECT instead of prematurely expanding each case into a low-level gate sequence. Reducing SELECT to ordinary multi-controlled gates would lose case boundaries, case order, and the additional identity states, preventing later selection of target-specific native implementations. `SelectOperation` therefore stores the width, the number of index-argument groups, and ordered `case_blocks` in a single node (`qamomile/circuit/ir/operation/select.py:29`). An operation-owned Block is an independent block that owns its formal arguments and return values, unlike a loop body that directly shares the parent value scope.

Activation values remain integers in the IR rather than being expanded into X-gate sequences. `normalize_control_value` canonicalizes the all-ones integer to `None`, ensuring that semantically equivalent ordinary controls always share the same internal representation. `control_pattern_for_value` converts the value into an LSB-first sequence of zero and one bits (`qamomile/circuit/ir/operation/control_value.py:8`, `qamomile/circuit/ir/operation/control_value.py:51`). This representation is shared by controlled gates, controlled invocation, inverse blocks, and Oracles, and forms the single activation-value contract consumed by the subsequent concretization step.

### Safe Case Composition

The next goal is to support built-in gates and qkernels through the same SELECT API while guaranteeing that every selected case is a unitary over the same wires. A naive implementation that derives result wiring only from case 0 would silently miswire another case that discards a target and returns a newly allocated qubit. `SelectGate` reuses the existing control-call preparation logic to split indices and targets under the same convention, then specializes each case to the invocation shape (`qamomile/circuit/frontend/operation/select.py:391`).

`_validate_case_target_footprint` compares logical IDs position by position between inputs and outputs; a logical ID identifies the same logical qubit even when SSA updates replace its value object (`qamomile/circuit/frontend/operation/select.py:230`). A logical ID is not a physical-qubit allocation index. SSA is an IR form in which every update produces a new value version. `_validate_case_operations_are_unitary` additionally traverses nested calls and rejects measurement, reset, expectation values, and internal allocation (`qamomile/circuit/frontend/operation/select.py:294`). The safe case blocks produced here become the input to CircuitProgram lowering.

### Deferred Concretization in CircuitProgram

The goal at this stage is to preserve SELECT semantics for target selection while also providing a fallback circuit that works with every circuit-family backend. Semantics alone would not be executable on current targets, while retaining only the fallback gate sequence would prevent future native selection, so both are required. `CircuitLoweringPass._emit_select` creates an independent `CircuitProgram` for each case and turns it into a `ReusableCircuit`, a reusable subcircuit callable from multiple locations, whose control count equals the index width (`qamomile/circuit/transpiler/circuit_ir/lowering.py:134-264`). The outer operation remains a semantic call carrying `SELECT_SEMANTIC_KEY`, meaning a call with a semantic identifier, so legalization can replace it with a native implementation in the future (`qamomile/circuit/transpiler/circuit_ir/model.py:642`).

In the fallback circuit, the case number itself becomes `control_value`, and `bracket_control_value` surrounds only the controls that require zero with X gates (`qamomile/circuit/transpiler/passes/emit_support/control_value_emission.py:16`). SELECT, ordinary control, inverse, composite, and outer-control paths all use this shared helper, preventing zero-valued controls from being lost because of nesting position. Each case’s global phase, the component that applies the same phase to every state inside the case, also remains attached to the independent CircuitProgram and therefore becomes a relative phase instead of being lost under SELECT control. The check that retains a case with no operations but a nonzero phase is located at `qamomile/circuit/transpiler/circuit_ir/lowering.py:228`.

When a case accepting a scalar qubit is broadcast over a qubit array, meaning that the same body is applied to every array element, the complete scalar unitary is applied once per element. Broadcasting `U = exp(iφ)V` over `N` elements means `U⊗N = exp(iNφ)V⊗N`, matching an explicit element loop. Separating the phase so that it is applied only once would make different representations of the same scalar unitary, such as `X` and `exp(iπ/2)RX(π)`, produce different operations after broadcast, so that design was rejected. The public API documents this contract (`qamomile/circuit/frontend/operation/control.py:2803-2808`, `qamomile/circuit/frontend/operation/select.py:630-635`).

SELECT lowering invokes the complete case `ReusableCircuit`, including its phase, once for each target element (`qamomile/circuit/transpiler/circuit_ir/lowering.py:239-248`), while ordinary `qmc.control` similarly emits the complete controlled scalar gate or fallback once per element (`qamomile/circuit/transpiler/passes/emit_support/controlled_emission.py:2731`). Because power applies `U**power` to each element and inverse applies `U†` to each element, their phases are respectively `N * power * φ` and `-Nφ`. By contrast, a case such as the LCU term `exp(iθ_j)P_j`, which needs one phase for the register as a whole, is expressed by writing `P_j` in a qkernel whose argument is itself `Vector[Qubit]` and wrapping that entire vector body with `qmc.global_phase`. This is one logical call rather than a broadcast, so the phase is preserved exactly once, including for an identity Pauli case.

### Consistency Across Transformation Passes and Serialization

The goal at this stage is to prevent SELECT-owned Blocks from being mistaken for ordinary nested bodies and to ensure that no compiler pass skips the contents of a case. If an existing pass traversed only operations in the parent scope, compile-time `if` operations, calls, and value substitutions could remain inside cases and fail only during lowering.

The inline pass replaces calls with their callee bodies, while partial evaluation precomputes branches and expressions from bound values; both explicitly descend into case blocks (`qamomile/circuit/transpiler/passes/inline.py:191`, `qamomile/circuit/transpiler/passes/partial_eval.py:132`).

Compile-time if lowering selects branches determined at compile time, substitution replaces specified calls with alternative implementations, and canonicalization normalizes representations without changing their meaning so they can be compared. These transformations likewise preserve the same ownership boundary while processing cases.

For serialization and restoration, `_encode_select` and `_decode_select` round-trip references to concrete or symbolic widths, index-argument groups, and case blocks (`qamomile/circuit/ir/serialize/encode.py:1304`, `qamomile/circuit/ir/serialize/decode.py:1780`). Protobuf is a structured binary format used to store the IR, and `control_value` values too wide for fixed-width integers are stored as `BigInteger` messages containing a sign and arbitrary-length magnitude (`qamomile/circuit/serialization/proto/qamomile_ir.proto:272`, `qamomile/circuit/serialization/proto/qamomile_ir.proto:390`). `_validate_select` rejects corrupted widths, result wiring, and case input/output contracts after restoration (`qamomile/circuit/serialization/validation.py:991`).

### Analysis and Explicit Target Boundaries

The final goal is to expose SELECT structure correctly to non-execution consumers and to prevent unsupported targets from silently producing incorrect circuits. The resource estimator analyzes gate counts and depths as expressions; `eval_select` evaluates each case under the index width and outer controls and also accounts for broadcast by repeating scalar cases over every element of a qubit array (`qamomile/circuit/estimator/resource_estimator.py:2827`). Visualization renders SELECT as a single box and displays zeros in `control_value` as open control points and ones as filled control points (`qamomile/circuit/visualization/analyzer.py:1755-1775`, `qamomile/circuit/visualization/renderer.py:1309`).

Visualization may be unable to resolve some control lines to physical wires, meaning qubit lines in the rendered diagram. Reinterpreting the original `control_value` using a width shorter than the declared width would then cause an out-of-range error or an incorrect anti-control display. `_control_pattern_for_resolved_wires` therefore uses the LSB-first pattern only when every line is resolved; when the widths differ, it safely draws the resolved lines as ordinary filled controls (`qamomile/circuit/visualization/analyzer.py:1424-1457`).

The circuit-family paths for Qiskit, QuriParts, CUDA-Q, and Quration can legalize and materialize the semantic call’s fallback circuit. In contrast, the PreparedModule path, which preserves reusable calls and structured control flow and lowers them directly to HUGR without splitting them into CircuitPrograms, cannot yet represent SELECT. HUGR is an intermediate representation that models programs as typed graphs, and this direct lowering explicitly rejects SELECT with `EmitError` (`qamomile/hugr/lowerer.py:606-613`). This prevents unsupported behavior from being hidden behind an incorrect fallback.

Expanding SELECT into per-case X gates and multi-controlled operations in the frontend or semantic IR was rejected. Although it would reduce the amount of implementation code, SELECT semantics would be lost before reaching CircuitProgram, making native selection through target capabilities, structure-aware visualization, and per-case resource estimation impossible. A single `SelectOperation` with late lowering follows Qamomile’s principle that the IR describes what an operation does while the target decides how to realize it.

Representing a quantum index with an ordinary `IfOperation` was also rejected. `IfOperation` represents runtime control flow based on measurement results or classical values, meaning classical branching during execution, whereas SELECT coherently chooses among multiple unitaries without collapsing a superposition. Owning cases as operation-owned Blocks allows transformation passes to descend into permitted compile-time `if` and `for` constructs while unitary validation rejects runtime `if` and `while` constructs that require measurement or HYBRID operations.

Expanding zero-valued controls into X gates at API invocation was rejected. Preserving `control_value` as semantic data allows a future backend with a `ctrl_state`-like capability to materialize it natively without X bracketing. The current shared fallback can also be centralized in one helper, ensuring the same convention under inverse, power, broadcast, and outer control.

Making SELECT specially accept only `GlobalPhaseGate` was rejected. Adding exceptions for individual combinators would fragment the contract among qkernels, composites, and gate-like objects, so cases follow the general qkernel contract, and phase-only cases are expressed through an explicit qkernel wrapper. The same wrapper can be reused by internal LCU helpers.

Rejecting Vector broadcast of phase-bearing scalar bodies and separating the phase so that it is applied only once were both rejected. Qamomile’s existing broadcast contract is syntactic sugar for an explicit loop that applies the complete scalar unitary to every element, so the tensor product including phase is the correct meaning. Consequently, no special path needs to reject operations based on the presence or value of a phase node, and no defensive code is required to special-case zero phase. A register-wide phase is made explicit with a Vector-signature qkernel, using the type signature to distinguish its meaning from scalar broadcast.

## 5. Known Limitations

When: Cases need different classical argument names, types, default values, or different numbers of runtime parameters. Why: SELECT currently forwards one invocation binding to every case in the same order and therefore requires all cases to have exactly the same signature. Future fix: Design a general `qmc.bind`-like combinator as a separate feature that binds some qkernel arguments to case-specific values and produces qkernels with a shared signature.

When: A `GlobalPhaseGate` returned by `qmc.global_phase(kernel, theta)` is passed directly in the case list. Why: To avoid a SELECT-specific exception, the current combinator contract is limited to qkernels and supported callables, so passing it directly raises `TypeError`. Future fix: Instead of special-casing only `GlobalPhaseGate`, define a general QKernelLike contract shared by all combinators such as control, inverse, and select, under which an object provides input types, invocation, and IR construction like a qkernel. Until then, use a small qkernel wrapper that invokes the phase gate.

When: A SELECT case internally allocates a clean ancilla, meaning a temporary auxiliary qubit returned to `|0⟩` after computation. Why: A case block is defined to own only a unitary over the shared targets, and lowering has no representation for allocating and releasing case-owned resources, so an internal `QInitOperation` is rejected early. Future fix: Express the number of ancillas, clean or dirty requirements, and lifetimes required by a case block as a resource contract that target capabilities and allocators can satisfy safely.

When: `num_index_qubits` remains a runtime parameter supplied only during execution, or a symbolic `num_controls=qmc.UInt` is combined with a fixed `control_value`. Why: The circuit structure, number of controls per case, and valid range of activation values must be known before materialization, and the current implementation resolves widths only as compile-time bindings. Future fix: Extend the contract only if backend capabilities and a dynamic resource model that genuinely support runtime-variable widths are introduced.

When: Direct HUGR PreparedModule lowering is used for a kernel containing SELECT. Why: This path bypasses CircuitProgram legalization and does not yet implement the quantum-multiplexer semantics of SELECT on a HUGR graph, so it is explicitly unsupported. In addition, the currently shipped circuit-family targets do not register native implementations of semantic SELECT; Qiskit, QuriParts, CUDA-Q, and Quration use the portable fallback. Future fix: Add structured SELECT lowering for HUGR and, when an SDK provides an appropriate native multiplexer, register a `SELECT_SEMANTIC_KEY` target capability and legalization rule.

PR #600, opened 2026-07-17
docs(algorithm): add Grover search tutorial documentation

jij-yuichiro-nakano:
## Summary

Adds bilingual documentation for Grover search as an algorithm tutorial.

## What Was Added

- Added Grover search tutorials under `docs/en/algorithm/grover_search.py` and `docs/ja/algorithm/grover_search.py`, with synced and executed notebooks.
- Explained phase-oracle marking, diffusion, amplitude amplification, and Grover iteration selection.
- Added a five-qubit example using four search qubits and one reusable oracle auxiliary qubit to find the marked state `|0101>`.
- Implemented the phase oracle as `oracle_operator` using Qamomile's multi-controlled X gate.
- Compared search-register probability histograms before and after Grover search.
- Added oracle-query resource estimation from 4 to 16 search qubits, comparing Grover's quadratic speedup with exhaustive search on a base-10 logarithmic scale.
- Added Grover search entries to the English and Japanese algorithm index pages.

## Test Plan

- `./docs/build.sh page-build docs/en/algorithm/grover_search.py docs/ja/algorithm/grover_search.py`
- `git diff --check`
- `uv run pytest tests/docs/test_tutorials.py -m docs -k 'algorithm/grover_search' -v`
  - Currently fails with 56 errors in `qamomile/hugr/lowerer.py`; the same errors are reproducible on `origin/main` and are unrelated to this documentation change.


docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,428 @@
...
+        for state, count in sample_result.results
+    }
+
+
+before_probabilities = state_probabilities(before_result)
+after_probabilities = state_probabilities(after_result)
+
+basis_states = [
+    format(index, f"0{SEARCH_QUBITS}b") for index in range(2**SEARCH_QUBITS)
+]
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Thank you for pointing this out. Fixed in `8d169d8a`. `state_probabilities()` now reverses the LSB-first measurement tuple when producing the conventional `|q3 q2 q1 q0>` display label, so `q0` is the rightmost bit. The oracle now flips `search[1]` and `search[3]`, which correctly marks the displayed state `|0101>`. The English and Japanese notebooks were rebuilt and executed.

docs/en/algorithm/grover_search.py
```diff
@@ -0,0 +1,503 @@
...
+        for state, count in sample_result.results
+    }
+
+
+before_probabilities = state_probabilities(before_result)
+after_probabilities = state_probabilities(after_result)
+
+basis_states = [
+    format(index, f"0{SEARCH_QUBITS}b") for index in range(2**SEARCH_QUBITS)
+]
```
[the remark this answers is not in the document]
jij-yuichiro-nakano: Thank you for pointing this out. Fixed in `8d169d8a`. `state_probabilities()` now reverses the LSB-first measurement tuple when producing the conventional `|q3 q2 q1 q0>` display label, so `q0` is the rightmost bit. The oracle now flips `search[1]` and `search[3]`, which correctly marks the displayed state `|0101>`. The English and Japanese notebooks were rebuilt and executed.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#     language: python
+#     name: python3
+# ---
+
+# %% [markdown]
+# ---
+# tags: [algorithm, oracle-based]
+# ---
+#
+# # Grover探索入門
```
ksk-jij: [my opinion]
I think we can omit "入門". Also, I wonder if Grover探索 is better than Groverの探索アルゴリズム or Groverのアルゴリズム.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# Grover探索は、確率振幅を増幅することで、構造を持たない探索空間からマークされた要素を見つけるアルゴリズムです{cite:p}`10.1145/237814.237866`。サイズ$N$の探索空間にマークされた要素が1つある場合、古典的な全探索では$O(N)$回の問い合わせが必要ですが、Grover探索では$O(\sqrt{N})$回のオラクル問い合わせで探索でき、古典計算に対して2次的な計算速度の加速を示すことが知られています。
```
ksk-jij: [my opinion]
"マークされた要素" feels like the element has already marked before we perform the algorithm. I suppose that the element (the quantum state) will be marked through the algorithm.  So, different word would be nice.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# このノートブックでは、Qamomileの組み込みヘルパー`grover_search`を使用し、Grover探索をQamomileで実装する方法を紹介します。4量子ビットの検索レジスタからマークされた状態$|0101\rangle$を探索し、位相オラクル内で補助量子ビットを1つ使う5量子ビット回路を構成します。最後に、振幅増幅の前後における検索レジスタの状態確率をヒストグラムで比較します。
```
ksk-jij: [my opinion]
We can use another word for レジスタ? Like "4量子ビットからなる量子状態"?

ksk-jij: [my opinion]
The sentence flow feels a bit weird? How about "具体的には、4量子ビットからなる量子状態|0101\rangle$を探索し、最後にアルゴリズム実行前後での量子状態の観察し、Groverのアルゴリズムによる確率振幅の変化を確認します．" or something like that?

ksk-jij: [question]
Sorry for my ignorance, but is "位相オラクル" a common word?

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+
+# %%
+# 最新のQamomileをpipからインストールします！
+# # !pip install qamomile
```
ksk-jij: [must]

To use, qiskit, we need to install qiskit. Also, visualisation as well?

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+
+# %%
+# 数値計算、プロット、シミュレータ、Qamomileのユーティリティを読み込みます。
```
ksk-jij: [my opinion]
We can omit this? In the other documents, I believe there is no similar comment.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+
+import qamomile.circuit as qmc
+from qamomile.qiskit import QiskitTranspiler
+
+transpiler = QiskitTranspiler()
+
+# %% [markdown]
+# ## アルゴリズム
+#
+# $x_\star\in\{0,1\}^n$を、1つだけマークされた$n$量子ビットの対象状態とします。探索空間の大きさを$N=2^n$とし、$f(x)$は$x=x_\star$の場合にのみ1、それ以外では0になる関数とします。対応する位相オラクルは、この条件を相対位相として表します。
```
ksk-jij: [question]
Sorry for my ignorance, but is $x_\star\in\{0,1\}^n$ common (like the star itself and the position)?

ksk-jij: [my opinion]
"1つだけマークされた" feels like the state has already been marked, but I believe the state will be marked through the algorithm. So, how about " $x_\star\in\{0,1\}^n$ を見つけ出したい量子状態とします．" or something like that?

ksk-jij: [my opinion]
"対応する位相オラクルは、この条件を相対位相として表します。" feels like the explanation is for one who knows the algorithm. How about "$x \in \{0 ,1 \}$に対して$f(x)$を$x=x_\star$の場合にのみ1、それ以外では0になる関数とします。また，オラクル$O_f$を以下のように定義します" or something like that?

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# $$
+# O_f|x\rangle = (-1)^{f(x)}|x\rangle.
+# $$
+#
+# したがって、オラクルは$|x_\star\rangle$を$-|x_\star\rangle$へ変化させ、ほかの基底状態は変化させません。位相だけでは測定確率は変化しません。Grover探索では位相オラクルと一様重ね合わせに関する反転を組み合わせ、対象状態の確率振幅を増幅します。
```
ksk-jij: [my opinion]
If we employ what I wrote (or similar one) above, then "オラクルは" feels a bit redundant. Also, "オラクルは" can be taken as stuff call ``oracle'' is always so. To avoid those, how about "$O_f$" or "オラクル$O_f$" instead of "オラクル"?

ksk-jij: [my opinion]
"位相だけでは測定確率は変化しません。Grover探索では位相オラクルと一様重ね合わせに関する反転を組み合わせ、対象状態の確率振幅を増幅します。" is a bit abrupt as Japanese. How about "Groverアルゴリズムではこのオラクルと後述する拡散演算子と呼ばれる操作を繰り返すことで与えられたオラクル$O_f$でマークされる状態の確率振幅を増幅させます．"?

ksk-jij: [my opinion]
Because we haven't clarified what we mean by "マークする". So, how about adding "この$|x_\star\rangle$をマイナス倍する操作を移行$|x_\star\rangle$をマークするといいます" or something like that after "したがって、オラクルは$|x_\star\rangle$を$-|x_\star\rangle$へ変化させ、ほかの基底状態は変化させません。"?

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# :::{note}
+# **オラクルとは？**
+#
+# 量子アルゴリズムにおけるオラクルとは、入力が条件を満たすかを判定する処理を、可逆な量子演算として表したものです。アルゴリズムからは内部実装に依存せず使えるため、ブラックボックスとして扱われます。これは実装が不要という意味ではなく、実際のオラクル回路は探索問題ごとに構成します。Grover探索では位相オラクルを使い、判定結果を別の量子ビットへ出力する代わりに、条件を満たす状態の符号を反転します。オラクルを1回適用することを、1回のオラクル問い合わせと数えます。
```
ksk-jij: [question]
Is "量子アルゴリズムにおけるオラクルとは、入力が条件を満たすかを判定する処理を、可逆な量子演算として表したものです。" always so? Like Qamomile has `Oracle` as a class for quantum operation without actual implementation. I personally wonder if this sentence makes the definition of oracle narrow. Or our `Oracle` is actually wierd.

ksk-jij: [my opinion]
How about "このため，オラクルを適用した回数をオラクルへの問い合わせとして数え，(量子ゲート数などとは異なる)計算量の指標として扱います。一方で、実際にアルゴリズムを適用するためにはオラクル部分も問題毎に固有に実装する必要があります．特にGroverの探索アルゴリズムでは見つけ出したい量子状態に合わせてオラクルを実装します．" instead of "これは実装が不要という意味ではなく、実際のオラクル回路は探索問題ごとに構成します。Grover探索では位相オラクルを使い、判定結果を別の量子ビットへ出力する代わりに、条件を満たす状態の符号を反転します。オラクルを1回適用することを、1回のオラクル問い合わせと数えます。"?

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# を準備します。対象状態が1つの場合、$|\mathrm{good}\rangle=|x_\star\rangle$とし、残りの$N-1$個の基底状態の正規化された重ね合わせを$|\mathrm{bad}\rangle$と定義します。
+#
+# $$
+# |\mathrm{bad}\rangle
+# = \frac{1}{\sqrt{N-1}}
+# \sum_{x\ne x_\star}|x\rangle.
+# $$
+#
+# 初期状態は次のように表せます。
```
ksk-jij: [my opinion]
I feel adding "すると” or similar conjunction would be nice.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+# \theta=\arcsin\frac{1}{\sqrt{N}}.
+# $$
+#
+# ### ステップ2：位相オラクルで対象状態をマークする
+#
+# 位相オラクルは、次の符号反転を適用します。
+#
+# $$
+# O_f|x_\star\rangle=-|x_\star\rangle.
+# $$
```
ksk-jij: [my opinion]
This has been already explained when we defined $O_f$ above. I feel showing the quantum state after applying $O_f$ would be nice rather than explaining $O_f$ again.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# $$
+# |s\rangle
+# = \cos\theta|\mathrm{bad}\rangle
+# + \sin\theta|\mathrm{good}\rangle,
+# \qquad
+# \theta=\arcsin\frac{1}{\sqrt{N}}.
+# $$
+#
+# ### ステップ2：位相オラクルで対象状態をマークする
```
ksk-jij: [question]
I think, as at least the naive way of explaining Grover (like in textbooks), there is a step to prepare $\ket{-}$ as another qubit to perform phase kickback. Indeed, we use five qubits in this notebook as well. Could I ask you why we skip that step here?

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# 位相オラクルは、次の符号反転を適用します。
+#
+# $$
+# O_f|x_\star\rangle=-|x_\star\rangle.
+# $$
+#
+# オラクルはGrover探索の問題に依存する部分です。対象状態だけをマークする回路、またはより一般に探索条件を満たす状態をマークする回路を用意し、一時的に使った補助量子ビットはオラクルを終了する前に元の状態へ戻す必要があります。
```
ksk-jij: [my opinion]
I feel we don't need to mention "対象状態だけをマークする回路、またはより一般に探索条件を満たす状態をマークする回路を用意し、" since we don't say anything in "ステップ1".

ksk-jij: [my opinion]
Is it alright not to say "一時的に使った補助量子ビットはオラクルを終了する前に元の状態へ戻す必要があります。"? I think this is not only for this case, but a general explanation of uncomputation not for only Grover. As a notebook for Grover and fixed oracle coming up, I feel we can omit this.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# ### ステップ3：一様重ね合わせ状態に関して反転する
+#
+# オラクルの後に拡散演算子
+#
+# $$
+# D=2|s\rangle\langle s|-I
+# $$
+#
+# を適用し、すべての確率振幅を平均値に関して反転します。したがって、1回のGrover反復は
```
ksk-jij: [question]
Sorry for my ignorance, I didn't really get what "すべての確率振幅を平均値に関して反転します" means. I mean I know quite small part of the geometric explanation of Grover, but like, what "平均値" is, what "反転" is (like probability amplitude is always positive, in this case, what is "反転" here)?

ksk-jij: [my opinion]
"したがって" does not make sense for me? We haven't mentioned normal iteration or Grover iteration until now. We should introduce this as one Grover iteration or something like that.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# $$
+# G=DO_f
+# $$
+#
+# です。$|\mathrm{bad}\rangle$と$|\mathrm{good}\rangle$が張る2次元平面では、オラクルと拡散演算子による2つの反転を組み合わせることで、対象状態へ向かって$2\theta$だけ回転します。
```
ksk-jij: [question]
Is "対象状態へ向かって$2\theta$だけ回転します。" true? I thought Grover iteration just rotates the state without seeing the direction, like if you perform the iteration unsuitable times, I thought the state will move away from the target state.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+# + \sin((2r+1)\theta)|\mathrm{good}\rangle
+# $$
+#
+# です。したがって、対象状態を測定する確率は
+#
+# $$
+# P_r=\sin^2((2r+1)\theta)
+# $$
+#
+# です。$P_r$が最初に最大になるのは、累積した角度$(2r+1)\theta$が最適な角度$\pi/2$に最も近づくときです。Qamomileの`grover_iteration_count(n, 1)`ヘルパーを使うと、検索量子ビット数から適切な整数の反復回数を自動的に計算できます。それ以上反復すると、対象状態から確率振幅が離れることがあります。
```
ksk-jij: [my opinion]
"Qamomileの`grover_iteration_count(n, 1)`ヘルパーを使うと、検索..." This is not about algorithm but about qamomile. I think we can mention this later not here.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# ### ステップ5：検索レジスタを測定する
+#
+# 選んだ回数だけ反復した後に測定すると、高い確率で$x_\star$が得られます。組み込みの`qmc.grover_search`関数は一様重ね合わせを準備し、指定された回数だけオラクルと拡散演算子を反復します。呼び出し側は問題に依存する位相オラクルを渡し、返された検索レジスタを測定します。
```
ksk-jij: [my opinion]
"適切な回数" instead of "選んだ回数" might be better?

ksk-jij: [my opinion]
This part is about quite Qamomile not algorithm.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# まとめると、Grover探索は次の手順で実行します。
```
ksk-jij: [my opinion]
How about creating "### ステップまとめ" or something like that?

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+# 4. 拡散演算子を適用し、対象状態の確率振幅を増幅します。
+# 5. オラクルと拡散演算子を選んだ回数だけ繰り返します。
+# 6. 検索レジスタを測定します。
+
+# %% [markdown]
+# ## Qamomileでの実装：`grover_search`関数
+#
+# ### 問題設定
+#
+# ここからは、一般のアルゴリズムを4量子ビットの検索レジスタへ適用します。そのため$N=2^4=16$であり、$|0101\rangle$だけをマークします。Grover探索を実行する前は、一様重ね合わせに含まれる各状態の確率が$1/16$です。マークされた状態が1つの場合、`grover_iteration_count(4, 1)`は反復回数3を返します。3回反復した後、マークされた状態を測定する理想的な確率は約$0.9613$です。基底状態のラベルは通常の$|q_3q_2q_1q_0\rangle$の順に表示し、`search[0]`を右端の最下位ビット（LSB）として扱います。
```
ksk-jij: [my opinion]
"ここからは、一般のアルゴリズムを4量子ビットの検索レジスタへ適用します。そのため$N=2^4=16$であり、$|0101\rangle$だけをマークします。" feels a bit weird.

ksk-jij: [my opinion]
I think we can simply omit "マークされた状態が1つの場合、`grover_iteration_count(4, 1)`は反復回数3を返します。3回反復した後、マークされた状態を測定する理想的な確率は約$0.9613$です。".

ksk-jij: [my opinion]
We defined `serach` variable yet. We can move this explanation the part where we need this explanation.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# 5番目の量子ビットはオラクルの補助量子ビットです。オラクルを適用する前後で$|0\rangle$となるため、Grover反復ごとに同じ物理量子ビットを再利用できます。
```
ksk-jij: [my opinion]
We haven't defined anything yet, I think we can move this where needed.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+print("Grover iterations:", ITERATIONS)
+print("ideal marked-state probability:", f"{ideal_marked_probability:.6f}")
+
+assert ITERATIONS == 3
+assert ideal_marked_probability > 0.96
+
+# %% [markdown]
+# ### 位相オラクル
+#
+# 以下のオラクルでは、マルチCNOTゲートによる位相マーキングを使います。この例では、専用の補助量子ビットを対象として使います。Xゲートとアダマールゲートで$|-\rangle$を準備し、マルチCNOTゲートの位相キックバックによって`|0101>`をマークします。最後にアダマールゲートとXゲートを適用し、補助量子ビットを$|0\rangle$へ戻します。位相キックバック部分の前後にあるXゲートは、`0101`の0であるビットを1へ写した後、もとの基底ラベルへ戻します。
```
ksk-jij: [my opinion]
I think we don't need to mention 位相キックバック here? Like "XゲートとマルチCNOTゲートを使って`|0101>`をマークします。" or something like that. Then, we can modify the last sentence as well.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+    search[3] = qmc.x(search[3])
+    return search
+
+
+oracle_operator.draw(search=SEARCH_QUBITS, fold_loops=False)
+
+# %% [markdown]
+# ### Grover探索の量子カーネル
+#
+# `uniform_superposition`は、Grover探索を適用する前の初期状態を取得します。Grover探索用の回路と同じように4つの検索量子ビットと1つの補助量子ビットを確保しますが、一様重ね合わせを作って測定するのは検索レジスタだけです。
```
ksk-jij: [question]
Do we need this here? This is not about Grover algorithm, but for result section. I feel it would be better to move this sentence and its definition to the position where this is needed.

docs/ja/algorithm/grover_search.py
ksk-jij, line 1: [question]
Do you think adding naive implementation of Grover algorithm without the function needs more than one or two cells? If it requires only one cell we might be able to add.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+    return qmc.measure(search)
+
+
+# 事前計算した反復回数に基づいて、Grover反復を繰り返します。
+@qmc.qkernel
+def grover_search_example(
+    iterations: qmc.UInt,
+) -> qmc.Vector[qmc.Bit]:
+    search = qmc.qubit_array(SEARCH_QUBITS, name="search")
+    search = qmc.grover_search(search, oracle_operator, iterations)
```
ksk-jij, line 282: [question]
Can we use `qmc.Oracle` here and insert actual implementation at transpilation stage? Sorry we haven't added easy way to insert that, but you can do that with TranspilerConfig.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+
+
+oracle_operator.draw(search=SEARCH_QUBITS, fold_loops=False)
+
+# %% [markdown]
+# ### Grover探索の量子カーネル
+#
+# `uniform_superposition`は、Grover探索を適用する前の初期状態を取得します。Grover探索用の回路と同じように4つの検索量子ビットと1つの補助量子ビットを確保しますが、一様重ね合わせを作って測定するのは検索レジスタだけです。
+#
+# `grover_search_example`は、`oracle_operator`と反復回数を`qmc.grover_search`へ渡します。`iterations`は回路構造を決める`qmc.UInt`入力として残し、トランスパイル時に具体値を設定します。これにより、ループ内にはオラクル本体が1つだけ配置されます。Pythonの呼び出しを展開して反復ごとに別の補助量子ビットを確保する必要がなく、3回の反復で同じ補助量子ビットを再利用できます。
```
ksk-jij: [my opinion]
Does this mean in Qamomile with this implementation the oracle will be appeared in `ForOperation` only once? If so, since the users don't see the implementation (for loop in the code), we can omit this part?

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+        )
+    )
+    return executable.sample(executor, shots=SHOTS).result()
+
+
+before_result = sample_kernel(uniform_superposition, seed=SAMPLER_SEED)
+after_result = sample_kernel(
+    grover_search_example,
+    bindings={"iterations": ITERATIONS},
+    seed=SAMPLER_SEED + 1,
```
ksk-jij: [question]
Really just curiosity, why did you change the seed from one defined?

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+assert max(after_probabilities, key=after_probabilities.get) == MARKED_STATE
+assert after_probabilities[MARKED_STATE] > 0.85
+
+print("sampled marked-state probability:", f"{after_probabilities[MARKED_STATE]:.6f}")
+print("ideal marked-state probability:", f"{ideal_marked_probability:.6f}")
+
+# %% [markdown]
+# 1つ目のヒストグラムでは、各状態の確率が一様分布の$1/16$に近くなっています。オラクルと拡散演算子を3回繰り返すと、ほぼすべての確率が`|0101>`に集中し、$\sin^2((2r+1)\theta)$から予測される理想値に近づきます。サンプリングによって、厳密な確率から小さなずれが生じます。
+#
+# どちらのヒストグラムにも補助量子ビットは含めていません。補助量子ビットはオラクルを適用する前後で$|0\rangle$となり、位相キックバックによって4量子ビットの検索レジスタへオラクルの負符号を付けるためだけに使われます。
```
ksk-jij: [my opinion]
It might be nice to move "補助量子ビットはオラクルを適用する前後で$|0\rangle$となり、位相キックバックによって4量子ビットの検索レジスタへオラクルの負符号を付けるためだけに使われます。" to where we deifned the ancilla.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+
+# %% [markdown]
+# ## リソース推定
+#
+# Grover反復では、位相オラクルを1回呼び出します。したがって、位相オラクルへの問い合わせ回数は`grover_iteration_count`が返す反復回数と一致します。ここでは、問題サイズとGrover探索における位相オラクルの問い合わせ回数の関係を見てみましょう。
```
ksk-jij: [my opinion]
In this notebook, "位相オラクルへの問い合わせ回数は`grover_iteration_count`が返す反復回数と一致します" is true, but if you set another integer to the transpiler, it is not true I suppose.

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+
+# %% [markdown]
+# マークされた状態が1つの場合、`grover_iteration_count(n, 1)`によって選ばれる回数は、
+#
+# $$
+# r(n)=\left\lfloor\frac{\pi}{4}\sqrt{2^n}\right\rfloor
+# =\Theta(2^{n/2})
+# $$
+#
+# です。マークされた状態が1つの場合、$\theta=\arcsin(1/\sqrt{2^n})\approx1/\sqrt{2^n}$です。主要項の係数$\pi/4$は、最適な角度$\pi/2$を、Grover反復1回あたりの回転角$2\theta$で割ることから得られます。対数プロットでは、Grover探索の$\Theta(2^{n/2})$回の問い合わせと、古典的な全探索の$\Theta(2^n)$回の問い合わせを比較します。したがって、この比較はオラクル問い合わせ計算量について、古典的な全探索に対する2次加速を示しています。
```
ksk-jij: [question]
Sorry for my ignorance, what you mean by "主要項"? I understand that you are mentioning "\left\lfloor\frac{\pi}{4}\sqrt{2^n}\right\rfloor", but for this term, there is only one term, so what is "主要" here?

ksk-jij: [question]
"対数プロットでは、Grover探索の$\Theta(2^{n/2})$回の問い合わせと、古典的な全探索の$\Theta(2^n)$回の問い合わせを比較します。したがって、この比較はオラクル問い合わせ計算量について、古典的な全探索に対する2次加速を示しています。" feels a bit weird?
1. this feels like you haven't done it and you will, but you have already shown the plot above.
2. the comparison is indeed about 2次加速 since the data is prepared so, but as we prepared data so, I feel like this plot is not "showing" the speed up?

docs/ja/algorithm/grover_search.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# :::{note}
+# この比較では、位相オラクルを1回適用することを1回の問い合わせとして数え、オラクルの内部回路は考慮していません。具体的な位相オラクルでは、1回の適用に複数のゲートや補助量子ビットが必要になることがあります。そのため、Grover探索では、問題に応じたオラクルの設計が重要となります。
```
ksk-jij: [my opinion]
"そのため、Grover探索では、問題に応じてゲート数や量子ビット数を考慮したオラクルの設計が重要となります。" might be a bit kinder than now?

ksk-jij (commented):
@jij-yuichiro-nakano Hi! Thank you for the PR, I've left comments, please have a look at them.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you for raising this question. “Phase oracle” is established terminology in the Grover search and amplitude-amplification literature for an operator of the form $O_f|x\rangle=(-1)^{f(x)}|x\rangle$. Although earlier foundational papers also use terms such as phase inversion or reflection, later peer-reviewed work uses “phase oracle” explicitly. I would therefore like to keep the term, with its operation defined immediately at first use as it is now.

jij-yuichiro-nakano: Thank you. I replaced “search register” with “search qubits” throughout the tutorial. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I replaced the wording that implied the state had already been marked with “the state we want to find” throughout the introduction. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I removed the introductory qualifier from both the page title and the algorithm index card. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I rewrote the introduction to state the concrete target, introduce the from-scratch and standard-library implementations, and explain that the tutorial compares their output probabilities and query-count scaling. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I updated the installation command to use `qamomile[qiskit,visualization]`, covering the backend and visualization dependencies used in the tutorial. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I added the explicit definition of the Boolean function $f$ and the corresponding phase oracle $O_f|x\rangle=(-1)^{f(x)}|x\rangle$. Fixed in `c028f5dc`.

jij-yuichiro-nakano: Agreed. I now introduce the state as the state we want to find and use “mark” only after defining it as the phase sign flip performed by the oracle. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I removed the redundant comment explaining the imports. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I revised the explanation so that the phase oracle $O_f$ is explicitly the operation that changes the target state’s sign. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I retained the conventional $x_\star$ notation, but now define it explicitly as the target $n$-bit string before using it. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I now define “marking” explicitly as multiplying the target state’s amplitude by $-1$ while leaving the other basis states unchanged. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I clarified that query complexity counts oracle invocations independently of the oracle’s internal gate count, while an executable implementation still requires a problem-specific circuit whose cost must be considered separately. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I clarified that the oracle’s sign flip alone does not change measurement probabilities, and that combining it with the diffusion operator amplifies the target-state amplitude. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I broadened the explanation of an oracle to a problem-specific operation that can be invoked without depending on its internal implementation, and then introduced the phase oracle as the specific oracle used here. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I added a transition so that the consequence of preparing the uniform superposition follows naturally from the preceding definition. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I now define the mean amplitude $\bar{a}$ and express inversion about the mean as $a_x\mapsto2\bar{a}-a_x$, while noting that amplitudes may be negative or complex. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I removed the generic uncomputation statement from the algorithm section and kept the reversal explanation local to the concrete oracle implementation. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I removed the repeated oracle definition from Step 2 and instead show the transformed state after applying it: $O_f|s\rangle=\cos\theta|\mathrm{bad}\rangle-\sin\theta|\mathrm{good}\rangle$. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I clarified that the algorithmic description treats $O_f$ as an abstract phase oracle, while the later implementation realizes it by preparing an auxiliary qubit in $|-\rangle$ and using phase kickback. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I removed the sentence about a generic target-only problem-condition circuit because the concrete oracle construction is explained later. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I moved the `grover_iteration_count` helper out of the algorithmic explanation and introduced it with the concrete problem setup and implementation. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I now explicitly define one application of the phase oracle followed by the diffusion operator as a Grover iteration before discussing repeated applications. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I revised Step 3 to state that each Grover iteration rotates the state vector by $2\theta$ in a fixed direction, rather than claiming that every iteration moves it closer to the target. I kept Step 4 because it already explains that further iterations after the first maximum can move the amplitude away from the target. Addressed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I removed the Qamomile-specific helper explanation from the algorithm section and moved it to the implementation section. Fixed in `c028f5dc`.

jij-yuichiro-nakano: Thank you. I revised the step to say that the Grover iteration is repeated an appropriate number of times before measurement. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I simplified the problem setting to state directly that the search uses four qubits, has $N=16$ candidate states, and targets $|0101\rangle$. Fixed in `c028f5dc`.

jij-yuichiro-nakano: Agreed. I removed the prose that duplicated the helper result and ideal probability; those values remain visible in the executable output and assertions. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. Rather than adding another heading, I removed the duplicated step summary because it repeated the algorithm steps immediately above. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I moved the bit-order explanation to the point where measurement results are converted into state probabilities. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I moved the auxiliary-qubit explanation to the phase-oracle definition, where the qubit is introduced and prepared. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I added a from-scratch implementation of the diffusion operator and Grover search, and the tutorial now compares it with the standard-library implementation. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I removed the explanation of the internal loop operation and oracle body because those implementation details are not needed to understand or use the tutorial. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I moved the definition and explanation of `uniform_superposition` to the execution and results section where it is used as the baseline. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I simplified the concrete construction to describe the mapping of $|0101\rangle$ to $|1111\rangle$, preparation of the auxiliary qubit, application of the multi-controlled operation, and reversal of the preprocessing. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you for the suggestion. I considered this approach, but kept the concrete qkernel in this tutorial. `qmc.Oracle` is an opaque callable without a body and would require a `TranspilerConfig` substitution before execution, adding compiler configuration and complicating direct circuit drawing and resource estimation. Since this page teaches a concrete phase-oracle circuit and compares from-scratch and standard-library Grover implementations, passing the executable oracle qkernel directly keeps that relationship visible. Oracle substitution is better suited to a more advanced tutorial on reusable compiler configuration. No change was made for this suggestion.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I clarified that the observed oracle-query count matches the helper result because this notebook uses that result directly as the number of Grover iterations. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I removed the unexplained seed offset, and all compared executions now use the same `SAMPLER_SEED`. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I rewrote the discussion in retrospective terms and clarified that the plotted curves are reference scalings that illustrate the quadratic difference in query complexity, rather than empirical evidence of a speedup from the constructed data alone. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I consolidated the auxiliary-qubit explanation at its definition in the phase-oracle section and removed the duplicated explanation from the results section. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Agreed. I removed the ambiguous “leading term” wording and describe the query-count scaling directly. Fixed in `c028f5dc`.

docs/ja/algorithm/grover_search.py
jij-yuichiro-nakano: Thank you. I added a note that practical oracle design must consider both the gate count required to implement the oracle and the number of auxiliary qubits it uses. Fixed in `c028f5dc`.

jij-yuichiro-nakano:
@ksk-jij Thanks for the feedback! I've addressed the comments and pushed the changes. Ready for another review.

PR #601, opened 2026-07-17
Fix Aer empty multiplexer execution

ksk-jij:
## Summary
- recursively detect Aer-unsafe empty-parameter multiplexers inside Qiskit control-flow blocks before execution
- decompose only empty-parameter multiplexers, leaving valid params-bearing multiplexers untouched
- document the Aer-specific executor-side lowering in LIMITATIONS.md

## Testing
- uv run pytest tests/transpiler/backends/test_qiskit.py -q
- env git diff --check
- uv run zuban check qamomile/ (fails in existing qamomile/hugr/lowerer.py imports/stubs unrelated to this change)

### DIFF
--- a/LIMITATIONS.md
+++ b/LIMITATIONS.md
@@ -124,6 +124,16 @@ Element-granularity consumption (`if sel: _ = qmc.measure(qs[0])` followed by re
 
 **Future fix**: materialize the native inverse as a directly controllable Qiskit gate, or decompose the annotated inverse before controlling the reusable circuit.
 
+## Qiskit Aer execution may apply an extra lowering pass for empty-parameter multiplexers
+
+**When it bites**: a Qiskit circuit sent through `QiskitExecutor` targets Qiskit Aer and, after Qiskit transpilation, contains a `multiplexer` instruction with an empty `params` list. This shape can be produced by Qiskit's native state-preparation inverse path, including Qamomile's native Qiskit realization of concrete amplitude encoding followed by an inverse inside retained control flow. A params-bearing `multiplexer` is a valid Aer instruction and is left unchanged.
+
+**Why this trade-off was chosen**: Aer treats the operation name `multiplexer` as a native instruction and reads its unitary matrices from `operation.params`. Qiskit can also produce a generic gate named `multiplexer` whose matrices live only in the gate definition, leaving `params` empty. Passing that empty-parameter form to Aer can crash native assembly. To keep Qamomile's public Qiskit circuit as high-level as possible while avoiding an Aer crash at execution time, `QiskitExecutor` recursively inspects the Aer-transpiled circuit, including nested control-flow blocks, and decomposes only empty-parameter multiplexers immediately before `backend.run`.
+
+**Workaround**: if exact Qiskit-native state-preparation structure is not needed, `QiskitTranspiler(use_native_composite=False)` avoids the native `StatePreparation` path and emits Qamomile's explicit fallback decomposition instead. Users who build Qiskit circuits manually can also transpile against a backend target that does not include `multiplexer`, such as a suitable fake backend target, before executing on Aer.
+
+**Future fix**: remove the executor-side lowering once Qiskit and Aer agree on a memory-safe representation for inverse-derived multiplexers, either by keeping the generic empty-parameter form away from Aer-native basis instructions or by having Aer reject/decompose it through the gate definition instead of entering native assembly.
+
 ## `qmc.inverse` requires trace-time-resolvable control flow
 
 **When it bites**: `qmc.inverse(layer)(...)` is traced for a `layer` qkernel whose quantum body contains control flow the eager inverse builder cannot resolve while constructing the fallback `implementation_block`:
--- a/qamomile/qiskit/transpiler.py
+++ b/qamomile/qiskit/transpiler.py
@@ -69,12 +69,17 @@ class QiskitExecutor(QuantumExecutor["QuantumCircuit"]):
         """Execute circuit and return bitstring counts.
 
         Args:
-            circuit: The quantum circuit to execute
-            shots: Number of measurement shots
+            circuit (QuantumCircuit): Qiskit circuit to execute.
+            shots (int): Number of measurement shots.
 
         Returns:
-            Dictionary mapping bitstrings to counts (e.g., {"00": 512, "11": 512).
-            A circuit without quantum or classical bits returns ``{"": shots}``.
+            dict[str, int]: Dictionary mapping bitstrings to counts. A circuit
+            without quantum or classical bits returns ``{"": shots}``.
+
+        Raises:
+            RuntimeError: If no Qiskit backend is available for execution, or
+                if Aer would still receive an empty-parameter multiplexer after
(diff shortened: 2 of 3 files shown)


ksk-jij:
## 1. Problem Overview

On `origin/main`, Qiskit Aer execution could crash the Python process when a Qiskit circuit contained a `multiplexer` instruction with an empty `params` list. A `multiplexer` is Qiskit's representation for a uniformly controlled operation, meaning one of several unitary matrices is selected by control qubits. The crash is not caused by all multiplexers: Aer can execute a normal params-bearing multiplexer, but it can crash when Qiskit gives Aer an instruction named `multiplexer` whose matrices are stored only in the gate definition rather than in `operation.params`.

The common Qamomile-facing path is concrete amplitude encoding materialized through Qiskit's native `StatePreparation`, followed by an inverse. Qiskit's inverse state-preparation path can leave an empty-parameter `multiplexer` after Aer-target transpilation. The previous Qamomile workaround only checked top-level circuit instructions, so it missed the same unsafe instruction when Qiskit kept it inside a `for_loop` control-flow block. A control-flow block is a nested Qiskit circuit body that executes under a structured operation such as a loop; the crash therefore survived when Grover-style repeated work stayed nested instead of being unrolled.

This branch changes `QiskitExecutor.execute()` so the Aer-specific workaround looks recursively through nested Qiskit control-flow blocks and only decomposes empty-parameter multiplexers before calling `backend.run`. Decomposition means replacing a higher-level instruction with lower-level Qiskit operations according to the instruction's definition. The branch preserves valid params-bearing multiplexers and preserves retained `for_loop` structure, so it fixes the crash path without treating every multiplexer as unsupported.

There is no new user-facing Qamomile API. Users can continue to write kernels that use amplitude encoding, inversion, and loops in the same way as before. The changed behavior appears when a user executes the emitted Qiskit circuit through the default Qiskit executor backed by Aer: instead of handing Aer an unsafe empty-parameter multiplexer that can crash the interpreter, Qamomile performs a targeted execution-time lowering step.

For example, a user-level circuit shape like this remains valid from the user's point of view:

```python
executable = QiskitTranspiler().transpile(kernel)
counts = executable.sample(QiskitTranspiler().executor(), shots=1024).result()
```

If the emitted Qiskit circuit contains an inverse state-preparation path inside retained control flow, the user does not need to manually disable Qiskit native composites just to avoid the Aer crash. The manual escape hatch still exists: `QiskitTranspiler(use_native_composite=False)` avoids the Qiskit native `StatePreparation` path and emits Qamomile's fallback decomposition instead. The branch documents this behavior in `LIMITATIONS.md:127-135` because the executor now has an Aer-specific compatibility step that users may care about when comparing emitted circuit resources.

The backend change has one objective: keep Qamomile's Qiskit circuit high-level until execution, but prevent Aer from receiving the one Qiskit instruction shape that can crash native assembly. Native assembly is the lower-level Aer code path that turns Qiskit circuit instructions into Aer simulator operations.

### Detecting the unsafe instruction shape

The first piece identifies only the crash-prone shape, not every multiplexer. The previous top-level scan could not solve the nested-loop case because a `for_loop` instruction owns one or more block circuits, and the unsafe operation can live inside those blocks rather than in `circuit.data` at the root. The helper `_contains_empty_parameter_multiplexer()` walks the root circuit and recursively walks every `ControlFlowOp.blocks` entry, where `ControlFlowOp` is Qiskit's base class for structured control-flow instructions such as loops. It returns true only for `operation.name == "multiplexer" and not operation.params`, so params-bearing multiplexers still count as safe; see `qamomile/qiskit/transpiler.py:297-318`.

### Lowering only before Aer execution

The second piece applies the workaround at the last practical moment: after the normal Qiskit transpilation for the selected backend, but before the Aer backend's `run` method. This matters because Qamomile should not rewrite the public circuit structure earlier than needed; the public Qiskit circuit can remain high-level, and non-Aer backends do not inherit an Aer-specific workaround. `QiskitExecutor.execute()` now calls `_decompose_empty_parameter_multiplexers()` only when the backend module starts with `qiskit_aer.`, after measurements have been ensured and after Qiskit's regular `transpile` call; see `qamomile/qiskit/transpiler.py:92-99`.

### Preserving control flow while decomposing nested blocks

The lowering helper keeps retained control flow instead of unrolling loops. It calls Qiskit's `decompose` with both `"multiplexer"` and `ControlFlowOp` in the decomposition filter, then re-transpiles for the same backend. Including `ControlFlowOp` is what lets Qiskit map the decomposition through control-flow block circuits while keeping the outer loop as a loop. If an empty-parameter multiplexer remains afterward, `_decompose_empty_parameter_multiplexers()` raises `RuntimeError` so the user gets a Python exception instead of a possible process crash; see `qamomile/qiskit/transpiler.py:321-354`.

### Capturing the intended boundary in tests and docs

The tests cover both sides of the boundary. `test_executor_decomposes_nested_empty_multiplexer_without_unrolling()` builds a Qiskit `for_loop` whose body contains `StatePreparation(...).inverse()`, verifies that the fixed circuit still has a top-level `for_loop`, verifies that no empty-parameter multiplexer remains, and executes through `QiskitExecutor`; see `tests/transpiler/backends/test_qiskit.py:62-93`. `test_executor_leaves_parameterized_multiplexer_untouched()` builds a params-bearing `UCGate` multiplexer, verifies that the helper returns the original circuit unchanged, and executes it on Aer; see `tests/transpiler/backends/test_qiskit.py:95-123`. The limitation entry in `LIMITATIONS.md:127-135` records that this is an Aer-specific executor-side compatibility lowering, not a general statement that all multiplexers are unsupported.

One alternative was to decompose every `multiplexer` before Aer execution. That would avoid the crash, but it would also rewrite valid params-bearing multiplexers that Aer can execute correctly. The branch rejects that broader approach and instead keys the workaround on the empty `params` list, which is the part of the instruction shape that makes Aer unsafe.

Another alternative was to unroll Qiskit `for_loop` operations before searching for multiplexers. That would make nested instructions visible at the root, but it would change circuit size and structure based on loop trip count, and it would lose the benefit of preserving structured control flow for backends that support it. The branch instead searches and decomposes through control-flow blocks directly, so the loop remains a loop.

A third alternative was to turn off Qiskit native semantic composites by default. That would avoid Qiskit's native `StatePreparation` path, but it would also remove useful backend-native behavior for unrelated circuits. The branch keeps native composites available and limits the compatibility fix to the Qiskit Aer executor path.

## 5. Known Limitations

When: a user inspects the Qiskit circuit before execution. Why: the branch intentionally applies the Aer-specific lowering inside `QiskitExecutor.execute()` after normal transpilation, so the circuit returned by `QiskitTranspiler` may still contain high-level Qiskit instructions. Future fix: if users need a public "sanitize for backend execution" API, expose the same final-execution preparation step without requiring an actual run.

When: upstream Qiskit or Aer changes the representation of inverse-derived multiplexers. Why: this branch is a compatibility workaround for the current contract mismatch where Qiskit can place matrices in a gate definition while Aer reads matrices from `operation.params`. Future fix: remove this workaround once Qiskit avoids producing the empty-parameter native-name collision or Aer safely rejects or decomposes it through the gate definition.

PR #605, opened 2026-07-18
Reject vector controls in inverse blocks

ksk-jij:
## Summary

Reject vector-valued control operands when constructing `InverseBlockOperation` and at the serialization validation boundary. Document the scalar-control invariant used by inverse emission and add regression coverage for constructor and forged-IR validation paths.

## Root cause

`num_control_qubits` counts leading control operands, while inverse emission expands each quantum operand into physical qubit indices. A vector control therefore contributed multiple physical indices but only one control operand, so the activation-control slice bracketed only the final physical qubit and implemented the wrong control pattern.

## Impact

Frontend-generated inverse blocks are unchanged because the frontend already scalarizes controls. Hand-built or deserialized IR with vector control operands now fails with a clear validation error instead of emitting a circuit with incorrect activation semantics.

## Validation

- `uv run pytest`: 9107 passed, 413 skipped, 2530 deselected, 1 xfailed


ksk-jij:
## 1. Problem Overview

`InverseBlockOperation` is an IR operation, where IR means the compiler's internal representation of a quantum program, that can carry a qkernel's original forward block and a block containing the reversed operations. Each control operand, meaning an input value passed to the operation, is expected to be a scalar qubit: one IR value representing one physical qubit.

On `origin/main`, the constructor, meaning the initialization method, checked only whether a control operand had a quantum type, so it also accepted an `ArrayValue`, which is one IR value representing an array of qubits. At the same time, `num_control_qubits` was used as the number of leading control operands and as the physical width of the operation's own control set. Those meanings diverged when one array operand expanded to several physical qubits (`qamomile/circuit/ir/operation/inverse_block.py:37-48`, `qamomile/circuit/transpiler/passes/emit_support/inverse_emission.py:122-151`, `qamomile/circuit/transpiler/passes/emit_support/inverse_emission.py:212-216`).

The following malformed operation could therefore be constructed without an exception on `origin/main`. `control` is one operand containing three qubits, while `num_control_qubits=1` is later interpreted as both one operand and one physical control. During emission, which converts IR operations into a backend-specific circuit such as a Qiskit circuit, `_expand_quantum_operands_to_phys` expands that array to three backend qubit indices (`qamomile/circuit/transpiler/passes/emit_support/controlled_block_support.py:800-886`). However, the emitter, which performs that conversion, uses `num_control_qubits` to select only the final index when applying `control_value=0`, the integer pattern that says which zero-or-one control state activates the operation. The pattern therefore does not cover the declared control array (`qamomile/circuit/transpiler/passes/emit_support/inverse_emission.py:212-222`).

```python
dimension = Value(type=UIntType(), name="dimension").with_const(3)
control = ArrayValue(type=QubitType(), name="control", shape=(dimension,))
target = Value(type=QubitType(), name="target")

op = InverseBlockOperation(
    operands=[control, target],
    results=[control.next_version(), target.next_version()],
    num_control_qubits=1,
    num_target_qubits=1,
    source_block=Block(),
    implementation_block=Block(),
    control_value=0,
)
```

This branch requires every inverse-block control operand to be a scalar qubit. As a result, `num_control_qubits` now has the same value when used to divide input operands from updated output results, validate the control activation value, and select physical control indices (`qamomile/circuit/ir/operation/inverse_block.py:70-106`).

User-written qkernel syntax does not change. In the concrete `qmc.control` path, the frontend, meaning the layer that converts Python qkernel code into IR, expands a `Vector[Qubit]`, which is a frontend array of qubits, or a vector slice into separate scalar control operands (`qamomile/circuit/frontend/operation/control.py:1054-1098`). The supported concrete-vector path therefore does not place an atomic `ArrayValue`, meaning one indivisible array operand, in an `InverseBlockOperation` control segment. Direct IR construction or a compiler transformation that alters an existing operation can still create that malformed shape, and this branch rejects it.

The closest supported frontend path uses a whole two-qubit vector as the control input, then applies the controlled qkernel and its inverse. The vector-control call form is covered by `tests/circuit/test_global_phase.py:3667-3691`, while the forward-then-inverse execution pattern is covered by `tests/circuit/test_inverse.py:1617-1637`. In the combined example below, both controls start in state one, the controlled X first flips the target to one, and its inverse returns the target to zero. The final state therefore has two one-valued controls and a zero-valued target; this exact example was also transpiled and executed successfully on the branch.

```python
@qmc.composite_gate(name="boxed_x")
def boxed_x(target: qmc.Qubit) -> qmc.Qubit:
    return qmc.x(target)

@qmc.qkernel
def controlled_layer(
    controls: qmc.Vector[qmc.Qubit],
    target: qmc.Qubit,
) -> tuple[qmc.Vector[qmc.Qubit], qmc.Qubit]:
    return qmc.control(boxed_x, num_controls=2)(controls, target)

@qmc.qkernel
def circuit() -> qmc.Vector[qmc.Bit]:
    qs = qmc.qubit_array(3, "qs")
    qs[0] = qmc.x(qs[0])
    qs[1] = qmc.x(qs[1])
    controls = qs[0:2]
    target = qs[2]
    controls, target = controlled_layer(controls, target)
    controls, target = qmc.inverse(controlled_layer)(controls, target)
    qs[0:2] = controls
    qs[2] = target
    return qmc.measure(qs)
```

Code that constructs IR directly now receives `ValueError: inverse block control operands must be scalar qubits.` as soon as it supplies an `ArrayValue` in the leading control range. The constructor-level behavior is covered by `tests/circuit/test_inverse.py:3743-3756`. This early diagnostic replaces an ambiguous operation that could otherwise survive until backend emission.

The internal changes establish one invariant, meaning a condition that must always remain true: every inverse-block control operand occupies exactly one physical qubit. The constructor establishes the invariant, serialization validation—the whole-IR consistency check run before stored IR is accepted—independently checks it, and the emitter documents why its control-width calculation is safe under that invariant.

### Reject Ambiguous Width At Construction

The first goal is to reject an invalid inverse block at the point where it is created. The pre-existing `operand.type.is_quantum()` check cannot do that because both scalar qubits and qubit arrays have quantum types. `InverseBlockOperation.__post_init__`, the initialization method for this Python data class, now checks each of the first `num_control_qubits` operands and raises before the broader quantum-type check if an operand is an `ArrayValue` (`qamomile/circuit/ir/operation/inverse_block.py:70-106`).

```python
for operand in self.operands[: self.num_control_qubits]:
    if isinstance(operand, ArrayValue):
        raise ValueError(
            "inverse block control operands must be scalar qubits."
        )
    if not operand.type.is_quantum():
        raise ValueError("inverse block control operands must be quantum.")
```

The class documentation now defines the operand order as scalar controls followed by quantum targets and then classical or object parameters. It also defines `num_control_qubits` as the number of leading scalar control operands and matching pass-through results, where a pass-through result is the updated output position corresponding to a control input (`qamomile/circuit/ir/operation/inverse_block.py:29-58`). This constructor contract is what the independent boundary check in the next subsection revalidates.

### Defend The Serialization Boundary

The second goal is to reject malformed IR even if it became invalid after construction. A constructor check alone is insufficient because the operation's `operands` list is mutable and can be replaced by a custom compiler transformation or external producer. Serialization validation, meaning the whole-IR consistency check performed around conversion to and from Qamomile's stored representation, therefore needs its own defense.

`_validate_inverse_block` now rejects an `ArrayValue` anywhere in the leading control range before it validates the control activation value (`qamomile/circuit/serialization/validation.py:1908-1933`). The test first constructs a valid scalar-controlled operation, then replaces its control operand and result with arrays; `validate_qkernel_ir` still rejects that altered operation, proving that the boundary check does not depend on the constructor having seen the bad value (`tests/circuit/serialization/test_qkernel.py:867-901`).

```python
if any(
    isinstance(value, ArrayValue)
    for value in operation.operands[: operation.num_control_qubits]
):
    raise ValueError(f"{location} control operands must be scalar qubits")
```

The invariant revalidated here is the condition used by the emitter logic in the next subsection.

### Keep Activation-Value Width Unambiguous

The final goal is to apply `control_value`, an integer encoding which zero-or-one state activates each control, only to the inverse operation's own controls. Nested emission prepends controls from enclosing operations to `control_indices`, the list of backend qubit positions, so the emitter must select a trailing segment rather than use the entire list. Before this branch, an array operand could make the number of operands smaller than the number of physical indices, so `num_control_qubits` could not unambiguously identify that segment.

```python
activation_controls = (
    control_indices[-op.num_control_qubits :]
    if op.num_control_qubits
    else []
)
```

The scalar-only invariant makes one control operand equal one physical control index, so the existing Python list slice now has a documented and enforceable meaning. `activation_controls` is then passed to `bracket_control_value`, which temporarily surrounds zero-valued controls with X gates so the backend can emit an operation that activates when every control is one (`qamomile/circuit/transpiler/passes/emit_support/inverse_emission.py:212-230`). The emitter code itself does not otherwise change; its updated comment records the constructor guarantee on which the calculation depends.

One alternative was to keep each control array as one indivisible operand and calculate its physical width from `ArrayValue.shape` during emission. That would fix only the trailing-index slice, while `num_control_qubits` would still mean an operand count when dividing operands and results and a bit width when validating `control_value`. Supporting that representation would therefore require a broader IR design that separates control operand count from scalar control width. This branch instead aligns inverse blocks with the existing concrete frontend representation, which already expands vector controls into scalar operands.

Another alternative was to reject array controls only when the emitter encountered them. That would preserve invalid IR through earlier compiler stages and delay the same structural error until a backend conversion was attempted. Constructor-only rejection was also insufficient because operation lists can be mutated after construction. The chosen two-layer validation reports ordinary construction errors immediately while still protecting serialization from altered or externally produced IR.

## 5. Known Limitations

When: A custom IR producer represents several inverse-block controls as one indivisible `ArrayValue` operand. Why: This branch deliberately rejects that representation because `num_control_qubits` still combines operand count and physical-width responsibilities. Future fix: If one-operand array controls become a requirement, introduce separate fields for the number of control operands and the scalar control width, then update activation-value validation, result pairing, serialization, and shared emission logic together.

PR #606, opened 2026-07-18
Fix SELECT serialization operand pairing

ksk-jij:
## Summary

- Pair SELECT case-block inputs with category-grouped operands before serialization type validation.
- Preserve explicit arity and operand-category rejection for malformed payloads.
- Add a dict encode/decode regression covering a classical parameter declared before the quantum target.

## Root cause

`SelectOperation.operands` stores quantum targets before classical parameters, while case-block inputs retain source declaration order. `_validate_select` compared those sequences positionally, so a valid case such as `(angle: Float, target: Qubit)` was rejected during both serialization and deserialization even though tracing and backend transpilation succeeded.

## Validation

- `uv run pytest tests/circuit/serialization/test_qkernel.py` — 89 passed
- `uv run pytest tests/circuit/test_select.py` — 66 passed
- `uv run pytest tests/transpiler/test_block_parameter_binding.py` — 11 passed
- Local review found no remaining issues.


ksk-jij:
## 1. Problem Overview

Qamomile's SELECT control-flow operation chooses one case block from several candidates according to an index represented by one or more qubits. A case block's formal inputs—the receiving arguments written in the function definition—retain Python declaration order. In Qamomile's intermediate representation (IR), its internal circuit model, the corresponding SELECT operands—the actual arguments at the call site—are stored by category: quantum targets first, followed by classical or object parameters.

On `origin/main`, serialization validation compared these two sequences positionally. A valid case declaration such as `(angle: Float, target: Qubit)` was therefore checked against operands stored as `(target: Qubit, angle: Float)`, producing an incorrect type error that expected `Qubit` for the leading `Float`. The same validation runs both when encoding IR to a dictionary and when decoding IR from a dictionary, so a valid program that could be traced, transpiled, and executed still failed specifically when saved or loaded. This branch aligns declaration order with category order before checking types.

The user-facing SELECT syntax and behavior do not change. Case functions may continue to interleave classical parameters and quantum targets in any declaration order. Programs that declare a parameter before a target, as shown below, can now be serialized and deserialized correctly.

```python
@qmc.qkernel
def rx_case(angle: qmc.Float, target: qmc.Qubit) -> qmc.Qubit:
    return qmc.rx(target, angle)


@qmc.qkernel
def rz_case(angle: qmc.Float, target: qmc.Qubit) -> qmc.Qubit:
    return qmc.rz(target, angle)


index, target = qmc.select([rx_case, rz_case])(index, 0.5, target)
```

The added regression program uses RX and RZ cases with this argument order. It converts the kernel to a dictionary, restores a kernel from that dictionary, and converts it again to confirm that the result is unchanged, preventing the user-visible failure on both encoding and decoding (`tests/circuit/serialization/test_qkernel.py:332-378`, `tests/circuit/serialization/test_qkernel.py:867-874`).

The internal objective is to align serialization validation with the argument-pairing convention already used by compiler stages that transform Python declarations into backend circuits, without changing the SELECT storage format or the `SelectOperation` operand contract. The implementation changes are limited to SELECT serialization validation and its regression coverage.

### Declaration-Order Operand Pairing

This part pairs each formal input with the correct actual operand category before comparing types. The previous positional check could not represent the relationship because case inputs use declaration order while SELECT operands use category order, even when both the counts and types are valid.

`_validate_select` removes the index operands and calls `pair_block_operands` instead of comparing the remaining operands directly with the case inputs (`qamomile/circuit/serialization/validation.py:2003-2018`). This shared helper, a utility reused by multiple compiler stages, splits actual operands into quantum and non-quantum pools and consumes the appropriate pool according to whether each formal input has a quantum type (`qamomile/circuit/transpiler/block_parameter_binding.py:10-76`). As a result, `(angle: Float, target: Qubit)` is paired with the classical angle operand and the quantum target operand in the correct order.

The same pairing convention is used by partial evaluation, which computes values known at compile time, and by emission, which converts IR into backend circuit instructions. Reusing the helper prevents serialization validation from drifting to a different argument convention.

### Preserving Rejection of Malformed IR

This part permits only valid declaration-order differences while continuing to reject malformed IR with missing inputs or mismatched operand categories. Reordering alone could otherwise hide an input that could not be paired.

Before pairing, `_validate_select` verifies that the case input count matches the number of SELECT case operands. After pairing, it verifies that a pair was produced for every operand, raising an explicit `ValueError` when the quantum and non-quantum category counts disagree. The existing `_require_types` check then rejects mismatched concrete types (`qamomile/circuit/serialization/validation.py:2003-2024`).

### Regression Coverage Across Encoding and Decoding

This part ensures that the fix applies regardless of whether validation is entered through saving or loading. Testing only one direction could leave the same false rejection in the other entry point.

`test_select_parameter_before_target_round_trips_through_dict` encodes with `kernel_to_dict`, decodes the payload with `kernel_from_dict`, and encodes the restored kernel again. Comparing the final dictionary with the original confirms both that validation succeeds and that the fix does not reorder or rewrite the persisted representation (`tests/circuit/serialization/test_qkernel.py:867-874`).

Changing only the expected slice order while retaining positional comparison would not handle the two valid conventions simultaneously: category order for SELECT operands and declaration order for case inputs. Changing `SelectOperation` itself to store operands in declaration order would also alter a contract used by partial evaluation, conversion into concrete backend circuits, and emission, making the scope too broad for a validation-only defect.

A serialization-specific reordering implementation would duplicate existing compiler logic and could drift from the compiler's pairing convention over time. Reusing `pair_block_operands` was chosen because it directly expresses the established formal-to-actual contract and already supports arbitrary interleaving of quantum and non-quantum inputs.

## 5. Known Limitations

None.

PR #607, opened 2026-07-18
Rollback output from declined callable emitters

ksk-jij:
## Summary

- Roll back circuit-builder state when a callable implementation emitter declines an operation.
- Remove the control-value bracket and any partial gates before the generic controlled-composite fallback runs.
- Cover a zero-valued control with an emitter that appends a gate and then returns `False`.

## Root cause

The callable implementation path opened the zero-control X bracket and wrote emitter output before learning whether the emitter accepted the operation. Returning `False` selected the fallback without reverting those writes, so the fallback was appended after dead bracket gates and partial native output.

## Validation

- `uv run pytest` (9,103 passed, 414 skipped, 2,425 deselected, 1 xfailed)
- `uv run zuban check qamomile/` reports the same 56 optional HUGR dependency errors as `origin/main`; no errors are introduced by this branch.


ksk-jij:
## 1. Problem Overview

On `origin/main`, a callable implementation emitter—a backend-specific hook that emits a composite operation—could append gates and then return a falsey value, meaning a value for which `bool(...)` is `False`, to decline the operation. For a controlled invocation with `control_value=0`, Qamomile had already opened a control-value bracket, which is a pair of X gates that temporarily converts a zero-valued control into an ordinary one-valued control, before the hook returned its decision. The generic fallback, meaning Qamomile's complete built-in implementation used after a decline, then ran after the abandoned bracket gates and the hook's partial output, producing a circuit with redundant bracket gates and extra operations that could change the circuit's behavior.

This branch makes decline handling transactional. A transaction is an emission attempt whose writes either all remain on success or are all discarded on decline. The circuit builder, an internal object that constructs a circuit incrementally, records its state before opening the bracket. If the hook's result converts to `False`, the builder restores that state before the existing fallback emits the complete operation. If the result converts to `True`, the bracket and the hook's output remain unchanged.

There is no new frontend API. Existing qkernels, which are functions that define Qamomile quantum programs, continue to use `qmc.control(..., control_value=0)` when applying a composite operation under a zero-valued control.

```python
@qmc.qkernel
def kernel() -> qmc.Bit:
    control = qmc.qubit("control")
    target = qmc.qubit("target")
    control, target = qmc.control(
        custom_composite,
        control_value=0,
    )(control, target)
    return qmc.measure(target)
```

The observable change is in the emitted circuit when a registered controlled implementation declines. Users now receive only the complete generic controlled operation surrounded by the single X-gate pair required for `control_value=0`; abandoned X gates and partial gates written by the hook no longer precede it. The regression scenario is represented by the declining emitter and qkernel in `tests/transpiler/test_circuit_ir.py:50-99` and verified at `tests/transpiler/test_circuit_ir.py:1206-1224`.

The backend changes collectively give callable emission a safe commit-or-decline boundary. They first capture the mutable builder state that an append-only emission attempt can advance, then use that record to erase a declined attempt before the normal fallback starts. Here, append-only means adding new state without rewriting state that existed before the attempt, and IR means the intermediate representation used between Qamomile's frontend and backend emission.

### Capturing a complete builder checkpoint

The goal is to save enough state that a declined emitter can be undone without corrupting later circuit construction. Saving only the instruction count is insufficient because appending a gate also advances wire versions, which are identifiers that connect each operation's output to the next operation's input. An emitter can also change the global phase, which is a phase contribution attached to a circuit region, or enter nested control-flow construction.

`_CircuitBuilderSnapshot` at `qamomile/circuit/transpiler/circuit_ir/model.py:1098-1118` records the current region object, its instruction count, its qubit-to-wire mapping, its global phase, the next wire identifier, and the depths of the region and control stacks. A region is the builder state for one circuit or nested control-flow body; the region stack selects the active nested body, while the control stack records the active for-loop, if-branch, or while-loop construction context. `CircuitBuilder.snapshot_state()` at `qamomile/circuit/transpiler/circuit_ir/model.py:1161-1177` creates a checkpoint, meaning a saved point to which the builder can later return, for an append-only emission attempt. This saved record is what the restoration step in the next subsection consumes.

### Reverting a declined emission attempt

The goal is to return the builder to the exact checkpoint before the generic fallback emits. Closing the control-value context alone cannot achieve this because its `finally` block appends the closing X gate but does not remove the opening X gate or the emitter's partial output.

`CircuitBuilder.restore_state()` at `qamomile/circuit/transpiler/circuit_ir/model.py:1179-1213` removes regions, control contexts, and operations appended after the checkpoint, then restores the saved wire mapping, global phase, and next wire identifier. It raises `RuntimeError` if the region or control stack is shallower than it was at the checkpoint, if the current region object was replaced, or if the number of operations that existed before the checkpoint was reduced. Guessing how to restore such non-append-only changes could hide deeper corruption. The callable helper at `qamomile/circuit/transpiler/passes/emit_support/composite_gate_emission.py:292-309` takes the checkpoint before entering the control-value bracket, keeps the attempt when the emitter's result converts to `True`, and restores the checkpoint when the result converts to `False`. The existing caller can therefore select its controlled-composite fallback against clean state.

### Proving that both bracket and partial output disappear

The regression coverage aims to distinguish full rollback from a narrower fix that removes only the extra X gates. An emitter that declines without writing anything would expose the bracket defect, but it would not expose the older hazard in which an emitter writes part of an implementation before declining.

The test emitter at `tests/transpiler/test_circuit_ir.py:50-73` deliberately appends a Hadamard gate and returns `False`, and the controlled qkernel at `tests/transpiler/test_circuit_ir.py:76-99` invokes it with `control_value=0`. The assertion at `tests/transpiler/test_circuit_ir.py:1206-1224` requires the final program to contain exactly the fallback's opening X gate, controlled call, closing X gate, and measurement, and separately verifies that no partial Hadamard gate remains. Circuit verification then confirms that the restored wire connections are internally consistent.

A capability pre-check could add a method that decides whether an emitter will handle an operation before it writes to the circuit. This would avoid speculative writes for new emitters, but it would require changing the emitter contract and every implementation, and it would not protect existing emitters whose final decision depends on details discovered during emission.

Emission could instead target a scratch builder, which is a temporary builder isolated from the real circuit, and splice successful instructions into the real builder. Splicing means transferring the temporary result into the destination after success. This is harder to make correct because temporary wire identifiers, global phase, and nested region state would need to be reconciled with the destination builder.

Truncating only the instruction list would be a smaller change, but it would leave advanced wire identifiers, updated wire mappings, phase changes, or nested construction state behind. The checkpoint-and-restore approach was chosen because it preserves the existing rule that a boolean return value reports whether the emitter accepted the operation, while rolling back every recorded part of the builder state that an append-only emitter can advance.

## 5. Known Limitations

When: A third-party emission target passed to `emit_callable_implementation_emitter` does not provide both `snapshot_state()` and `restore_state()` hooks. Why: The helper checks for these hooks at runtime to preserve compatibility with circuit objects other than `CircuitBuilder`, so an object without them retains the previous no-rollback behavior if its emitter writes output and then declines. Future fix: Define a common transactional circuit protocol, meaning a shared interface for checkpoint and restoration, and require callable emitters that can decline after writing to use a target implementing that protocol.

When: A custom emitter removes or replaces circuit state that existed before its checkpoint instead of only appending new state. Why: The rollback contract intentionally supports append-only emission attempts, and `restore_state()` raises `RuntimeError` for detectable violations rather than guessing how to reconstruct arbitrary mutations. Future fix: If non-append-only emitters become a supported use case, isolate every attempt in a scratch builder with an explicit commit operation instead of expanding the current checkpoint contract.

PR #622, opened 2026-07-21
docs: update PCE tutorial headings and index summary

jij-yuichiro-nakano:
## Summary

- Update the English and Japanese algorithm index cards to describe solving a 20-node MaxCut problem with 3-qubit Pauli Correlation Encoding and comparing the result with the exact solution.
- Simplify the implementation subsection headings by removing the numbered `Step` prefixes.
- Sync the paired English and Japanese notebooks without re-executing them, preserving the existing outputs.

## Verification

- `uv run jupytext --to ipynb --update docs/en/algorithm/pce_maxcut.py docs/ja/algorithm/pce_maxcut.py`
- `git diff --check`
- `./docs/build.sh build-en`
- `./docs/build.sh build-ja`


docs/en/algorithm/index.md
```diff
@@ -35,7 +35,7 @@ Implement multidimensional QFT and classical preprocessing for inputs whose grid
 :::{card}
 :header: **Pauli Correlation Encoding (PCE)**
```
ksk-jij: [my opinion]
Quite small stuff, but it might be nice to omit "(PCE)" since the website is not that large. This card itself is to be fair alright, but the length of the side bar is not that wide and "(PCE)" makes it 2 rows. It might be nice to reduce it. If we omit (PCE) from the title, the header here also can be reduced like so.

Image

docs/ja/algorithm/pce_maxcut.py
ksk-jij: [my opinion]
Might be nice to omit (PCE) from the title for the side bar.

ksk-jij (commented):
@jij-yuichiro-nakano Hi, thank you for the PR. I've left tiny comments, please have a look!

docs/en/algorithm/index.md
jij-yuichiro-nakano: Thank you for the suggestion. Removed `(PCE)` from the English and Japanese index card headers and tutorial titles in `cfe5e224`, keeping the sidebar labels shorter while retaining the abbreviation in the tutorial text.

docs/ja/algorithm/pce_maxcut.py
jij-yuichiro-nakano: Thank you for the suggestion. Updated the Japanese title, its paired notebook, and the corresponding English title in `cfe5e224` so the sidebar uses the shorter `Pauli Correlation Encoding` label.

jij-yuichiro-nakano:
@ksk-jij Thanks to your comments! I agree too, and approved your suggestion. Please have a look!

ksk-jij (approved):
@jij-yuichiro-nakano Thank you for updating them! Looks good to me :)

PR #631, opened 2026-07-23
Add QSVT eigenstate-filtering converter 

baptistechev:
## Summary

Adds **QSVT eigenstate filtering** to Qamomile: a new optimization converter and
the supporting circuit kernels that implement the Lin & Tong (2020,
arXiv:2002.12508) ground-energy search on top of the existing `qmc.qsvt`
primitive and the `LCUBlockEncoding` producer family.

The converter compiles a QUBO/HUBO into a sampling circuit: the Ising cost
Hamiltonian is block encoded, shifted by `-mu I`, filtered with a QSVT
approximation of the sign function, and probed with one Hadamard-test qubit. The
fraction of shots with all ancillas zero estimates `|| P_<mu |phi_0> ||^2` — the
predicate a classical binary search over `mu` consumes to locate the ground
energy. The search itself stays outside the library; the converter only emits
quantum programs, and the tutorial shows the search driving them.

## What this branch adds to `main`

**New optimization converter** — `qamomile/optimization/qsvt_filter.py`

- `QSVTFilterConverter(MathematicalProblemConverter)`. Block encodes the cost
  Hamiltonian's non-constant part once (`ising_z_block_encoding`); each threshold
  composes a second LCU for `H - mu I` (`lcu_block_encoding`).
- **The Ising constant is held out of the encoding** and exposed as
  `energy_offset` (`c`) alongside `normalization` (`alpha`). A constant shifts
  every eigenvalue equally, so it changes neither the eigenvectors nor their
  order, but as an identity term it would consume subnormalization budget — the
  quantity that sets how finely the filter resolves `mu`. The residual shift
  therefore costs `alpha + |mu - c|` rather than `alpha + |c| + |mu|`, which the
  triangle inequality never makes worse and which is much smaller when `mu` sits
  near `c` — the usual case, since a spectrum is centred on its own constant.
  `mu` keeps its meaning in the problem's own energy units, and the pair
  `[c - alpha, c + alpha]` bounds the spectrum, which is what a threshold gets
  chosen against.
- Public API: `transpile(transpiler, *, mu, degree, delta, scale, phi=None)`,
  `success_probability(result)`, `num_ancilla_bits(mu)`,
  `decode_to_binary_sampleset(result)` / `decode(result)` (post-select
  all-ancilla-zero shots, then reuse the base decoder — OMMX round trip
  preserved). `get_cost_hamiltonian` raises `NotImplementedError`: the cost
  operator is block encoded internally, not exposed as an observable.
- Phase synthesis (`pyqsp`) converts the sign polynomial into the
  projector-rotation convention `qmc.qsvt` consumes; cached per
  `(degree, delta, scale)`, so a binary search pays for it once.
- `DEFAULT_POLYNOMIAL_SCALE = -1.10` keeps the eigenspace **below** `mu`. That is
  the orientation every OMMX problem wants, since `Instance.to_hubo` negates a
  maximization into a minimization before the converter sees it, so the optimum
  is always the smallest eigenvalue. A positive `scale` inverts the filter into a
  largest-eigenvalue search; nothing else about the circuit changes.
- **QSP bound is checked numerically.** Whether `scale` times the sign
  approximation stays under `|p| <= 1` is a joint property of `degree`, `delta`
  and `scale`, so no bound on `scale` alone can decide it. `_qsp_phases`
  measures `max|scale * p|` on `POLYNOMIAL_BOUND_SAMPLES` points and raises
  before the expensive phase optimization. Without it, a combination such as
  `(degree=21, delta=20)` reaches `|p| = 1.093` and yields phases that encode
  nothing — observed as a filter keeping ~80% of the spectrum where the true
  answer was ~3%, i.e. silent garbage rather than a soft failure.
- `pyqsp` prints unconditionally (its `verbose` flags are not consulted), so its
  output is captured and logged at `DEBUG`. Synthesis instead prints one summary
  line naming the filter being built; clear `SHOW_PHASE_SYNTHESIS_SUMMARY` to
  silence it.

**New circuit kernels** — `qamomile/circuit/algorithm/qsvt_filter.py`
(exported from `qamomile.circuit.algorithm`)

- `eigenstate_filter_projector(encoding)` — Hadamard-test projector `(I - R)/2`
  built by controlling a `qmc.qsvt` reflection. The kernels are
  direction-neutral: which of the two eigenspaces survives is carried entirely by
  the phases, so the same kernels serve both the ground-energy and the
  largest-eigenvalue search.
- `eigenstate_filter_probe(encoding)` — prepares the uniform superposition,
  applies the projector, measures projector/signal/system registers.

**Dependencies** — new optional extra `qsvt` (`pyqsp>=0.2.0`) in `pyproject.toml`
(plus `dev` and `uv.lock`), needed only for automatic phase synthesis; callers
can pass precomputed `phi` and skip `pyqsp`.

**Docs** — `docs/{en,ja}/algorithm/qsvt_filter_graph_partitioning.{py,ipynb}`

An end-to-end article on a 6-node triangular-prism graph-partitioning instance,
whose optimal equal partition is unique. It builds the converter, probes one
threshold, draws the three nested circuits (block encoding `U`, the `qmc.qsvt`
alternation, the Lin & Tong projector), then hands the predicate to a classical
binary search that brackets the ground energy in three probes without being told
where it is. Both languages carry assertions on the bracket, the post-selection
purity, and the recovered partition. Section index cards added for both.

## Test plan

- `tests/circuit/algorithm/test_qsvt_filter.py` — descriptor-type rejection, the
  projector realizes `(I - R)/2` (parametrized over seeds), and cross-backend
  execution of the probe (Qiskit / QuriParts / CUDA-Q) through both the sampler
  and estimator paths, parametrized over seeds and system widths.
- `tests/optimization/test_qsvt_filter.py` — the constant is held out of the
  encoding and tracked as `energy_offset`; the shifted normalization grows as
  `alpha + |mu - c|` and the identity term collapses when `mu == energy_offset`;
  holding the constant out never widens the subnormalization; the predicate is
  unchanged by that optimization (same `mu` selects the same fraction of the
  spectrum); phase length/caching; the QSP bound check accepts a bounded
  polynomial and rejects `(degree=21, delta=20, scale=1.10)` with the measured
  peak in the message; pyqsp output is demoted to `DEBUG` and the summary line is
  suppressible; odd-length phase rejection; circuit sizing;
  `get_cost_hamiltonian` raising; the `mu`-threshold spectrum staircase
  (`-3.5 -> 0.0`, `-0.5 -> 0.75`, `3.5 -> 1.0`); that a positive `scale` inverts
  the filter into the complementary half; and ground-state recovery from
  post-selected samples.
- Docs tests cover all four page sources (EN/JA x `.py`/`.ipynb`), discovered
  automatically by `TUTORIAL_PATTERNS`.
- `ruff check` / `ruff format` clean; `zuban` clean on the changed files.

## Notes for reviewers

- The tutorial calls `converter._shifted_encoding(mu)`, a private method, so the
  circuit drawings match the circuit that actually runs at that threshold. The
  call is commented as internal; exposing a public accessor would be the
  alternative if reviewers prefer.
- `_qsp_phases` contains the package's only `print`. Synthesis blocks for seconds
  inside `transpile`, so the one-line summary explains the pause; it is
  suppressible via `SHOW_PHASE_SYNTHESIS_SUMMARY`.
- `zuban check qamomile/` reports 56 errors in `qamomile/hugr/lowerer.py`. They
  are pre-existing on `main` (verified against a clean worktree) and untouched by
  this branch.


qamomile/optimization/qsvt_filter.py
```diff
@@ -0,0 +1,534 @@
...
+                "`pip install 'qamomile[qsvt]'`, or pass precomputed phases via "
+                "the `phi` argument of transpile()."
+            ) from error
+
+        if SHOW_PHASE_SYNTHESIS_SUMMARY:
+            print(
+                f"[qamomile.qsvt_filter] synthesizing sign filter: "
+                f"degree={degree}, delta={delta}, scale={scale} "
+                f"-> {degree + 1} phases"
+            )
```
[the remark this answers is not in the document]
baptistechev: Fixed in 4ef8d1dc. Agreed — the summary now goes to the module logger at INFO:

```python
_LOGGER.info(
    "synthesizing sign filter: degree=%s, delta=%s, scale=%s -> %s phases",
    degree, delta, scale, degree + 1,
)
```

`qamomile/optimization/qsvt_filter.py` no longer contains any `print`, so it is consistent with the pyqsp capture beside it and with the rest of the package.

I also took the second half of your suggestion and dropped `SHOW_PHASE_SYNTHESIS_SUMMARY` rather than keeping it as a toggle. Once the line is a log record the level *is* the toggle, and the flag only existed because `print` cannot be level-filtered — keeping it would have been redundant public surface. Silencing is now the standard `logging.getLogger("qamomile.optimization.qsvt_filter").setLevel(logging.WARNING)`.

The summary existed because synthesis blocks for several seconds inside `transpile` and the tutorials wanted to explain the pause, so the two notebooks now print the filter shape themselves from the `degree`/`delta` they already hold. The rendered pages read as before, with nothing emitted from library code.

tests/circuit/algorithm/test_qsvt_filter.py
```diff
@@ -0,0 +1,369 @@
...
+    Returns:
+        float: Fraction of shots whose first ``num_bits`` bits are zero.
+    """
+    total = sum(count for _, count in results)
+    zero = sum(
+        count
+        for outcome, count in results
+        if all(bit == 0 for bit in _flatten(outcome)[:num_bits])
+    )
+    return zero / total
```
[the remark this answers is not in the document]
baptistechev: Fixed in 4ef8d1dc.

```python
total = sum(count for _, count in results)
assert total > 0, "sampling returned no shots"
```

I went with the assertion rather than `return 0.0`. In a test helper a silent 0.0 would surface at the call site as a confusing probability mismatch, whereas the assertion names the actual condition — which is your point about not obscuring the real failure. The docstring gained a `Raises:` section recording why, so the next reader does not re-simplify it back into an unguarded division.

baptistechev:
Addressed the two suppressed comments from the latest review in 78641c08.

**`qamomile/optimization/qsvt_filter.py:257` — phase-wrap interval.** Correct, and the docstring had it exactly backwards. `(reflection + pi) % (2*pi) - pi` inherits the divisor's sign from Python's modulo, so the result lands in `[-pi, pi)`:

```
wrap(+3.1416) = -3.1416     # +pi maps to -pi, upper end exclusive
wrap(-3.1416) = -3.1416     # -pi is included
```

The docstring now states `[-pi, pi)`.

**`docs/en/.../qsvt_filter_graph_partitioning.py:342` — "Li & Tong".** Fixed to "Lin & Tong". Only the English source was affected; the Japanese translation already read Lin & Tong, so it needed no change.

Both are prose-only, so the paired notebook was synced with `jupytext --update` and its stored outputs are untouched. Verification after the change: ruff check and format clean (511 files), zuban clean on the changed file, 43 unit tests and 4 docs tests passing (EN/JA x .py/.ipynb).

qamomile/circuit/algorithm/__init__.py
```diff
@@ -27,6 +27,10 @@
     qaoa_state,
     x_mixer,
 )
+from .qsvt_filter import (
```
jij-yuichiro-nakano: Could we revisit the public naming before exporting this API? `qsvt_filter` is broader than eigenstate or eigenvalue filtering because QSVT can implement other kinds of filters. A name such as `qsvt_eigenstate_filter` or `eigenstate_filter` would clarify the scope. Once the module name is specific, the builder names inside it may also be shortened. Please align the module, converter, and exported function names.

qamomile/optimization/qsvt_filter.py
```diff
@@ -0,0 +1,531 @@
...
+
+        Raises:
+            ImportError: If ``pyqsp`` is not installed.
+            ValueError: If ``degree`` is not a positive odd integer, ``delta``
+                is not positive, ``scale`` is zero or exceeds
+                :data:`MAX_POLYNOMIAL_SCALE` in magnitude, or the scaled
+                polynomial breaks the QSP bound :math:`\lvert p \rvert \le 1`
+                (measured on :data:`POLYNOMIAL_BOUND_SAMPLES` points).
+        """
+        if not isinstance(degree, int) or degree < 1 or degree % 2 == 0:
```
jij-yuichiro-nakano: These guards currently accept at least `degree=True`, `delta=inf`, and `scale=True`. Please reject booleans and non-finite values explicitly, and add tests for boolean, `nan`, and `inf` values. Please also define and test the contract for empty or non-finite `phi`.

qamomile/optimization/qsvt_filter.py
```diff
@@ -0,0 +1,531 @@
...
+
+        key = (degree, float(delta), float(scale))
+        cached = self._phase_cache.get(key)
+        if cached is not None:
+            return list(cached)
+
+        try:
+            from pyqsp.angle_sequence import QuantumSignalProcessingPhases
+            from pyqsp.poly import PolySign
+        except ImportError as error:  # pragma: no cover - depends on install
```
jij-yuichiro-nakano: This optional-dependency path is not exercised because the phase tests skip without `pyqsp` and CI installs it. Please test both contracts with `pyqsp` unavailable: omitting `phi` should raise this guided `ImportError`, while supplying precomputed `phi` should still transpile. This path should not need `pragma: no cover`.

tests/optimization/test_qsvt_eigenstate_filter.py
```diff
@@ -0,0 +1,483 @@
...
+    instance = ommx.v1.Instance.from_components(
+        decision_variables=[x0, x1],
+        objective=-10.0 * x0,
+        constraints=[(x0 + x1 == 1).set_id(0)],
+        sense=ommx.v1.Instance.MINIMIZE,
+    )
+    converter = QSVTFilterConverter(instance, uniform_penalty_weight=2.0)
+
+    # Measured bit 1 decodes to spin -1, i.e. binary 1.
+    raw: SampleResult[Any] = SampleResult(
```
jij-yuichiro-nakano, line 478: `Any` hides a public type mismatch here. The actual type is `SampleResult[tuple[list[int], list[int], list[int]]]`, while the inherited `decode()` accepts `SampleResult[list[int]]`. Please use the actual tuple type, add this call to `tests/typecheck/public_api.py`, and adjust or override `decode()` so the probe result is accepted statically.

tests/circuit/algorithm/test_qsvt_filter.py
```diff
@@ -0,0 +1,375 @@
...
+    reflection — on every backend, through both the sampler and the estimator.
+    """
+    rng = np.random.default_rng(seed)
+    # At most two Ising terms keep the encoding to one signal qubit.
+    words = ((0,),) if num_system_qubits == 1 else ((0,), (1,))
+    coefficients = {word: float(rng.uniform(-1.0, 1.0)) for word in words}
+    wx_phases = rng.uniform(-math.pi, math.pi, size=4)
+    phases = _to_reflection_phases(wx_phases)
+    encoding = qmc.ising_z_block_encoding(coefficients, num_system_qubits)
+    assert encoding.num_signal_qubits == 1
```
jij-yuichiro-nakano: This restricts cross-backend execution to one signal qubit. Please add a case with at least three Ising terms and `num_signal_qubits == 2`, using deterministic phases containing `0`, `π`, and `2π`, for both sampling and estimation on every backend.

To balance the CI cost, this test's seeds could be reduced to `[0, 42]`; the Qiskit projector test already exercises all four seeds.

tests/optimization/test_qsvt_eigenstate_filter.py
```diff
@@ -0,0 +1,483 @@
...
+    assert isinstance(decoded, ommx.v1.SampleSet)
+    assert decoded.get(0).objective == pytest.approx(-10.0)
+    assert not decoded.get(0).feasible
+
+
+def test_success_probability_of_an_empty_result_is_zero() -> None:
+    """A zero-shot result reports no filtered weight instead of dividing by 0."""
+    model = BinaryModel.from_higher_ising({(0,): 1.0})
+    converter = QSVTFilterConverter(model)
+    empty: SampleResult[Any] = SampleResult(results=[], shots=0)
```
jij-yuichiro-nakano, line 504: This only covers a result with zero shots. Please also test a non-empty result where every shot is rejected by projector or signal post-selection. Verify zero success probability, an empty `BinarySampleSet`, and the expected OMMX `decode()` behavior.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+# tags: [algorithm, optimization, oracle-based]
+# ---
+#
+# # QSVT Filtering to solve Graph Partitioning
+#
+# This tutorial explains how to use Quantum Singular Value Transform (QSVT) 
+# and chemistry inspired methods to find the ground energy of a given Hamiltonian
+# and solve the Graph Partitioning problem.
+#
+# > L. Lin and Y. Tong, *Near-optimal ground state preparation*, Quantum 4, 361 (2020) [arXiv:2002.12508](https://arxiv.org/abs/2002.12508)
```
jij-yuichiro-nakano: Please replace this manual reference with the project-standard MyST citation syntax:

```markdown
{cite:p}`10.48550/arXiv.2002.12508`
```

Please reuse the citation where the binary-search procedure is attributed to Lin and Tong, and mirror it in the JA source.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+# shots surviving post-selection tells a classical binary search which way to move $\mu$.
+#
+# This notebook runs that whole loop on a graph-partitioning instance, and 
+# demonstrates how to use the Qamomile library through the `QSVTFilterConverter`: 
+# build the block encoding, probe one threshold, then let
+# the binary search locate the ground energy on its own.
+
+# %%
+# Install the latest Qamomile through pip!
+# # # !pip install qamomile
```
jij-yuichiro-nakano: This cell is doubly commented and does not install the dependencies used by the page. Please make it a valid Jupytext install cell, for example:

```python
# # !pip install "qamomile[qiskit,qsvt,visualization]" networkx
```

Please mirror the change in the JA source.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+from qamomile.qiskit import QiskitTranspiler
+
+# %% [markdown]
+# ## Problem Settings
+#
+# Graph partitioning splits the vertices into two equal halves while cutting as
+# few edges as possible. We minimise the number of cut edges subject to an
+# equal-size constraint, which `to_hubo` absorbs into the objective as a penalty.
+#
+# (see the QAOA tutorial for more details about Graph Partitioning itself)
```
jij-yuichiro-nakano: Please make this an actual link to the existing tutorial, for example `[QAOA for Graph Partitioning](qaoa_graph_partition)`, and mirror it in the JA source.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+plt.show()
+
+# %%
+instance_data = {"V": num_nodes, "E": edge_list}
+instance = problem.eval(instance_data)
+
+# %% [markdown]
+# ## Algorithm
+#
+# The converter block encodes the Ising cost Hamiltonian $H$, composes a second
```
jij-yuichiro-nakano: Please keep this section focused on the algorithm rather than the Qamomile converter, and expand the theoretical explanation. In particular, state explicitly that `p` approximates the sign function of the normalized shifted operator `(H - μI) / α'`. Converter-specific details can move to the implementation section.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+#
+# $$P(\mu)=\lVert P_\mu |\varphi_0\rangle \rVert^2 = \frac{\#\{i : E_i < \mu\}}{2^n},$$
+#
+# the fraction of the spectrum below $\mu$. That single number is the predicate a
+# classical binary search needs: it is bounded away from zero exactly when an
+# eigenstate below $\mu$ carries weight in $|\varphi_0\rangle$.
+#
+
+# %% [markdown]
+# ## Using the `QSVTFilterConverter()`
```
jij-yuichiro-nakano: Please merge this section and “Building blocks” into one “Implementation with Qamomile” section, following the algorithm-page skeleton. Please apply the same structure to the JA page.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+#
+# `transpile` bakes $\mu$ into the circuit, so each threshold needs its own
+# compilation. The filter keeps the eigenspace *below* $\mu$ by default, which is
+# what a search for the minimum eigenvalue (ground energy) wants.
+#
+# `degree` and `delta` are paired: `delta` sets the width of the polynomial's
+# transition and `degree`, is the degree of the Chebyshev expansion, 
+# and must be large enough to represent it. 
+#
+# ::: if not chosen, `degree` and `delta` are set to a default value by the converter.
```
jij-yuichiro-nakano: This is not valid MyST syntax. Please convert this and the later note to `:::{note}` blocks, and mirror the fixes in the JA source.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+
+def probe(mu, shots=2000):
+    """Estimate the fraction of the spectrum below `mu`."""
+    executable = converter.transpile(
+        transpiler,
+        mu=mu,
+        degree=DEGREE,
+        delta=DELTA,
+    )
+    result = executable.sample(transpiler.executor(), shots=shots).result()
```
jij-yuichiro-nakano: Please create and reuse a seeded executor here and for the final sampling, for example with `AerSimulator(seed_simulator=SEED, max_parallel_threads=1)`. Otherwise the rendered EN/JA results can change between runs.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+    return converter.success_probability(result)
+
+
+# Two of the 64 states sit at the ground energy, so the predicate saturates at
+# 2 / 64 once mu clears it.
+print(f"p(mu=4.0) = {probe(4.0):.4f}   (exact: {(energies < 4.0).mean():.4f})")
```
jij-yuichiro-nakano: Please store this sampled value and assert that it is close to the exact result with an explicit shot-noise tolerance. Please also add `assert x_lower <= ground_energy <= x_upper` after the binary search. Printing alone does not protect these key results from regressions.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+print(f"alpha' = {shifted.normalization}   signal={n_signal}  system={n_system}")
+
+
+# %% [markdown]
+# ### 1. The block encoding $U$
+#
+# The Ising Hamiltonian is written as a linear combination of unitaries and
+# realised as $\operatorname{PREPARE}$, $\operatorname{SELECT}$, $\operatorname{PREPARE}^\dagger$. $\operatorname{PREPARE}$ loads the coefficient
+# amplitudes onto the signal register, $\operatorname{SELECT}$ applies the corresponding Pauli-$Z$
+# word to the system conditioned on the signal, and $\operatorname{PREPARE}^\dagger$ un-computes.
```
jij-yuichiro-nakano: What does “Pauli-Z word” mean here? If this is one component of the Pauli expansion, please use “Pauli-Z term” and define it if necessary.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+qsvt_reflector.draw(phi=phi_demo, inline=True, inline_depth=0, fold_loops=False)
+
+# %% [markdown]
+# ### 3. The Lin & Tong projector
+#
+# The goal is to turn the reflection into projection $P_\mu = (I \pm R)/2$. 
+# However, this projector is **not unitary**, so no circuit can simply apply it.
+#
+# The trick is that it is a sum of two unitaries, $I$ and $R$, so a Hadamard test
+# can realise it probabilistically. One probe qubit is put in superposition, the
```
jij-yuichiro-nakano: Please define the probe qubit before using this term, for example as the ancilla that coherently selects the identity and reflection branches in the Hadamard test.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+
+
+# %%
+gamma = np.sqrt(degeneracy / energies.size)
+x_lower, x_upper = binary_search_ground_energy(probe, low, high, gamma=gamma, h=1.0)
+
+print(f"\nlambda_0 in [{x_lower:+.2f}, {x_upper:+.2f}]   true lambda_0 = {ground_energy:+.2f}")
+
+# %% [markdown]
+# ## Optional
```
jij-yuichiro-nakano: Please turn this into a “Result” section and combine the binary-search result with the final state-preparation result. It would also be helpful to plot the recovered partition here, using different node colors and highlighting the cut edges.

docs/en/algorithm/qsvt_eigenstate_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+print(f"most sampled partition: {side} | cut={cut}")
+
+assert optimal / kept > 0.85
+assert cut == 3 and len(side) == num_nodes // 2
+
+# %% [markdown]
+# From the final sample distribution `sampleset`, we recover the most sampled partition and use it as the solution to the partition problem. The two triangles $\{0,3,4\}$ and $\{1,2,5\}$ describe the same split, so either side may come out on top from one run to the next; the corresponding objective (minimal cut) is $3$.
+
+# %% [markdown]
+# ## Summary
```
jij-yuichiro-nakano, line 562: Please follow the algorithm-page skeleton and reduce this summary to about three concise takeaways. The current bullets repeat too much implementation detail.

docs/ja/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,413 @@
...
+# tags: [algorithm, optimization, oracle-based]
+# ---
+#
+# # QSVT Filteringによるグラフ分割問題の求解
+#
+# 本チュートリアルでは、Quantum Singular Value Transform(QSVT)と量子化学に着想を得た手法を使って、与えられたハミルトニアンの基底エネルギーを求め、グラフ分割問題を解く方法を説明します。
+#
+# > L. Lin and Y. Tong, *Near-optimal ground state preparation*, Quantum 4, 361 (2020) [arXiv:2002.12508](https://arxiv.org/abs/2002.12508)
+#
+# Lin & Tongの基底エネルギーアルゴリズムは、*フィルタリング*によってコストハミルトニアンの最小固有値を求めます。閾値パラメータ$\mu$を1つ決めると、Linear Combination of Unitaries(LCU)によって$H - \mu I$のblock encodingを構築できます。さらにsign関数のQSVT近似を用いて、閾値パラメータ$\mu$の片側にある固有空間へのprojectorを作ります。post-selectionを通過したショットの割合が、$\mu$をどちらへ動かせばよいかを古典的な二分探索に伝えます。
```
jij-yuichiro-nakano: Please make the Japanese terminology consistent throughout this page:

- `block encoding` → `ブロックエンコーディング`
- `Ising` → `イジング`
- use `判定条件` instead of the formal-logic translation of “predicate”

The final abstract sentence could be rewritten as:

`本チュートリアルでは、この手順をグラフ分割問題に適用し、ブロックエンコーディングの構築、単一の閾値での評価、二分探索による基底エネルギーの推定までを示します。`

For the predicate explanation, I suggest:

`この値を、古典的な二分探索で探索方向を決める判定条件として用います。`

docs/en/algorithm/qsvt_filter_graph_partitioning.py
```diff
@@ -0,0 +1,490 @@
...
+# # QSVT Filtering to solve Graph Partitioning
+#
+# This tutorial explains how to use Quantum Singular Value Transform (QSVT) 
+# and chemistry inspired methods to find the ground energy of a given Hamiltonian
+# and solve the Graph Partitioning problem.
+#
+# > L. Lin and Y. Tong, *Near-optimal ground state preparation*, Quantum 4, 361 (2020) [arXiv:2002.12508](https://arxiv.org/abs/2002.12508)
+#
+# Lin & Tong's ground-energy algorithm finds the lowest eigenvalue of a cost
+# Hamiltonian by *filtering*. For a given threshold parameter $\mu$, we use
```
jij-yuichiro-nakano: [Optional] Please remove decorative italics where no term is being introduced. The same pattern appears in several later paragraphs.

jij-yuichiro-nakano (changes requested):
@baptistechev  Thank you for adding this feature! I left inline comments about the public API contract, optional-dependency behavior, validation, backend coverage, and tutorial structure.

qamomile/circuit/algorithm/__init__.py
baptistechev: Renamed. The module is now `qsvt_eigenstate_filter` in both trees, the converter is `QSVTEigenstateFilterConverter`, and the builders are `qsvt_filter_projector` / `qsvt_filter_probe`. The test files and the tutorial page follow the same name. The specificity now lives in the module name, which is why the builders could drop back to the shorter form; if you would rather see the eigenstate token in the function names too (`eigenstate_filter_projector` / `eigenstate_filter_probe`), say so and I will rename them — it is a one-line change on the export list. Fixed in 6e3a3184.

qamomile/optimization/qsvt_filter.py
baptistechev: Fixed in 6e3a3184. All three guards now reject `bool` explicitly (it satisfies every numeric comparison while meaning nothing as a degree, a width, or a rescaling) and `delta` / `scale` additionally require `math.isfinite`. Tests cover `True`, `nan`, `+inf` and `-inf` for each.

The `phi` contract is now stated in `transpile`'s docstring and enforced: it must hold an even number of at least two finite phases. An empty sequence was already rejected by the length check but only implicitly, so it is documented and has its own test case; non-finite entries raise a separate `ValueError` naming the offending values, rather than compiling into a rotation angle the backend turns into garbage.

qamomile/optimization/qsvt_filter.py
baptistechev: Fixed in 6e3a3184. The `pragma: no cover` is gone and both contracts are tested by hiding `pyqsp` with `monkeypatch.setitem(sys.modules, ...)`, which is the interpreter's own "module unavailable" marker and so reproduces a missing install even in a CI environment that has it:

- `test_transpile_without_pyqsp_points_at_the_extra` — omitting `phi` raises the guided `ImportError` mentioning `qamomile[qsvt]`.
- `test_transpile_with_precomputed_phases_does_not_need_pyqsp` — supplying `phi` still transpiles and produces a correctly sized circuit.

Neither is guarded by `importorskip`, since they are exactly the with-and-without contract.

tests/optimization/test_qsvt_eigenstate_filter.py
baptistechev: Fixed in 6e3a3184. `decode_to_binary_sampleset` now takes `SampleResult[tuple[list[int], list[int], list[int]]]`, and a `decode()` override with the same payload type delegates to the base implementation. Both carry `# type: ignore[override]` because they narrow the base class's `SampleResult[list[int]]` — the same pattern `QRAOBaseConverter.decode` already uses.

Added `_check_qsvt_eigenstate_filter_decode_types` to `tests/typecheck/public_api.py`, which `assert_type`s both entry points against a probe-shaped `SampleResult`. `zuban check` is clean on it.

tests/circuit/algorithm/test_qsvt_filter.py
baptistechev: Fixed in 6e3a3184. Seeds on the randomized test are down to `[0, 42]` as suggested.

Added `test_probe_with_a_two_qubit_signal_register_on_every_sdk`: three Ising terms `{(0,), (1,), (0,1)}` on two system qubits, so `num_signal_qubits == 2` and PREPARE carries an unused fourth amplitude inside the controlled QSVT reflection. Phases are the fixed Wx sequence `[0, pi, 2*pi, pi/3]`, so the boundary angles are always exercised. It runs both the sampler and the estimator through the shared `sdk_transpiler` fixture, so it covers every backend.

One thing worth recording: the reference value did not need a new derivation. The QSVT good block is a polynomial of `A/alpha` alone, so it stays `-i` times the Wx-convention QSP polynomial however wide the signal register is — verified numerically against the dense `_reference_qsvt` path to 4e-15. The closed form is now factored into `_expected_success_probability` and shared by both tests. Verified locally on all three backends after installing the `cudaq-cu12` and `quri_parts` extras: 20 passed, no skips (CUDA-Q 0.14.0).

tests/optimization/test_qsvt_eigenstate_filter.py
baptistechev: Fixed in 6e3a3184. Added `test_a_fully_rejected_result_decodes_to_nothing`: a five-shot result where three shots are dropped by the projector bit and two by a non-zero signal register.

The contract it pins, which I checked rather than assumed: `success_probability` is `0.0`, `decode_to_binary_sampleset` returns a `BinarySampleSet` with empty `samples` / `num_occurrences` / `energy`, and OMMX tolerates the empty sample list — `decode()` returns a `SampleSet` with no `sample_ids` and empty `objectives` rather than raising. So no special-casing was needed in the converter; the test now locks that behaviour in.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. The manual blockquote is gone, replaced by `{cite:p}` with the arXiv DOI, and the citation is reused where the binary search is attributed in the Algorithm section. Mirrored in the JA source. Verified in the built HTML: the DOI resolves to the full reference (title, authors, arXiv link) with no unresolved-citation warning.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. The cell is now a valid Jupytext install cell and lists the extras the page actually uses:

```python
# # !pip install "qamomile[qiskit,qsvt,visualization]" networkx
```

Mirrored in the JA source.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. Now `[QAOA for Graph Partitioning](qaoa_graph_partition)`, matching the link format used in `qaoa_maxcut.py`. The JA source links to the same page under its Japanese title.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. The Algorithm section no longer mentions the converter at all. It now walks the algorithm itself: the shifted operator `H - mu*I` and why the sign function is what separates the two sides; what a block encoding is and where the subnormalization `alpha'` comes from, so the encoded object is explicitly the normalized operator `(H - mu*I) / alpha'`; the QSVT alternation and the statement that `p((H - mu*I)/alpha') ~= sign((H - mu*I)/alpha')`; the Hadamard test turning the resulting reflection into `P_mu`; and the predicate `P(mu)`. The Lin and Tong binary-search procedure, which used to sit in its own section, is now described here too, since it is algorithm rather than implementation. Converter-specific material moved into the implementation section. Mirrored in the JA source.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. The page now follows the algorithm skeleton: `Problem Settings` -> `Algorithm` -> `Implementation with Qamomile` -> `Result` -> `Summary`. "Using the converter" and "Building blocks" are subsections of `Implementation with Qamomile` (the three circuit drawings demoted to `####`), and the binary-search function definition joins them there. Same structure applied to the JA page.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. Both bare `::: ` lines are now proper `:::{note}` ... `:::` blocks, matching the usage in `pce_maxcut.py`. Mirrored in the JA source, and both pages rebuilt to confirm they render as admonitions.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. The page now creates one seeded executor up front and reuses it for the probe and for the final sampling:

```python
executor = transpiler.executor(
    backend=AerSimulator(seed_simulator=SEED, max_parallel_threads=1)
)
```

Mirrored in the JA source, and both `.ipynb` files were re-executed against it, so the committed outputs are the reproducible ones.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. The probe value is stored and checked against the exact fraction with an explicit tolerance:

```python
shot_noise = 6.0 * np.sqrt(exact_p4 * (1.0 - exact_p4) / SHOTS)
assert abs(sampled_p4 - exact_p4) < shot_noise + 0.01
```

Six standard deviations of shot noise at 2000 shots is about 0.023, and the extra 0.01 covers the blur of a finite-degree sign approximation; the observed gap under the seeded executor is 0.00025, so there is plenty of headroom without the assertion becoming vacuous. `assert x_lower <= ground_energy <= x_upper` now follows the binary search. Mirrored in the JA source.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. It was one component of the Pauli expansion, so it now reads "Pauli-$Z$ term". Mirrored in the JA source (`Pauli-$Z$項`).

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. The probe qubit is now introduced where it first appears, as the ancilla that coherently selects between the identity and reflection branches of the Hadamard test, before the text goes on to use the term. Mirrored in the JA source.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. The section is now `## Result` and holds both the binary-search bracket and the final state-preparation decode. Added a plot of the recovered partition on the same `spring_layout` as the input graph, so the two figures are directly comparable: node colour is the side of the split, and the cut edges are drawn dashed in crimson with the cut size in the title. Mirrored in the JA source.

docs/en/algorithm/qsvt_eigenstate_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. The summary is down to three takeaways: the end-to-end result, why the algorithm is probabilistic (a non-unitary projector realized by a Hadamard test, hence the post-selection), and the single number every knob traces back to — the normalized gap, which is also why the Ising constant is held out. Implementation detail that was repeated there now lives only in the sections that introduce it. Mirrored in the JA source.

docs/ja/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. `block encoding` -> `ブロックエンコーディング` and `Ising` -> `イジング` throughout the page (and in the JA algorithm index card, which had the same wording). "predicate" is now `判定条件` everywhere, including inside the code comments and the `binary_search_ground_energy` docstring. Both of your suggested sentences are used verbatim for the abstract's closing line and for the predicate explanation.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Fixed in a834877a. Removed the decorative italics through the page (`*below*`, `*worse*`, `*control*` and friends). What is left is emphasis that carries meaning: terms at their point of introduction — `*filtering*`, the `*signal*` register, the `*probe qubit*`, the `*reflection*`, the `*bracket*` of two grid points — and one contrastive `*only*` where the text distinguishes keeping only the ground eigenspace from merely answering the predicate. Same pass applied to the JA source.

tests/circuit/algorithm/test_qsvt_filter.py
jij-yuichiro-nakano: Thank you. The multi-signal-qubit coverage, sampler/estimator execution, backend matrix, and seed reduction look good.

One detail remains: `[0, pi, 2*pi, pi/3]` is defined in the Wx convention and converted before binding. The phases actually passed to `qmc.qsvt` are `[pi/4, 3*pi/2, 5*pi/2, 7*pi/12]`, so the backend lowering does not exercise `0`, `pi`, or `2*pi`.

Please bind a deterministic reflection-convention phase sequence containing `0`, `pi`, and `2*pi` directly, and adapt the reference calculation accordingly.

tests/optimization/test_qsvt_eigenstate_filter.py
jij-yuichiro-nakano: Thank you. The converter override and the public API type-checking test address the main contract issue. One requested change remains: the valid probe results in the runtime tests are still annotated as `SampleResult[Any]`. Please use `SampleResult[tuple[list[int], list[int], list[int]]]` there as well.

Also, the test docstring still says that `decode()` is inherited, although it is now overridden.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
jij-yuichiro-nakano: Thank you. The changes addressing the original comment look sufficient. I have one additional request: please add a brief overview at the beginning of the `Implementation with Qamomile` section explaining that Qamomile provides eigenvalue-filtering functionality through `QSVTEigenstateFilterConverter`. Stating this explicitly will make it clear to readers that the algorithm introduced in this tutorial is available as part of Qamomile.

tests/optimization/test_qsvt_eigenstate_filter.py
baptistechev: Fixed in ecfd54ad. The valid probe results in the runtime tests (the OMMX decode test, the zero-shot test, and the fully-rejected test) are now annotated as `SampleResult[tuple[list[int], list[int], list[int]]]`. The docstring now says `decode()` is overridden to accept the `(projector, signal, system)` payload and routes through `decode_to_binary_sampleset`.

`test_decode_rejects_results_that_are_not_probe_measurements` still uses `SampleResult[Any]` on purpose: it passes a flat `list[int]` payload to check the `ValueError`, so the tuple type would not fit there.

tests/circuit/algorithm/test_qsvt_filter.py
baptistechev: Good catch, thanks. Fixed in ecfd54ad. `test_probe_with_a_two_qubit_signal_register_on_every_sdk` now binds the reflection-convention phases `[0, pi, 2*pi, pi/3]` directly to `qmc.qsvt`, so the backend lowering sees `0`, `pi`, and `2*pi` as given. For the reference, a new `_to_wx_phases` helper (the exact inverse of `_to_reflection_phases`) converts them back to the Wx convention, and `_wx_qsp_polynomial` evaluates the analytic success probability from there. The test passes for sampling and estimation on qiskit, quri_parts, cudaq, and braket.

docs/en/algorithm/qsvt_filter_graph_partitioning.py
baptistechev: Done in ecfd54ad. `Implementation with Qamomile` now opens with a short overview: Qamomile provides the eigenstate-filtering algorithm as `QSVTEigenstateFilterConverter` in `qamomile.optimization.qsvt_eigenstate_filter`, and the converter takes care of the block encoding, phase synthesis, probe circuit, and post-selection/decoding. The overview also says what the subsections cover (one threshold, building blocks, binary search). The same paragraph is in the JA page, and both notebooks are synced.

PR #634, opened 2026-07-27
Thread module-global-shadowed locals through if branches

ksk-jij:
## Summary

A kernel-local variable that shares a name with a module global was treated as a global by the control-flow transform. `VariableCollector` excludes global names from a branch's input/output set, so a same-named local was dropped from `if`-branch dataflow and assignments to it inside the branch were silently discarded (BACKLOG-8898). This is easy to hit in notebooks, where a variable left behind by a previous cell becomes a module global.

Changes:

- **Primary fix.** `transform_control_flow` now subtracts function-local bound names (`co_varnames | co_cellvars`) from `global_names`. A name assigned anywhere in the function body is a Python local for the whole scope regardless of a same-named global, so it must be threaded as dataflow. Read-only globals and `global`-declared names never appear in `co_varnames`/`co_cellvars`, so genuine globals stay classified as globals and are unaffected.
- **Secondary fix.** `VariableCollector`'s call-target exclusion was visit-order dependent: a name seen as a variable before the call that names it leaked into the collected set, while the same name seen after the call was excluded. A new two-pass `collect(*nodes)` method pre-scans call targets across all nodes before collecting variable dataflow, making the result order-independent. All collector call sites now route through it. Nested function definitions (sync and async) are skipped consistently by both passes.

The IR abstraction level and layer boundaries are unchanged; this is a frontend-only transform fix.

## Test plan

- `uv run pytest tests/circuit/test_control_flow_if.py tests/circuit/test_ast_transform.py` — new `TestIfModuleGlobalShadowing` (plain-local and cell-variable shadowing paths, plus the compile-time-folded value resolving to the branch assignment) and `TestVariableCollectorVisitOrderIndependence` (call-target exclusion is order independent) pass alongside the existing suites.
- `uv run pytest tests/` — full unit suite green (9636 passed).
- `uv run ruff check qamomile/ tests/` and `uv run ruff format --check qamomile/ tests/` — clean.
- `uv run zuban check qamomile/` — no new type errors in changed files.

https://claude.ai/code/session_01PHbNShC7tKEB8uDojG2F5j

### DIFF
--- a/qamomile/circuit/frontend/ast_transform.py
+++ b/qamomile/circuit/frontend/ast_transform.py
@@ -54,6 +54,49 @@ class VariableCollector(ast.NodeVisitor):
         self._store_names: set[str] = set()
         self._store_order: list[str] = []
 
+    def _prescan_call_targets(self, node: ast.AST) -> None:
+        """Record call-target names so exclusion is visit-order independent.
+
+        ``visit_Call`` excludes the name of a called function (``foo`` in
+        ``foo(q)``) from the collected variables. Doing so incrementally makes
+        the result depend on visit order: a name seen as a plain variable
+        *before* the call that names it would leak into ``vars``, while the same
+        name seen after the call would be excluded. Collecting every call target
+        up front removes that dependency. Nested function definitions (sync and
+        async) are not descended into, mirroring the main pass
+        (``visit_FunctionDef`` / ``visit_AsyncFunctionDef``), so their call
+        targets do not affect the enclosing scope's dataflow.
+
+        Args:
+            node (ast.AST): Subtree to scan for called-function names.
+        """
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
+            return
+        if isinstance(node, ast.Call) and isinstance(node.func, ast.Name):
+            self._exclude.add(node.func.id)
+        for child in ast.iter_child_nodes(node):
+            self._prescan_call_targets(child)
+
+    def collect(self, *nodes: ast.AST) -> "VariableCollector":
+        """Collect variable dataflow over ``nodes`` in two order-independent passes.
+
+        The first pass records every call target across all nodes; the second
+        pass gathers variable dataflow while honouring those exclusions. Running
+        both passes over the full node set makes the result independent of the
+        order in which statements (and the calls within them) appear.
+
+        Args:
+            *nodes (ast.AST): Statements or expressions to collect over.
+
+        Returns:
+            VariableCollector: This collector, to allow fluent property access.
+        """
+        for node in nodes:
+            self._prescan_call_targets(node)
+        for node in nodes:
+            self.visit(node)
+        return self
+
     def visit_Call(self, node: ast.Call):
         """Exclude the function name of a call."""
         if isinstance(node.func, ast.Name):
@@ -154,7 +197,26 @@ class VariableCollector(ast.NodeVisitor):
             self.visit(target)
 
     def visit_FunctionDef(self, node: ast.FunctionDef):
-        """Skip traversal of inner function definitions."""
+        """Skip traversal of inner function definitions.
+
+        A nested definition introduces its own scope; its names are not part of
+        the enclosing block's dataflow, so it is not descended into (which also
+        keeps this pass consistent with ``_prescan_call_targets``).
+
+        Args:
+            node (ast.FunctionDef): Inner function definition to skip.
+        """
+        pass
+
+    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef):
+        """Skip traversal of inner async function definitions.
+
+        Mirrors ``visit_FunctionDef`` so both function-definition kinds are
+        treated identically by the main pass and the call-target pre-scan.
+
+        Args:
+            node (ast.AsyncFunctionDef): Inner async function definition to skip.
+        """
         pass
 
     def visit_Name(self, node: ast.Name):
@@ -283,7 +345,7 @@ class ControlFlowTransformer(ast.NodeTransformer):
         per_stmt_collectors: list[VariableCollector] = []
         for stmt in body:
             c = VariableCollector(global_names=self._global_names)
-            c.visit(stmt)
+            c.collect(stmt)
             per_stmt_collectors.append(c)
 
         # Build suffix-union of read vars in reverse (O(n))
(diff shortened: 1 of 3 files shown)


ksk-jij:
## Branch summary (English translation)

### Problem overview

This branch fixes a bug where an assignment made to a function-local variable inside an `if` branch is silently dropped when that local shares its name with a module global that shadows it (BACKLOG-8898). On `main`, a kernel that uses a local such as `count` inside the body of the qkernel decorator (`@qm.qkernel`, the decorator that compiles a user-written quantum kernel) loses the branch assignment when a same-named `count` also exists at module scope: the AST transform misclassifies the local as a "global", excludes it from the `if` branch's dataflow (the input/output variable set), and discards the `count = 1.0` inside the branch. This is especially easy to hit in notebooks, where a variable defined in a previous cell lingers as a module global. On this branch, the local is correctly threaded through the `if` branch regardless of whether a same-named global exists, and the in-branch assignment takes effect.

The reproduction is: put `count = 123` at module scope and `build(flag=1)` a kernel of the form `count = 0.0; if flag == 1: count = 1.0; return count`. On `main`, the generated `IfOperation` (the IR operation representing a conditional branch) has empty `true_yields` and `results`, and the `Block` (the IR container holding one kernel body's operation sequence) has an empty `output_values`, so `count`'s final value is pinned to the pre-branch `0.0`. On this branch these are populated correctly, and folding the compile-time binding (`flag=1`) resolves the final value to `1.0`. The root cause is that `transform_control_flow` (`qamomile/circuit/frontend/ast_transform.py:2308`) built `global_names` (the set of names to exclude from dataflow) from the keys of `func.__globals__` without subtracting function-local names, so `VariableCollector` (`ast_transform.py:33`, the AST visitor that collects variables used/mutated within a block) excluded the local as if it were a global (`global_names` additionally gains the free variables `co_freevars` when a closure is present, but either way there was no step subtracting function-local names). This branch also resolves a secondary issue in the same file: `VariableCollector`'s exclusion of call-target names (the function name in a call) depended on visit order.

### Frontend (user-facing code level) changes

From the user's perspective the change is behavior-only; the ergonomics, API signatures, and error messages of the qkernel decorator (`@qm.qkernel`) are unchanged. Previously, only when a same-named name existed at module scope, assigning to or updating a local inside an `if` branch could silently lose the result; with this fix such locals are always threaded correctly through the branch. In the example below, `main` discards the in-branch `count = 1.0` even with `flag=1` and `count` never reaches the branch dataflow (`Block.output_values` is empty, and a downstream use of `count` stays pinned to the pre-branch `0.0`), whereas this branch resolves to the post-branch value `1.0` as expected.

```python
import qamomile.circuit as qm
from qamomile.circuit.frontend.handle.primitives import Float

count = 123  # a same-named module global (e.g. left over from a previous cell)

@qm.qkernel
def circuit(flag: qm.UInt) -> Float:
    count = 0.0
    if flag == 1:
        count = 1.0
    return count

# this branch: flag=1 resolves to 1.0 (on main the in-branch assignment was dropped and never reached the output)
```

### Backend (IR and overall) changes

The first fix, in the construction of `global_names` in `transform_control_flow` (`qamomile/circuit/frontend/ast_transform.py:2308`), adds a step that subtracts function-local bound names (`ast_transform.py:2376-2377`). By Python's scoping rules, a name assigned anywhere in the function body (and parameters) is local to the whole function scope unless a `global` declaration is present; such names appear in `func.__code__.co_varnames`, and locals captured by a nested function appear in `func.__code__.co_cellvars`. Subtracting both sets from `global_names` makes a shadowed local be treated as dataflow even when a same-named global exists. Read-only global references and names under a `global` declaration never appear in `co_varnames`, so genuine globals remain on the exclusion side.

The second fix introduces two-pass collection into `VariableCollector` (`ast_transform.py:33`). `_prescan_call_targets` (`ast_transform.py:57`) walks the target subtree and gathers call-target names into `_exclude` first, and `collect` (`ast_transform.py:80`) collects in two passes ("pre-scan over all nodes, then the normal visit"). This makes the result independent of visit order when a name appears both as a variable use and as a call target. In line with this, all 10 sites that previously called `VariableCollector(...).visit(...)` were switched to `.collect(...)` (`collect(node)` for a single node, `collect(*nodes)` for a list). To keep the pre-scan and the normal pass consistent, nested function definitions are not descended into for either sync or async, and `visit_AsyncFunctionDef` (`ast_transform.py:211`) was added so it is skipped just like the existing `visit_FunctionDef` (`ast_transform.py:199`).

For regression tests, `TestIfModuleGlobalShadowing` (`tests/circuit/test_control_flow_if.py:1988`) was added for the first fix, verifying the `co_varnames` path (a plain local), the `co_cellvars` path (a cell variable captured by a nested function), and the final-value resolution after compile-time folding. For the second fix, `TestVariableCollectorVisitOrderIndependence` (`tests/circuit/test_ast_transform.py:463`) was added, verifying that call-target names are excluded consistently regardless of statement order.

### Alternatives considered and why they were not chosen

For how to collect "function-local names" in the first fix, the options were (A) `func.__code__.co_varnames` only, (B) `co_varnames | co_cellvars`, and (C) walk the function-definition AST to collect names in Store context. Option (B) was chosen. Option (A) misses a global shadowed by a cell variable captured into a nested function (which appears only in `co_cellvars`, not `co_varnames`); the cell-variable shadowing case was confirmed to regress to empty output under (A). Option (C) conceptually matches the "collect Store names" intent, but `transform_control_flow` does not always compute the function-definition AST node on the path where `region_signatures` is supplied, so (B), which relies on the always-available `__code__`, is more robust.

For the second fix's approach, the options were (A) route every collection call through the two-pass `collect`, and (B) add a post-processing step that subtracts `_exclude` from the final `vars` after collection. Option (A) was chosen. The post-processing approach layers a subtraction on top of the existing incremental collection logic and increases the need to reconcile it with order-dependent derived attributes such as `store_order`, whereas the pre-scan approach has the clear meaning of "settle all call targets first, then collect" and structurally guarantees visit-order independence.

### Known limitations

For a pattern that reassigns a local holding a callable across `if` branches, the two-pass exclusion always drops that name from dataflow. When: when you want to thread such a local through a branch. Why: the intent is to consistently exclude names that appear as call targets, so when a variable use and a call share the same name, the call-side exclusion takes precedence. Future fix: if a concrete need to thread a callable local through branches in a qkernel arises, a collection rule that distinguishes calls from variable uses can be considered separately. This pattern does not arise in ordinary qkernel code (calling gates or kernels directly), so there is no practical impact at present. Additionally, the exotic pattern of writing to a module global via a `global` declaration is not explicitly covered by tests on this branch (that genuine globals remain on the exclusion side is guaranteed by the nature of `co_varnames`).

---

PR #641, opened 2026-08-03
Use allocated qubit names in affine diagnostics

ksk-jij:
## Summary

Fix consumed-qubit diagnostics to prefer the user-provided IR value name before falling back to an internal handle UUID.

Add a regression test covering a locally allocated qubit whose stale handle is reused after a gate.

## Root cause

The qubit constructor stores its display name in `Value.name`, while `Handle.name` remains unset. `Handle.validate_consumable()` checked only `Handle.name` before generating a UUID-based fallback, so diagnostics exposed a random label such as `qubit_90060270` instead of the requested name.

## User impact

Affine-type errors now identify `qmc.qubit("qubit")` as `Qubit 'qubit'`, making the diagnostic stable and directly traceable to user code.

## Validation

- `uv run pytest tests/circuit/test_affine_types.py -q` (388 passed)

### DIFF
--- a/qamomile/circuit/frontend/handle/handle.py
+++ b/qamomile/circuit/frontend/handle/handle.py
@@ -314,7 +314,7 @@ class Handle(abc.ABC):
             QubitConsumedError: If this quantum handle was already consumed.
         """
         if self._consumed and self._should_enforce_linear():
-            display_name = self.name or f"qubit_{self.id[:8]}"
+            display_name = self.name or self.value.name or f"qubit_{self.id[:8]}"
             first_use, reuse, consumed_at = _describe_consume_sites(
                 self, operation_name
             )
--- a/tests/circuit/test_affine_types.py
+++ b/tests/circuit/test_affine_types.py
@@ -488,6 +488,23 @@ class TestErrorMessageQuality:
         error = exc_info.value
         assert error.handle_name is not None
 
+    def test_error_message_uses_allocated_qubit_name(self):
+        """Allocated qubit diagnostics should use the user-provided name."""
+
+        @qkernel
+        def bad_circuit() -> qm.Bit:
+            qubit = qm.qubit("qubit")
+            qm.x(qubit)
+            qm.h(qubit)
+            return qm.measure(qubit)
+
+        with pytest.raises(QubitConsumedError) as exc_info:
+            bad_circuit.build()
+
+        error = exc_info.value
+        assert error.handle_name == "qubit"
+        assert "Qubit 'qubit'" in str(error)
+
     def test_error_message_contains_operation_names(self):
         """Error messages should name both operations involved."""
 


ksk-jij:
## 1. Problem Overview

Qamomile's affine type system ensures that a handle referring to a qubit can be used at most once, preventing an old handle from being reused accidentally after a gate is applied. The check itself worked correctly on the latest `origin/main`, but the diagnostic did not consult the user-provided name stored by `qmc.qubit("qubit")`. Instead, it displayed the first eight characters of an internally generated UUID (an identifier designed to avoid collisions), prefixed with `qubit_`. As a result, the same source code produced a different name on each run, making it difficult to identify which qubit allocation the error referred to.

The following code correctly raises an affine-type error on the latest `origin/main`, but the displayed name was not the user-provided `qubit`. `<UUID prefix>` is a placeholder for the eight characters that change on each run.

```python


@qmc.qkernel
def affine_example() -> qmc.Bit:
    qubit = qmc.qubit("qubit")
    qmc.x(qubit)
    qmc.h(qubit)
    return qmc.measure(qubit)

# Diagnostic pattern on the latest origin/main:
# Qubit 'qubit_<UUID prefix>' was already consumed by 'X' ...
```

This branch does not change affine-type checking or circuit semantics; it makes the same violation consistently report `Qubit 'qubit'`.

The qkernel API and reassignment rules for user-written code are unchanged. A gate consumes its input handle and returns a new handle, so valid code must capture the return value, for example with `qubit = qmc.x(qubit)`. If the return value is discarded and the old handle is reused, this branch places the name passed to `qmc.qubit()` in both the error message and `QubitConsumedError.handle_name`.

In the following example, the build fails as intended, and the output always starts with `Qubit 'qubit'`.

```python
try:
    affine_example.build()
except Exception as error:
    print(error)

# Qubit 'qubit' was already consumed by 'X' ...
```

With this change, rerunning the same violation in a notebook or regular Python file yields the name specified in the source code rather than a random internal name, making the problem easier to locate and fix.

This branch does not change any backend (the layer that converts programs for execution targets such as Qiskit) or the IR (Intermediate Representation). In this template section, the change concerns the frontend (the layer that converts user code into IR): the qubit-ownership check now receives the user-facing display name that was already stored internally, using the correct priority. The gate sequence remains unchanged; only the path that constructs the diagnostic string is modified.

### Propagating the User-Provided Name into Diagnostics

The internal flow is `qmc.qubit(name) → Value.name → Value.next_version() → new Qubit → Handle.validate_consumable()`. A `Handle` is the Python-side object that tracks a qubit's consumption state, while a `Value` represents an IR value and stores its display name.

First, `qmc.qubit(name)` stores the name in `Value.name` and returns a `Qubit` wrapping that `Value`. At this point, the Python-side `Handle.name` is unset (`qamomile/circuit/frontend/constructors.py:194-203`). When a gate subsequently creates the next `Value` through `Value.next_version()`, it preserves the same name, and a new `Qubit` wraps that value (`qamomile/circuit/ir/value.py:598-619`). Therefore, `Value.name` is a stable display name retained by both the initially allocated handle and gate-result handles.

Finally, `Handle.validate_consumable()`, which detects reuse of a consumed handle, selects the name used in the diagnostic. Before this change, it immediately fell back to the UUID whenever `Handle.name` was empty. After the change, it checks `Handle.name`, then `Value.name`, and finally the UUID, so an explicitly named qubit no longer exposes an internal identifier (`qamomile/circuit/frontend/handle/handle.py:306-331`).

```python
display_name = self.name or self.value.name or f"qubit_{self.id[:8]}"
```

### Preventing Display-Name Regressions with a Test

A regression test detects if a fixed bug returns in the future. The added test assigns the result of `qmc.qubit("qubit")` to `qubit`, discards the new handle returned by `qmc.x(qubit)`, and then passes the now-consumed original `qubit` to `qmc.h(qubit)` to raise `QubitConsumedError`. It verifies that both the error object's `handle_name` attribute and the message body use `qubit` (`tests/circuit/test_affine_types.py:491-506`). The test directly covers reuse of the initial allocation handle, while the implementation reads the diagnostic target's `Value.name`, so the same rule also applies to gate-result handles that preserve that name. After merging the latest main, all 388 affine-type tests and all 9,639 unit tests passed, as did Ruff linting and formatting checks and Zuban type checking.

One alternative was to change only the constructor so that `qmc.qubit()` returned `Qubit(value=value, name=name)`. However, gate results create new `Qubit` handles and currently do not copy `Handle.name` explicitly. By contrast, `Value.next_version()` already preserves `Value.name`, so reading `Value.name` in the diagnostic handles both the initial handle and post-gate handles under the same rule.

Another alternative was to reuse the existing display-name helper in `qamomile/circuit/frontend/qkernel_utils.py:65-74` directly. However, that module imports the `Handle` implementation, so importing it in the opposite direction would create a circular dependency in which each module attempts to load the other. Moving the helper into a different module would allow broader refactoring, but this branch chooses the smallest targeted change by applying the same priority in one line.

## 5. Known Limitations

When: Both `Handle.name` and `Value.name` are empty, regardless of how the handle was created. Why: No stable user-derived display name exists, so the diagnostic continues to use the first eight characters of the UUID as a fallback. Future fix: If the frontend gains a stable mapping to source variables, it could prefer that information over the UUID.

When: The import alias used in the error message's `Fix` example differs from the user's `qmc` alias. Why: The existing diagnostic string hardcodes `qm` and cannot detect the alias chosen by the caller. Future fix: A separate change could replace the example with alias-independent wording or standardize it on `qmc`.

PR #648, opened 2026-08-28
Fix Grover diffusion for single-qubit registers

jij-yuichiro-nakano:
## Summary

- Special-case single-qubit Grover diffusion with a `Z` reflection instead of constructing an MCX with an empty control register.
- Mark `_diffusion` as a qkernel so register-width-dependent control flow is resolved by the compiler pipeline.
- Extend cross-backend Grover coverage to `n` values of 1, 2, 3, and 5 for both sampling and expectation-value execution across Qiskit, QURI Parts, CUDA-Q, and Amazon Braket.
- Compare sampling outcomes against analytic marked-state probabilities and verify that single-qubit resource estimation requires no clean control ancillas.

## Problem

For a one-qubit register, `top = n - 1` is zero and `reg[0:top]` is empty. Passing that slice to `mcx` creates a zero-control multi-controlled X operation, which supported SDK emitters do not consistently accept. As a result, single-qubit Grover diffusion could not be transpiled or executed.

## Solution

For `n == 1`, use `Z` as the zero-control specialization of the multi-controlled Z reflection. The surrounding `H X ... X H` sequence yields the expected single-qubit diffusion up to a global phase. Registers with `n >= 2` retain the existing MCX construction.

## Changed Files

- `qamomile/circuit/stdlib/grover.py`

  - Mark `_diffusion` as a qkernel.
  - Handle single-qubit diffusion with `Z` instead of an MCX with no controls.
  - Preserve the existing MCX implementation for larger registers.

- `tests/circuit/test_grover.py`

  - Add single-qubit support to the test oracle.
  - Add resource-estimation coverage for single-qubit diffusion.
  - Extend sampling and expectation-value tests to `n = 1, 2, 3, 5` across all supported SDK backends.
  - Compare execution results against analytic references.

## Testing

- `uv run pytest tests/circuit/test_grover.py -m "" -q` (`89 passed`)

### DIFF
--- a/qamomile/circuit/stdlib/grover.py
+++ b/qamomile/circuit/stdlib/grover.py
@@ -115,6 +115,7 @@ def grover_iteration_count(
     return sp.floor((sp.pi / 4) * sp.sqrt(sp.Integer(2) ** n / m))
 
 
+@qmc.qkernel
 def _diffusion(reg: Vector[Qubit]) -> Vector[Qubit]:
     """Apply the Grover diffusion operator ``2|s><s| - I`` in place.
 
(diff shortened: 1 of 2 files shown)


ksk-jij (approved):
@jij-yuichiro-nakano Thank you! Looks good to me :)

